Issue #11, opened 2018-11-20
Add E2E tests with OpenSSL


Sean-Der:
Add some basic testing using s_client + s_server

The author should also refactor/re-use our existing E2E testing code so that we don't duplicate in both places.

backkem:
We should probably implement these types of tests globally so we can throw them at any layer of the stack. The beginnings of this are [here](https://github.com/pions/webrtc/blob/dtls-native/internal/transport/test/test.go). Eventually it could look like [this](https://github.com/libp2p/go-libp2p-transport/blob/master/test/utils.go).

igolaizola:
OpenSSL tests have been added in https://github.com/pion/dtls/pull/193

TODO:
- [ ] Make OpenSSL work with ed25519 test

at-wat:
I found that our implementation of Ed25519 signature calculation and hash used in `ServerKeyExchange`/`CertificateVerify` messages are wrong. Since both of our server and client have same bug, they can communicate each other.
I have made a fix for it, and now I'm cleaning the code.

One bad news is that OpenSSL haven't support DTLS Ed25519 yet.
I could confirm my fix works with this openssl branch: https://github.com/openssl/openssl/pull/11115

at-wat:
Ed25519 implementation of the master branch should be fixed now. We can enable the E2E test once OpenSSL supports DTLS 1.2 Ed25519.

PR #23, opened 2018-12-11
Implement E2E test for Client/Server

Sean-Der:
Relates to #11

### DIFF
--- a/pkg/dtls/conn.go
+++ b/pkg/dtls/conn.go
@@ -186,7 +186,7 @@ func (c *Conn) Write(p []byte) (int, error) {
 // Close closes the connection.
 func (c *Conn) Close() error {
 	c.notify(alertLevelFatal, alertCloseNotify)
-	c.stopWithError(errConnClosed)
+	c.stopWithError(ErrConnClosed)
 	return c.nextConn.Close()
 }
 
@@ -347,8 +347,9 @@ func (c *Conn) notify(level alertLevel, desc alertDescription) {
 
 func (c *Conn) signalHandshakeComplete() {
 	select {
-	case c.handshakeCompleted <- true:
+	case <-c.handshakeCompleted:
 	default:
+		close(c.handshakeCompleted)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


pkg/dtls/internal/udp/conn.go
```diff
@@ -105,7 +104,7 @@ readLoop:
 			n = copy(cBuf, buf[:n])
 			conn.sizeCh <- n
 		case <-conn.doneCh:
-			continue readLoop
+			return
```
Sean-Der: I get leaked routines with this, before it would deadlock on the ReadFrom above.

Not sure this is the right change, but wanted to throw it up here.

backkem: This doesn't seem like the right approach. The readLoop should keep passing packets to the Conns until the last one is closed. We'll have to look into it further.

pkg/dtls/conn.go
```diff
@@ -347,8 +347,9 @@ func (c *Conn) notify(level alertLevel, desc alertDescription) {
 
 func (c *Conn) signalHandshakeComplete() {
 	select {
```
backkem: I think this is racy since signalHandshakeComplete can be called from different routines. Two routines may decide to take the default case at the same time and both try to close the handshakeCompleted channel.

test/e2e/pion/util.go
```diff
@@ -0,0 +1,58 @@
...
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/hex"
+	"math/big"
+	"time"
+)
+
+func createSelfSigned() (*x509.Certificate, crypto.PrivateKey, error) {
```
backkem: This is starting to pop up in multiple places now. Should we conciser adding it to the package?

test/e2e/pion/simple_test.go
```diff
@@ -0,0 +1,149 @@
...
+				if seenClient && seenServer {
+					return
+				}
+			}
+		}
+	}()
+
+	// We have seen Client+Server communication! Now count goroutines to make sure we haven't leaked
+	time.Sleep(time.Second) // TODO racey
+	goRoutineCount := runtime.NumGoroutine()
```
backkem: Would be a nice addition to `pion/transport/test`.

pkg/dtls/conn.go
Sean-Der: ACK.

This will be fixed by merging your `shutdown` work right? I will fix in that PR

PR #26, opened 2018-12-17
Fix goroutine leak in handshake

kc5nra:


kc5nra:
Oopsy poopsy, this was an old branch.

PR #31, opened 2019-01-23
Cleanup

backkem:
Changed repo structure to match other Pions repos.
Changed CI to match other Pions repos.


conn.go
```diff
@@ -161,6 +161,7 @@ func (c *Conn) Read(p []byte) (n int, err error) {
 		return 0, c.getConnErr()
 	}
 	if len(p) < len(out) {
+		fmt.Println(errBufferTooSmall)
```
Sean-Der: Do we need this change? Mind putting it in another PR (just in case we need to revert debug)

conn.go
```diff
@@ -117,13 +117,13 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 			if err != nil {
 				c.stopWithError(err)
 				return
-			} else if c.getConnErr() != nil {
+			} /*else if c.getConnErr() != nil {
```
Sean-Der: Same as below, mind keeping this PR just to string/moving files around. We had a big move like this in pions/webrtc once and really burned me.

conn.go
backkem: Good catch 😳

PR #39, opened 2019-02-16
Implement Client Authentication

### LINKED ISSUE #38: Implement ClientAuthType
We need to implement [ClientAuthType](https://golang.org/pkg/crypto/tls/#ClientAuthType) from crypto/tls

Currently pion/webrtc only verifies a certificate when running as a client

Sean-Der:
Copy ClientAuthType pattern from crypto/tls, we don't verify the
certificates only fetch them for now

Resolves #38

### DIFF
--- a/server_handlers.go
+++ b/server_handlers.go
@@ -70,6 +70,15 @@ func serverHandshakeHandler(c *Conn) error {
 				return err
 			}
 
+		case *handshakeMessageCertificateVerify:
+			if c.remoteCertificate == nil {
+				return errCertificateVerifyNoCertificate
+			}
+			if err := verifyCertificateVerify(c.handshakeCache.combinedHandshake(serverExcludeRules(), true), h.hashAlgorithm, h.signature, c.remoteCertificate); err != nil {
+				return err
+			}
+			c.remoteCertificateVerified = true
+
 		case *handshakeMessageCertificate:
 			if c.currFlight.get() == flight4 {
 				c.remoteCertificate = h.certificate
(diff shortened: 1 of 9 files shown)


Sean-Der:
@backkem @kixelated all the tests pass, just having issues uploading the coverage results.

server_handlers.go
```diff
@@ -70,6 +70,15 @@ func serverHandshakeHandler(c *Conn) error {
 				return err
 			}
 
+		case *handshakeMessageCertificateVerify:
+			if c.remoteCertificate == nil {
+				return errCertificateVerifyNoCertificate
+			}
+			if err := verifyCertificateVerify(c.handshakeCache.combinedHandshake(serverExcludeRules(), true), h.hashAlgorithm, h.signature, c.remoteCertificate); err != nil {
```
kixelated: Can you break long lines for readability? There are too many nested layers.

```
if err := verifyCertificateVerify(c.handshakeCache.combinedHandshake(serverExcludeRules(), true), h.hashAlgorithm, h.signature, c.remoteCertificate); err != nil {
```

vs

```
combined := c.handshakeCache.combinedHandshake(serverExcludeRules(), true)
err := verifyCertificateVerify(combined, h.hashAlgorithm, h.signature, c.remoteCertificate)
if err != nil {
```

server_handlers.go
```diff
@@ -111,8 +120,20 @@ func serverHandshakeHandler(c *Conn) error {
...
 				}
+
+				if c.clientAuth == RequireAnyClientCert && c.remoteCertificate == nil {
+					return errClientCertificateRequired
+				} else if c.remoteCertificate != nil && !c.remoteCertificateVerified {
+					return errClientCertificateNotVerified
+				}
+
+				if c.clientAuth > NoClientCert {
+					c.localSequenceNumber = 6
```
kixelated: Is this something in the spec? Or are we incrementing the `localSequenceNumber` for another part of the code?

Sean-Der: Incrementing for another part of the code.

We only increment when we move a flight, because the connection is lossy we keep replaying and it makes it easier.

What would be nice is to have a function `baseSequenceNumberForFlight` or something like that, would probably be better then me just setting them manually.

PR #41, opened 2019-03-13
Implement resuming a connection from a previously serialized state

igolaizola:
New feature to export dtls state of an already running dtls connection in order to be able to resume it later. 

Use case:

Serialize and persist to an external data source the dtls state in order to resume it from a different running instance of the application, avoiding performing a new handshake.

### DIFF
--- a/resume.go
+++ b/resume.go
@@ -1,0 +1,36 @@
+package dtls
+
+import (
+	"net"
+)
+
+// Export extracts dtls state and inner connection from an already handshaked dtls conn
+func (c *Conn) Export() (*State, net.Conn, error) {
+	state, err := c.state.clone()
+	if err != nil {
+		return nil, nil, err
+	}
+	return state, c.nextConn, nil
+}
+
+// Resume imports an already stablished dtls connection using a specific dtls state
+func Resume(state *State, conn net.Conn, config *Config) (*Conn, error) {
+	// Custom flight handler that sets imported data and signals as handshaked
+	flightHandler := func(c *Conn) (bool, error) {
+		c.state = *state
+		c.signalHandshakeComplete()
+		return true, nil
+	}
+
+	// Empty handshake handler, since handshake was already done
+	handshakeHandler := func(c *Conn) error {
+		return nil
+	}
+
+	c, err := createConn(conn, flightHandler, handshakeHandler, config, state.isClient)
+	if err != nil {
+		return nil, err
+	}
+
+	return c, c.getConnErr()
+}
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Check out the **[contributing wiki](https://github.com/pions/webrtc/wiki/Contrib
 * [Sean DuBois](https://github.com/Sean-Der) - *Original Author*
 * [Michiel De Backker](https://github.com/backkem) - *Public API*
 * [Chris Hiszpanski](https://github.com/thinkski) - *Support Signature Algorithms Extension*
+* [Iñigo Garcia Olaizola](https://github.com/igolaizola) - *Support serialization and resumption"
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/client_handlers.go
+++ b/client_handlers.go
@@ -23,27 +23,27 @@ func clientHandshakeHandler(c *Conn) error {
 					if !ok {
 						return fmt.Errorf("Server responded with SRTP Profile we do not support")
 					}
-					c.srtpProtectionProfile = profile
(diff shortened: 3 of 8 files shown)


backkem:
Hi, this looks like a cool feature. For documentation purposes I'm wondering if this is based of a spec or some other implementation we can refer to?

Some housekeeping in order to make the CI happy:
- Please add yourself to the Contributing section of the root readme file.
- Please change your commit messages to follow the following rules:
https://github.com/pions/dtls/blob/ce8ae7c846ee3feae922e26a3a2dbcd5ae98b158/.github/lint-commit-message.sh#L9-L16

igolaizola:
> Hi, this looks like a cool feature. For documentation purposes I'm wondering if this is based of a spec or some other implementation we can refer to?

I don't have other implementation references on hand, but I'll say OpenSSL and others already have mechanism to do this.  There a some use cases that would benefit from this implementation:
 - accept dtls connection and then transfer to another process
 - when the transport protocol is not UDP, but wrapped inside another protocol

I've already discussed it here https://github.com/golang/go/issues/25351#issuecomment-415357066 for the `crypto/tls` package  and even tried to "hack it" here: https://github.com/igolaizola/resume-tls

Sean-Der:
This is really cool @igolaizola I would love to see this land. You should also post about it! I think a lot of people could build cool things with this.

What do you think of embedding `State` in `Conn` that way we don't have to do as much copying back and forth?

I would also add a timeout to your tests, maybe it is causing things to deadlock? Do you have failures locally.

igolaizola:
> This is really cool @igolaizola I would love to see this land. You should also post about it! I think a lot of people could build cool things with this.

Thanks! I really think that is an interesting feature. Hopefully something similar will be added to `crypto/tls`.

> What do you think of embedding State in Conn that way we don't have to do as much copying back and forth? The only thing that might be weird is the atomics :/ but I just worry things will diverge when have multiple structures, and will cause weird bugs in the future.

I'll take a look to it. I've coded it this ways in order to be less intrusive, but as you say in the future extra work must be done to have my code in sync.

> I would also add a timeout to your tests, maybe it is causing things to deadlock? Do you have failures locally.

I am dealing with some data race issues and I might added some deadlocks while I was trying to fix them.

igolaizola:
@Sean-Der PTAL. I've added changes to embed the `State` within the `Conn`. Also changed the serialization part to implement `encoding.BinaryMarshaler` and `encoding.BinaryUnmarshaler`

`go test -v -race ./...` seems to be working now locally but still failing on CI

resume.go
```diff
@@ -0,0 +1,32 @@
+package dtls
+
+import (
+	"net"
+)
+
+// Export extracts dtls state and inner connection from an already handshaked dtls conn
+func (c *Conn) Export() (*State, net.Conn, error) {
+	return &c.state, c.nextConn, nil
```
Sean-Der: Should we copy the state here? Right now it gives users a chance to do weird things (by accident, but I could maybe see weird bugs happening)

Sean-Der:
Can you move `srtpProtectionProfile` and `remoteCertificate` into state, otherwise after resumption users will get broken values.

Nice it looks like you are green in CI! Just a few nits from me and I am happy to merge :) Would you mind squashing into a single commit, and tagging with the github issue (just helps in the future with debugging/clean history)

thank you so much for all the work, this is a really great feature.

resume.go
igolaizola: I've added a clone method to resolve this

igolaizola:
> Can you move srtpProtectionProfile and remoteCertificate into state, otherwise after resumption users will get broken values.

Done!

> Nice it looks like you are green in CI! Just a few nits from me and I am happy to merge :) Would you mind squashing into a single commit, and tagging with the github issue (just helps in the future with debugging/clean history)

Done! Also improved a little bit the serialization in order to add a `clone` method.

> thank you so much for all the work, this is a really great feature.

Thanks to you for dealing with this project. Right now dtls is a missing feature in the standard crypto library.

Sean-Der:
Hey @igolaizola 
I added you to the Pions organization, and you can now merge your PR :)

Thank you so much for working on this, and if you have the time I would love your continued involvement working on DTLS! I think it would be a great thing for the community to have a good implementation, so hoping we can continue growing it until we get to that point!

igolaizola:
> Thank you so much for working on this, and if you have the time I would love your continued involvement working on DTLS! I think it would be a great thing for the community to have a good implementation, so hoping we can continue growing it until we get to that point!

I'll try to help and improve the package 👍

PR #54, opened 2019-04-06
Don't modify config.LoggerFactory

hugoArregui:
Avoid possible data race

Related to pion/webrtc#361


Sean-Der:
Nice, thank you @hugoArregui do you mind reverting my change [here](https://github.com/pion/dtls/commit/f365b82f4a96e8b5fa614b649dfcc7dda5224805#diff-12a8f43cbd312666a757938b8140deebL27) also. Have all the tests share the same config, I fixed it and didn't think it through.

hugoArregui:
@Sean-Der done!

PR #58, opened 2019-04-09
Move locks to write locks where needed

Sean-Der:
flight5 for client sets localCertificateVerify

flight4 for server sets localKeySignature


Sean-Der:
@backkem no worries, I shouldn't have written such garbage code in the first place :p

Issue #47, opened 2019-03-27
[WIP] Cannot trace dtls details.


jinleileiking:
If this package do not log, We cannot trace the detalls of dtls. For example, what ciphers does the server give, etc.

Sean-Der:
Closed by f365b82f4a96e8b5fa614b649dfcc7dda5224805

Feel free to add more @jinleileiking thanks again for the PR :)

PR #59, opened 2019-04-10
fix log error and add more trace logs

jinleileiking:
Relates #47

#### Reference issue
Fixes #...

### DIFF
--- a/client_handlers.go
+++ b/client_handlers.go
@@ -188,12 +188,13 @@ func clientHandshakeHandler(c *Conn) error {
 }
 
 func clientFlightHandler(c *Conn) (bool, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
 	switch c.currFlight.get() {
 	case flight1:
 		fallthrough
 	case flight3:
-		c.lock.RLock()
-
 		extensions := []extension{
 			&extensionSupportedEllipticCurves{
 				ellipticCurves: []namedCurve{namedCurveX25519, namedCurveP256},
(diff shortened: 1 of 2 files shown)


jinleileiking:
Not ready for merge. I will and more trace logs.

jinleileiking:
@Sean-Der  Ready for merge :) 
@backkem

conn.go
```diff
@@ -377,13 +378,16 @@ func (c *Conn) handleIncomingPacket(buf []byte) error {
...
+		c.log.Trace("<- Alert")
 		if content.alertDescription == alertCloseNotify {
 			return c.Close()
 		}
 		return fmt.Errorf("alert: %v", content)
 	case *changeCipherSpec:
+		c.log.Trace("<- ChangeCipherSpec")
 		c.setRemoteEpoch(c.getRemoteEpoch() + 1)
 	case *applicationData:
+		c.log.Trace("<- ApplicationData")
```
backkem: I would avoid this trace since this will be an extremely hot code path in more demanding applications.

conn.go
```diff
@@ -377,13 +378,16 @@ func (c *Conn) handleIncomingPacket(buf []byte) error {
 
 	switch content := r.content.(type) {
 	case *alert:
+		c.log.Trace("<- Alert")
```
backkem: Maybe print the type of alert as well.

conn.go
jinleileiking: I did not know when will go to this path

jinleileiking: If I do not enable the trace. Are There will be  any performance issue？

PR #63, opened 2019-05-07
invoke flight handler earlier to speed up DataChannel handshake

LanderN:
#### Description
This PR speeds up DataChannel creation by about 1 second.
Connection flightHandler is invoked immediately on startHandshakeOutbound instead of waiting for the first tick of workerTicker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Daniele Sluijters](https://github.com/daenney) - *AES-CCM support*
 * [Jin Lei](https://github.com/jinleileiking) - *Logging*
 * [Hugo Arregui](https://github.com/hugoArregui)
+* [Lander Noterman](https://github.com/LanderN)
 
 ### License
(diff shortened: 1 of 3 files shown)


Sean-Der:
Oh nice! I keep making this mistake and using tickers and forgetting that they don't have a 'startup tick'

If you don't have the bandwidth please push back, but it would great to see something like this PR https://github.com/pion/ice/commit/f58e725a2cbc15c66dbd66d268f1deede8ad1748

We already have `FlightInterval` so it should be pretty easy to just add a test + channel to force communication.

Sean-Der:
Oh nice, yea that works perfectly @LanderN! Would you mind making it non-blocking (in the edge case that channel never gets pulled) I would just buffer `workerTrigger` with 1

Merge from me then!

Sean-Der:
oh man I love this change, just tried it out locally and things feel SO much faster. 

This is going to make a lot of people happy, nice work!

LanderN:
I added the buffer (and somehow got the coverage down), is it okay like this?

PR #65, opened 2019-05-18
Expose CipherSuites via Config

### LINKED ISSUE #64: Allow CipherSuite to be chosen via config
This is required to support PSK

This also will allow us to test more suites

Sean-Der:
Allow users to configure what CipherSuites they wish
to use

Resolves #64


Sean-Der:
Nice, thank you @jinleileiking !

Issue #45, opened 2019-03-24
Support TLS_PSK_WITH_AES_128_CCM_8 for IoT usecases


daenney:
## Summary

I'd like to request inclusion of at least one [RFC 6655](https://tools.ietf.org/html/rfc6655) cipher suite, [`TLS_PSK_WITH_AES_128_CCM_8`](https://tools.ietf.org/html/rfc6655#section-4). This suite is often used for IoT products, most notably IKEA Tradfri uses it for CoAP over DTLS.

## Motivation

I'd like to use a Go native DTLS library in a number of my IoT projects. Some of these projects involve talking to (or exposing as) gateways that leverage this particular cipher suite as it's fairly common in IoT products. Currently [bocajim/dtls](https://github.com/bocajim/dtls) exists which implements just that cipher.

## Describe alternatives you've considered

I can always use this and the other DTLS library when I happen to need more than just `TLS_PSK_WITH_AES_128_CCM_8`. But I would much prefer to have one maintained library with all capabilities instead of being split between multiple packages from different maintainers.

## Checklist

* [x] Add CCM support #46 
* [x] Add PSK support
* [ ] Add `TLS_PSK_WITH_AES_128_CCM_8`

daenney:
@bocajim Would it be feasible to contribute `TLS_PSK_WITH_AES_128_CCM_8` to this library instead?

Sean-Der:
Hey @daenney 

That would be fantastic! I am open to any/all improvements to the library. We are missing features just because they were not needed for `pions/webrtc`. I don't think I will have a lot of time, but would love to help this land.

This change won't be trivial, but we should totally do it! We don't support PSK, so we will need to handle that. The actual encrypt/decrypt is easy to update we already support CBC and GCM so adding CCM shouldn't be bad!

Hopefully we don't have the break the API for existing users, but if it is needed I understand.

If you want to join our [Slack](https://pion.ly/slack) I can help and will try to explain a little bit here also.
----

The communication is split up into 'flights', I designed it this way so we can handle retransmissions/out of order packets.

each side accepts incoming messages, and when we get everything we want we start processing [code](https://github.com/pions/dtls/blob/master/server_handlers.go#L139)

each side just keeps re-sending until it gets to the next flight [code (https://github.com/pions/dtls/blob/master/server_handlers.go#L218)

Sean-Der:
I started digging into this, the good news is OpenSSL does support PSK! The bad news is no CCM :/ 

[...]

PR #66, opened 2019-05-22
Allow user to specify PSK

Sean-Der:
Add configuration option for user to pass PSK. When a user
passes a PSK then we only allow CipherSuites that do PSK.

If user passes a PSK and a certificate return an error.

Relates to #45

### DIFF
--- a/cipher_suite.go
+++ b/cipher_suite.go
@@ -88,7 +88,7 @@ func encodeCipherSuites(c []cipherSuite) []byte {
 	return out
 }
 
-func parseCipherSuites(userSelectedSuites []CipherSuiteID, psk []byte) ([]cipherSuite, error) {
+func parseCipherSuites(userSelectedSuites []CipherSuiteID, excludePSK, excludeNonPSK bool) ([]cipherSuite, error) {
 	cipherSuitesForIDs := func(ids []CipherSuiteID) ([]cipherSuite, error) {
 		cipherSuites := []cipherSuite{}
 		for _, id := range ids {
@@ -116,10 +116,11 @@ func parseCipherSuites(userSelectedSuites []CipherSuiteID, psk []byte) ([]cipher
 	}
 
 	for _, c := range cipherSuites {
-		if (psk != nil && c.isPSK()) || (psk == nil && !c.isPSK()) {
-			cipherSuites[i] = c
-			i++
+		if excludePSK && c.isPSK() || excludeNonPSK && !c.isPSK() {
+			continue
 		}
+		cipherSuites[i] = c
+		i++
 	}
 
 	cipherSuites = cipherSuites[:i]
--- a/cipher_suite.go
+++ b/cipher_suite.go
@@ -2,6 +2,7 @@ package dtls
 
 import (
 	"encoding/binary"
+	"fmt"
 	"hash"
 )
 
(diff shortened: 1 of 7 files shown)


cipher_suite.go
```diff
@@ -85,3 +87,45 @@ func encodeCipherSuites(c []cipherSuite) []byte {
...
+	} else {
+		cipherSuites = defaultCipherSuites()
+	}
+
+	for _, c := range cipherSuites {
+		if (psk != nil && c.isPSK()) || (psk == nil && !c.isPSK()) {
+			cipherSuites[i] = c
+			i++
+		}
+	}
```
daenney: Wouldn't this loop potentially result in cipher suites with and without PSK being considered valid? As in you could end up with a mix of PSK and server certificate. I'm not sure if that's an actual problem as long as both the PSK and the certs are provided but I don't think I've ever seen that in the wild.

LanderN: Since `psk` doesn't change during this loop, I don't think this will be the case?
Only PSK suites will be selected if `psk != nil` and vice-versa.

examples/listen/main.go
```diff
@@ -25,7 +25,7 @@ func main() {
 	//
 
 	// Prepare the configuration of the DTLS connection
-	config := &dtls.Config{Certificate: certificate, PrivateKey: privateKey}
+	config := &dtls.Config{Certificate: certificate, PrivateKey: privateKey, PSK: []byte{0x00, 0x01}}
```
backkem: Maybe this should be it's own example?

conn.go
```diff
@@ -320,19 +301,35 @@ func (c *Conn) internalSend(pkt *recordLayer, shouldEncrypt bool) {
...
-		err := c.handleIncomingPacket(p)
+	b := make([]byte, inboundBufferSize)
+	for {
+		i, err := c.nextConn.Read(b)
 		if err != nil {
-			return err
+			c.stopWithError(err)
+			return
+		} else if c.getConnErr() != nil {
+			return
```
LanderN: is `c.stopWithError` not needed here?

examples/listen/main.go
LanderN: Wouldn't this result in an error since both a certificate and a PSK are specified?

Sean-Der: Oops nice catch, accident on my part when testing :/

Sean-Der: Oops nice catch, accident on my part when testing :/

config.go
```diff
@@ -38,6 +37,10 @@ type Config struct {
 	// defaults to time.Second
 	FlightInterval time.Duration
 
+	// PSK sets the pre-shared key used by this DTLS connection
+	// If PSK is non-nil only PSK CipherSuites will be used
+	PSK []byte
```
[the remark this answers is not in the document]
Sean-Der: Oh good point thanks @igolaizola! Will update PR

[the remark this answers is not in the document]
Sean-Der: @igolaizola updated! I believe that API now gives us the flexibility to pass/receive the identity hint.

cipher_suite.go
Sean-Der: Updated! The intent of the function is to filter so you only get one type at the time (either PSK or non-PSK) 

I also assert earlier that you can't pass a certificate and a PSK.

I am trying to follow how OpenSSL acts here. If you pass a `psk` via the CLI they only do PSK suites.

conn.go
Sean-Der: Nope! In this case if `c.getConnErr()` returns an error the entire connection has already been shutdown (so we can just return)

PR #68, opened 2019-05-24
Fuzzing initial commit

ernado:
* Fix discovered errors
* Add makefile commands for fuzzing
* Commit corpus


ernado:
I've found the https://github.com/jedmao/eclint
We can try to use it

PR #70, opened 2019-05-24
Fuzz: part 2

ernado:
Fuzzing for a couple of hours discovered some more bugs.


Sean-Der:
LGTM! Mind squashing these down (just so crashers/fix are in the same commit)

thanks

ernado:
I've squashed into one commit.
Thought that it will be easier to read the changes if there are code commit + corpus commit.

handshake_message_server_key_exchange.go
```diff
@@ -53,23 +53,29 @@ func (h *handshakeMessageServerKeyExchange) Unmarshal(data []byte) error {
 
 	publicKeyLength := int(data[3])
 	offset := 4 + publicKeyLength
-	if len(data) <= publicKeyLength {
```
Sean-Der: I approved (so feel free to merge away!)

Style wise do you think it is better to do one assert for total length `if len(data) <= handshakeMessageServerKeyExchangeLen` or small asserts before each data access?

ernado: I think it is better to check as early as we can.
Probably we can use something like https://godoc.org/golang.org/x/crypto/cryptobyte to decrease probability of error like TLS library does.

PR #71, opened 2019-05-24
E2E: Decrease flakyness

ernado:
Try to pick port that is free.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	"errors"
 	"io"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


e2e/e2e_test.go
```diff
@@ -63,7 +80,8 @@ func simpleReadWrite(errChan chan error, outChan chan string, conn io.ReadWriteC
 */
 func TestPionE2ESimple(t *testing.T) {
 	expectedGoRoutineCount := runtime.NumGoroutine()
-	serverPort := baseServerPort
+
+	serverPort := pickPort(t)
```
Sean-Der: I would move this to line 77, instead of `serverPort++` do `serverPort == pickPort(t)`

Great fix though, thanks for grabbing :)

ernado:
@Sean-Der seems like I've found the real cause.

PR #76, opened 2019-06-30
Start TLS_PSK_WITH_AES_128_CCM8

Sean-Der:
@daenney This works 100% !`OpenSSL <-> Pion` and `Pion <-> Pion`

I am going on vacation on Tuesday, mind helping me catch corner cases etc...? Would love to get this merged and tagged :)


Sean-Der:
To use 
```
diff --git a/examples/listen/main.go b/examples/listen/main.go
index 37b8261..6310ce2 100644
--- a/examples/listen/main.go
+++ b/examples/listen/main.go
@@ -16,16 +16,13 @@ func main() {
        // Prepare the IP to connect to
        addr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 4444}

-       // Generate a certificate and private key to secure the connection
-       certificate, privateKey, genErr := dtls.GenerateSelfSigned()
-       util.Check(genErr)
-
-       //
-       // Everything below is the pion-DTLS API! Thanks for using it ❤️ .
-       //
-
-       config := &dtls.Config{Certificate: certificate, PrivateKey: privateKey}
+       config := &dtls.Config{
+               PSK: func(hint []byte) ([]byte, error) {
+                       return []byte{0xAB, 0xC1, 0x23}, nil
+               },
+               CipherSuites: []dtls.CipherSuiteID{dtls.TLS_PSK_WITH_AES_128_CCM8},
+       }

        // Connect to a DTLS server
        listener, err := dtls.Listen("udp", addr, config)
```

Against
```
openssl s_client -dtls1_2 -connect 127.0.0.1:4444 -psk abc123 -cipher PSK-AES128-CCM8
```

Sean-Der:
@daenney @igolaizola 

I am happy with this w/e I can get a review I would really appreciate it!

daenney:
This is awesome, thank you so much! I'm currently also on vacay but I have access to the IoT equipment at home, so I'll try to test this today.

README.md
```diff
@@ -68,6 +71,28 @@ Pion DTLS can connect to itself and OpenSSL.
...
+```
+
+#### OpenSSL
+```
+  // Use with examples/dial-psk/main.go
+  openssl s_server -dtls1_2 -accept 4444 -nocert -psk abc123 -cipher PSK-AES128-CCM8
+
+  // Use with examples/listen-psk/main.go
+  openssl s_client -dtls1_2 -connect 127.0.0.1:4444 -psk abc123 -cipher PSK-AES128-CCM8
+```
```
daenney: I think these two examples need to be swapped? dial-psk should be `s_client`, not `s_server`?

examples/dial-psk/main.go
```diff
@@ -0,0 +1,39 @@
...
+	// Prepare the IP to connect to
+	addr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 4444}
+
+	//
+	// Everything below is the pion-DTLS API! Thanks for using it ❤️.
+	//
+
+	// Prepare the configuration of the DTLS connection
+	config := &dtls.Config{
+		PSK: func(hint []byte) ([]byte, error) {
```
daenney: Since `hint` isn't used it might be nice to use the blank, `_`, identifier instead to make it more obvious to the reader?

examples/listen-psk/main.go
```diff
@@ -0,0 +1,54 @@
...
+	// Prepare the IP to connect to
+	addr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 4444}
+
+	//
+	// Everything below is the pion-DTLS API! Thanks for using it ❤️.
+	//
+
+	// Prepare the configuration of the DTLS connection
+	config := &dtls.Config{
+		PSK: func(hint []byte) ([]byte, error) {
```
daenney: Same thing as in dial-psk, use the blank identifier instead of `hint`?

conn.go
```diff
@@ -78,7 +78,9 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 		return nil, errNoConfigProvided
 	case nextConn == nil:
 		return nil, errNilNextConn
-	case config.Certificate != nil && (config.PSK != nil || config.PSKIdentityHint != nil):
+	case (config.PSK == nil) != (config.PSKIdentityHint == nil):
```
igolaizola: ¿PSKIdentityHint is always mandatory?

>  To help the client in selecting which identity to use, the server can provide a "PSK identity hint" in the ServerKeyExchange message.  If no hint is provided, the ServerKeyExchange message is omitted.

https://tools.ietf.org/html/rfc4279#section-2

daenney:
I'm happy to say I've just managed to ping my IKEA Tradfri gateway with this! So this has been validated in protoduction :smile:.

If anyone's looking on and would like to try it, here's the sample code:

```go
package main

import (
	"fmt"
	"net"

	"github.com/dustin/go-coap"
	"github.com/pion/dtls"
)

func main() {
	addr := &net.UDPAddr{IP: net.ParseIP("gateway-ip"), Port: 5684}
	config := &dtls.Config{
		PSK: func(_ []byte) ([]byte, error) {
			return []byte("psk-on-bottom-of-gateway"), nil
		},
		PSKIdentityHint: []byte("Client_identity"), # For Tradfri Gateway the IdentityHint MUST be Client_identity
		CipherSuites:    []dtls.CipherSuiteID{dtls.TLS_PSK_WITH_AES_128_CCM8},
	}

	dtlsConn, err := dtls.Dial("udp", addr, config)
	if err != nil {
		panic(err)
	}
	defer dtlsConn.Close()

	req := coap.Message{
		Type:      coap.Confirmable,
		Code:      coap.GET,
		MessageID: 1,
		Token:     []byte{},
		Payload:   []byte{},
	} # This is a CoAP Ping, empty confirmable message

	data, err := req.MarshalBinary()
	if err != nil {
		panic(err)
	}

	_, err = dtlsConn.Write(data)
	if err != nil {
		panic(err)
	}

	resp := make([]byte, 2048)
	dtlsConn.Read(resp)
	msg, err := coap.ParseMessage(resp)
	if err != nil {
		panic(err)
	}
	fmt.Println(msg.MessageID)
	fmt.Println(msg.Type)
	fmt.Println(msg.Code)
	fmt.Println(msg.Token)
	fmt.Println(string(msg.Payload))

}
```
It'll output:

```
1
Acknowledgement
NotFound
[]
Not Found
```

conn.go
Sean-Der: oh nice catch, do you know if `ClientKeyExchange` is mandatory?

Sean-Der:
@daenney that is awesome! So glad it works :)

Thanks for doing half the work, really glad we did this. I am hoping we can get `pion/dtls` good enough so it can get wide adoption/get the community around one Go implementation.

examples/dial-psk/main.go
Sean-Der: Oh yea good point, what do you think of https://github.com/pion/dtls/pull/76/commits/0a9c0ced8977b98ce9037943e11b4d089f696ca0

The `hint` is used, and we print it to try and help give the user a better understanding of its purpose.

examples/listen-psk/main.go
Sean-Der: Addressed above https://github.com/pion/dtls/pull/76/commits/0a9c0ced8977b98ce9037943e11b4d089f696ca0

examples/dial-psk/main.go
daenney: That works too! :+1:

conn.go
daenney: Did a quick search for `ClientKeyExchange` through the RFC, no mention of it being optional in certain cases. Though mind you, there is also this:

>  When these ciphersuites are used, the ServerKeyExchange and
>  ClientKeyExchange messages also include the Diffie-Hellman
>  parameters.  The PSK identity and identity hint fields have the same
>  meaning as in the previous section (note that the ServerKeyExchange
>  message is always sent, even if no PSK identity hint is provided).
https://tools.ietf.org/html/rfc4279#section-3

That's only for DHE_PSK though, which is not what we're doing here.

Sean-Der: @igolaizola @daenney fixed https://github.com/pion/dtls/pull/76/commits/42310ae98b4c0c9a71753bd381a9421df11cde1a!

`DHE_PSK` probably will not be easy to add :/ currently our Handshake parsing is stateless which makes it a little harder to parse things. Right now it is easy to distinguish DHE vs PSK, but the more we add we will have to think about refactoring.

Personally I would love to see someone majorly refactor stuff. It would be great if someone went in with fresh eyes and moved stuff around. I would like to see a bunch of small packages in `internal` instead of the bloated main directory (my 2 cents)

igolaizola: `ClientKeyExchange` is mandatory and `ServerKeyExchange` optional depending if a identity hint has been provided (assuming DHE is not supported).

Sean-Der: `DHE` isn't supported (yet)

I added a test to assert ^ https://github.com/pion/dtls/pull/76/files#diff-f9df1a974544ddcc7ccada9d10f705a4R185

It does a full test with ClientIdentity nil and non-nil (and we assert the values match up)

PR #81, opened 2019-07-23
Verify server and client certificates

igolaizola:
Description:
- Verify certificates using CAs `Config.RootCAs`, current time and an optional callback `Config.VerifyPeerCertificate`
- Verification for server certificates is enabled by default and for client certificates based on `Config.ClientAuth (VerifyClientCertIfGiven, RequireAndVerifyClientCert)`
- Verification can be disabled by `Config.InsecureSkipVerify`
- Config parameters are inspired on `crypto/tls`

TODO: Verify server name

### DIFF
--- a/client_handlers.go
+++ b/client_handlers.go
@@ -2,9 +2,7 @@ package dtls
 
 import (
 	"bytes"
-	"crypto/x509"
 	"fmt"
-	"time"
 )
 
 func clientHandshakeHandler(c *Conn) error {
@@ -52,14 +50,7 @@ func clientHandshakeHandler(c *Conn) error {
 				if err := verifyKeySignature(expectedHash, h.signature, h.hashAlgorithm, c.state.remoteCertificate); err != nil {
 					return err
 				}
-				opts := x509.VerifyOptions{
-					Roots:       c.rootCAs,
-					CurrentTime: time.Now(),
-					// TODO: Add server name validation
-					// DNSName:       c.serverName,
-					Intermediates: x509.NewCertPool(),
-				}
-				if _, err := c.state.remoteCertificate.Verify(opts); err != nil {
+				if err := verifyServerCert(c.state.remoteCertificate, c.rootCAs, c.serverName); err != nil {
 					return err
 				}
 			}
--- a/client_handlers.go
+++ b/client_handlers.go
@@ -2,7 +2,9 @@ package dtls
 
 import (
 	"bytes"
+	"crypto/x509"
 	"fmt"
+	"time"
 )
 
 func clientHandshakeHandler(c *Conn) error {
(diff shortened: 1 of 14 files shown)


Sean-Der:
woohoo awesome work @igolaizola 

I will look at this right after work, but feature wise it is a merge from me :)

client_handlers.go
```diff
@@ -45,9 +47,26 @@ func clientHandshakeHandler(c *Conn) error {
...
 		if c.localPSKCallback == nil {
-			expectedHash := valueKeySignature(clientRandom, serverRandom, h.publicKey, h.namedCurve, h.hashAlgorithm)
-			if err := verifyKeySignature(expectedHash, h.signature, h.hashAlgorithm, c.state.remoteCertificate); err != nil {
-				return err
+			if !c.insecureSkipVerify {
+				expectedHash := valueKeySignature(clientRandom, serverRandom, h.publicKey, h.namedCurve, h.hashAlgorithm)
+				if err := verifyKeySignature(expectedHash, h.signature, h.hashAlgorithm, c.state.remoteCertificate); err != nil {
+					return err
+				}
+				opts := x509.VerifyOptions{
```
Sean-Der: I would move the actual verification into `crypto.go`, just so it is one easy place to check out

Sean-Der:
@igolaizola ping me again when server name validation is good. happy to help in any way I can!

Thanks for grabbing this!

client_handlers.go
igolaizola: Done. I've created `verifyClientCert` and `verifyServerCert` at `crypto.go`.

igolaizola:
> @igolaizola ping me again when server name validation is good. happy to help in any way I can!

I've added ServerName validation.

In [crypto/tls](https://golang.org/pkg/crypto/tls/#Config) `Config.ServerName` is also used for SNI extension in the ClientHello, but the implementation of that feature should go in a different PR.

igolaizola:
Important note about this PR: configurations that were working previously (selfsigned certs) may no longer work unless `InsecureSkipVerify` param is enabled. From a security point of view I think this breaking change is necessary.

daenney:
I have a PR in mind to add a proper CHANGELOG, both to the repo and the release we tag, where we should mention these things.

Sean-Der:
ACK! As soon as this lands we can make a new tag and I can update pion/webrtc that is my big worry :)

Might be nice to look at godoc and reach out to users!

daenney:
Good point. Took a look, it's pion/webrtc and go-ocf/coap. I know the OCF folks, they recently integrated pion/dtls at my request to support coaps over UDP. I don't think anyone is really using it yet or would be surprised by having to set `InsecureSkipVerify` but I'll raise the issue with them. Though even when using go-ocf/coap you're expected to pass in your own pion/dtls.Config.

igolaizola:
Should we merge this, or wait until receive some feedback?

Sean-Der:
@igolaizola I am going to open a PR using your branch against `pion/webrtc` if that is green (will know in 10 mins) 👍 on merge from me!

Sean-Der:
@igolaizola created PR! Mind approving https://github.com/pion/webrtc/pull/757 ?

When that is green I will merge both PRs so we don't break master at all (and will bump version numbers!)

I am not going to do a new major version. If we do `github.com/pion/dtls/v2` it will break a lot of people who don't use modules. I sort of regret doing it with `pion/webrtc` since tooling is still bad. They work great for us, but outside users really get hung up on it.

thanks!

Sean-Der:
Great work @igolaizola 

You and @daenney have really taken DTLS to the next level recently, so great to see how good this project is getting.

igolaizola:
@Sean-Der @daenney thanks for dealing with the affected projects.

PR #92, opened 2019-08-14
Less allocations for encrypt/decrypt

hugoArregui:
By simply reserving memory before append

### DIFF
--- a/crypto_gcm.go
+++ b/crypto_gcm.go
@@ -9,6 +9,7 @@ import (
 )
 
 const cryptoGCMTagLength = 16
+const cryptoGCMNonceLength = 12
 
 // State needed to handle encrypted input/output
 type cryptoGCM struct {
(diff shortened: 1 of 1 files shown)


crypto_gcm.go
```diff
@@ -47,20 +47,22 @@ func (c *cryptoGCM) encrypt(pkt *recordLayer, raw []byte) ([]byte, error) {
 	payload := raw[recordLayerHeaderSize:]
 	raw = raw[:recordLayerHeaderSize]
 
-	nonce := append(append([]byte{}, c.localWriteIV[:4]...), make([]byte, 8)...)
+	nonce := make([]byte, 12)
```
Sean-Der: Mind making the 12 a constant? Beyond that great work! You are killing :)

hugoArregui: done!

PR #93, opened 2019-08-16
Implement "Extended Master Secret" extension (RFC7626)

### LINKED ISSUE #86: Extended Master Secret support [RFC7627]
## Summary

We wish to add support for the "Extended Master Secret" extension as recommended by [RFC7627](https://tools.ietf.org/html/rfc7627)

## Motivation

Some dTLS clients use the "Extended Master Secret" extension in the `Client Hello` message. Per the RFC7627 spec, if the client requests "Extended Master Secret" the server must also specify supporting "Extended Master Secret" in the `Server Hello` message. If the server does not, then it is recommended that the client must close the connection and some clients do in fact do this.

What this means in effect, that some dTLS are unable to communicate with a `pion/dtls` server  without this extension support.

It would be possible to disable "Extended Master Secret" extension in these clients, but increases security risks as describe in RFC7627.

RyanGordon:
#### Description
This change implements RFC7627 "Extended Master Secret" extension with
optional enforcement by default

#### Reference issue
Fixes #86


config.go
```diff
@@ -73,6 +77,10 @@ type Config struct {
 	ServerName string
 
 	LoggerFactory logging.LoggerFactory
+
+	// ConnectTimeoutMs is the timeout threshold for new connection handshakes
+	// to complete (default is no timeout)
+	ConnectTimeout time.Duration
```
Sean-Der: This looks to be unused

server_handlers.go
```diff
@@ -56,6 +63,7 @@ func serverHandshakeHandler(c *Conn) error {
 				}
 			}
 
+			c.log.Trace("Moving to flight2")
```
Sean-Der: Can you generalize this log message, just do a `trace` inside `flight.go` itself

client_handlers.go
```diff
@@ -76,14 +94,20 @@ func clientHandshakeHandler(c *Conn) error {
...
 					if !ok {
 						return fmt.Errorf("Server responded with SRTP Profile we do not support")
 					}
 					c.state.srtpProtectionProfile = profile
+				case *extensionUseExtendedMasterSecret:
+					c.state.extendedMasterSecret = true
 				}
 			}
+			if c.requireExtendedMasterSecret && !c.state.extendedMasterSecret {
+				return fmt.Errorf("Client required Extended Master Secret extension, but server does not support it")
```
Sean-Der: Would it be worth offering some internal switch so we can disable EMS? It would be nice to assert that these error conditions are hit/returned.

server_handlers.go
RyanGordon: Good idea yes :+1:

client_handlers.go
RyanGordon: Yes that's a good idea, maybe a config option so the user can disable it if wanted?

config.go
RyanGordon: Whoops, fixed!

client_handlers.go
RyanGordon: I switched it to an enum with three options:
Request (default)
Required
Disabled

RyanGordon: And tests added!

cipher_suite_tls_ecdhe_ecdsa_with_aes_128_gcm_sha256.go
```diff
@@ -30,6 +31,10 @@ func (c cipherSuiteTLSEcdheEcdsaWithAes128GcmSha256) isPSK() bool {
 	return false
 }
 
+func (c cipherSuiteTLSEcdheEcdsaWithAes128GcmSha256) isInitialized() bool {
+	return c.initialized
```
Sean-Der: Would it be possible to do `return c.gcm != nil` ?

cipher_suite_tls_ecdhe_ecdsa_with_aes_256_cbc_sha.go
```diff
@@ -30,6 +31,10 @@ func (c cipherSuiteTLSEcdheEcdsaWithAes256CbcSha) isPSK() bool {
 	return false
 }
 
+func (c cipherSuiteTLSEcdheEcdsaWithAes256CbcSha) isInitialized() bool {
+	return c.initialized
```
Sean-Der: `return c.cbc != nil`

cipher_suite_tls_psk_with_aes_128_ccm8.go
```diff
@@ -30,6 +31,10 @@ func (c cipherSuiteTLSPskWithAes128Ccm8) isPSK() bool {
 	return true
 }
 
+func (c cipherSuiteTLSPskWithAes128Ccm8) isInitialized() bool {
+	return c.initialized
```
Sean-Der: `return c.ccm != nil`

bench_test.go
```diff
@@ -15,16 +15,16 @@ func TestSimpleReadWrite(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	config := &Config{
```
Sean-Der: Is this to fix a race inside `pion/dtls`? 

I would rather we copy values, I think our raciness would be a surprise to users.

handshake_cache.go
```diff
@@ -78,3 +78,38 @@ func (h *handshakeCache) pullAndMerge(rules ...handshakeCachePullRule) []byte {
...
+	}
+
+	hash := hf()
+	_, err := hash.Write(merged)
+	if err != nil {
+		return []byte{}, err
+	}
+
+	sum := hash.Sum(nil)
+	return sum, nil
```
Sean-Der: return `hash.Sum(nil), nil` is fine IMO. Just less for the reader to consider when reading the code!

handshake_cache.go
```diff
@@ -78,3 +78,38 @@ func (h *handshakeCache) pullAndMerge(rules ...handshakeCachePullRule) []byte {
...
+		if p == nil {
+			continue
+		}
+
+		merged = append(merged, p.data...)
+	}
+
+	hash := hf()
+	_, err := hash.Write(merged)
+	if err != nil {
```
Sean-Der: `if _, err := hash.Write(merged); err != nil {`

Sorry this might be style, but it helps me reading the code. Just keeping the scope of `err` down prevents us from accidentally letting it effect any new code later.

server_handlers.go
```diff
@@ -43,9 +43,17 @@ func serverHandshakeHandler(c *Conn) (*alert, error) {
...
 					c.state.srtpProtectionProfile = profile
+				case *extensionUseExtendedMasterSecret:
+					if c.extendedMasterSecret != DisableExtendedMasterSecret {
+						c.state.extendedMasterSecret = true
+					}
 				}
 			}
 
+			if c.extendedMasterSecret == RequireExtendedMasterSecret && !c.state.extendedMasterSecret {
+				return &alert{alertLevelFatal, alertInsufficientSecurity}, fmt.Errorf("Server requires the Extended Master Secret extension, but the client does not support it")
```
Sean-Der: `errors.go`

client_handlers.go
```diff
@@ -76,14 +94,22 @@ func clientHandshakeHandler(c *Conn) (*alert, error) {
 
 		case *handshakeMessageServerHello:
 			for _, extension := range h.extensions {
-				if e, ok := extension.(*extensionUseSRTP); ok {
+				switch e := extension.(type) {
+				case *extensionUseSRTP:
 					profile, ok := findMatchingSRTPProfile(e.protectionProfiles, c.localSRTPProtectionProfiles)
 					if !ok {
 						return &alert{alertLevelFatal, alertIllegalParameter}, fmt.Errorf("Server responded with SRTP Profile we do not support")
```
Sean-Der: `errors.go`

client_handlers.go
```diff
@@ -76,14 +94,22 @@ func clientHandshakeHandler(c *Conn) (*alert, error) {
...
 					}
 					c.state.srtpProtectionProfile = profile
+				case *extensionUseExtendedMasterSecret:
+					if c.extendedMasterSecret != DisableExtendedMasterSecret {
+						c.state.extendedMasterSecret = true
+					}
 				}
 			}
+			if c.extendedMasterSecret == RequireExtendedMasterSecret && !c.state.extendedMasterSecret {
+				return &alert{alertLevelFatal, alertInsufficientSecurity}, fmt.Errorf("Client required Extended Master Secret extension, but server does not support it")
```
Sean-Der: `errors.go`

client_handlers.go
```diff
@@ -76,14 +94,22 @@ func clientHandshakeHandler(c *Conn) (*alert, error) {
...
+			}
 			if len(c.localSRTPProtectionProfiles) > 0 && c.state.srtpProtectionProfile == 0 {
 				return &alert{alertLevelFatal, alertInsufficientSecurity}, fmt.Errorf("SRTP support was requested but server did not respond with use_srtp extension")
```
Sean-Der: `errors.go`

Sean-Der:
@RyanGordon So not using `errors.go` isn't a huge deal, but is just nice to see the full range of errors and give us a little type safety.

In the future we may export these, and would be great if we can just upcase them all when we get  to that point. I think that could be a huge advantage over OpenSSL, actually let people really understand what happened and why.

client_handlers.go
```diff
@@ -5,64 +5,82 @@ import (
...
+		return &alert{alertLevelFatal, alertInternalError}, err
+	}
+	serverRandom, err := c.state.remoteRandom.Marshal()
+	if err != nil {
+		return &alert{alertLevelFatal, alertInternalError}, err
+	}
+
+	if c.state.extendedMasterSecret {
+		var sessionHash []byte
+		sessionHash, err = c.handshakeCache.sessionHash(c.state.cipherSuite.hashFunc())
```
Sean-Der: `sessionHash, err := c.handshakeCache.sessionHash(c.state.cipherSuite.hashFunc())`

Sean-Der: so you don't need to do `var sessionHash []byte`

client_handlers.go
```diff
@@ -5,64 +5,82 @@ import (
...
-
-			preMasterSecret = prfPSKPreMasterSecret(psk)
+	if err := c.state.cipherSuite.init(c.state.masterSecret, clientRandom, serverRandom /* isClient */, true); err != nil {
+		return &alert{alertLevelFatal, alertInternalError}, err
+	}
 
-		} else {
-			if c.localKeypair, err = generateKeypair(h.namedCurve); err != nil {
-				return &alert{alertLevelFatal, alertIllegalParameter}, err
+	if c.localPSKCallback == nil {
```
Sean-Der: Is it possible for `insecureSkipVerify` or `verifyPeerCertificate` to be set when dealing with PSK? Hopefully we have error handling higher up (if we don't mind ticketing!) 

I am tempted to say drop the `c.localPSKCallback` check. I like that this reads as a long chain of operations. I don't want us to assume internal state is always hostile (otherwise we might miss things sometime and trip up)

This does disagree with `defensive coding` so I am a little torn

state.go
```diff
@@ -19,6 +19,10 @@ type State struct {
 	remoteCertificate     *x509.Certificate
 
 	isClient bool
+
+	serverKeyExchange    *handshakeMessageServerKeyExchange
```
Sean-Der: Have you ever read anything about what may/may be kept in memory?

`crypto/tls` itself only keeps the `masterSecret []byte` so I think we should push back on this

flight.go
```diff
@@ -90,6 +101,7 @@ func (f *flight) get() flightVal {
 
 func (f *flight) set(val flightVal) {
 	f.Lock()
+	f.log.Tracef("[handshake] Moving from %s to %s", f.val.String(), val.String())
```
Sean-Der: 👍 love it

examples/dial/main.go
```diff
@@ -21,7 +21,12 @@ func main() {
 	//
 
 	// Prepare the configuration of the DTLS connection
-	config := &dtls.Config{Certificate: certificate, PrivateKey: privateKey, InsecureSkipVerify: true}
+	config := &dtls.Config{
+		Certificate:          certificate,
+		PrivateKey:           privateKey,
+		InsecureSkipVerify:   true,
+		ExtendedMasterSecret: dtls.RequestExtendedMasterSecret,
```
Sean-Der: Do you think it is worth surfacing to the user the default value?

 I think it might even be ok to do `Require` and put a comment that you can/should lower to `Request` to support Clients/Servers that don't support it yet.

conn_test.go
```diff
@@ -560,6 +560,139 @@ func TestClientCertificate(t *testing.T) {
 	}
 }
 
+func TestExtendedMasterSecret(t *testing.T) {
+	t.Parallel()
```
Sean-Der: Did not know about this! Should we set this for all our tests? How much of an improvement do you see locally?

server_handlers.go
```diff
@@ -119,9 +127,22 @@ func serverHandshakeHandler(c *Conn) (*alert, error) {
 				}
 			}
 
-			c.state.masterSecret, err = prfMasterSecret(preMasterSecret, clientRandom, serverRandom, c.state.cipherSuite.hashFunc())
-			if err != nil {
-				return &alert{alertLevelFatal, alertInternalError}, err
+			if c.state.extendedMasterSecret {
+				var sessionHash []byte
+				sessionHash, err = c.handshakeCache.sessionHash(c.state.cipherSuite.hashFunc())
```
Sean-Der: `sessionHash, err := c.handshakeCache.sessionHash(c.state.cipherSuite.hashFunc())`

cipher_suite_tls_ecdhe_ecdsa_with_aes_128_gcm_sha256.go
RyanGordon: Yes, good idea!

bench_test.go
RyanGordon: Yes! This test sometimes fails when the race condition is detected by go. Though after some thought, I am going to pull this out into a separate PR so it doesn't block this PR

conn_test.go
RyanGordon: I didn't either, I actually just took it from the `TestClientCertificate` test 😅

server_handlers.go
RyanGordon: Fixed!

client_handlers.go
RyanGordon: Fixed!

examples/dial/main.go
RyanGordon: Hmm yes I'm happy with making Require the default; but that means it can be somewhat of a breaking change

RyanGordon: Misunderstood - this is only for examples, fixed!

state.go
RyanGordon: So two problems that I ran into:
1) In order to calculate the `extendedMasterSecret` hash, you have to keep a binary "log" of the handshakes, until you calculate the hash
2) For the client, the last part of the handshake is sending out the `clientKeyExchange` so we have to wait for the `initalizeCipherSuite` to run at that point which then calculates the `extendedMasterSecret` and uses the serverKeyExchange  (specifically `publicKey`, `namedCurve`, `signature` and `hashAlgorithm` from it) to validate the key signature (if not disabled)

I'm happy to see if there are ways to more directly scrub it from memory but I am struggling to see how I (or another library?) could not keep it around

Sean-Der: Can we pull the `*handshakeMessageServerKeyExchange` from the `*handshakeCache`?

state.go
```diff
@@ -32,6 +36,7 @@ type serializedState struct {
 	SRTPProtectionProfile uint16
 	RemoteCertificate     []byte
 	IsClient              bool
+	ExtendedMasterSecret  bool
```
Sean-Der: Do we need to serialize this? `ExtendedMasterSecret` doesn't have any effect post-handshake right?

I think the serialize/de-serialize is only used post-handshake. Sorry for being so nitty, just don't want people to depend on any state in future features by accident that we don't actually need to provide.

I already am planning on ripping more out of `Conn` when I get a chance :)

Sean-Der:
Sorry I realize a lot of issues I have with code... is actually mine -_-

If we can bring down the stored state to the absolute minimum (less bugs and random things changing the world) this is an approve from me!

I will go back and send PRs your way to make things better.

state.go
daenney: Even the handshake cache would still be an unprotected memory region right?

We could consider introducing a dependency on https://godoc.org/github.com/awnumar/memguard, for creating a secure enclave where we store sensitive information. It's a little more extreme than just zeroing out a buffer but it goes to fairly great lengths to help you keep a memory area protected.

RyanGordon: @Sean-Der Yes, we could pull it from the handshake cache but I would have to unmarshal it again which would be a bit extra computation. I don't have a strong feeling either way

@daenney Yes, I think it's definitely worthwhile to have a discussion on how we could better protect memory in these cases. I opened up an issue here so we can have a fuller discussion on it: https://github.com/pion/dtls/issues/97

state.go
RyanGordon: Ahh yes makes sense, I will move this out

RyanGordon: Fixed!

server_handlers.go
RyanGordon: Actually lint fails with it this way:

```
client_handlers.go:19:16: declaration of "err" shadows declaration at client_handlers.go:9 (govet)
		sessionHash, err := c.handshakeCache.sessionHash(c.state.cipherSuite.hashFunc())
		             ^
server_handlers.go:131:18: declaration of "err" shadows declaration at server_handlers.go:106 (govet)
				sessionHash, err := c.handshakeCache.sessionHash(c.state.cipherSuite.hashFunc())
```

client_handlers.go
Sean-Der: Ignore this! Not code you added, I will fix myself!

state.go
Sean-Der: My preference is double computation. Until we measure we don’t know how expensive something is.

But the more state we have the more difficult things become to keep in your head/and we have to guard and more nil checks (that’s just like my opinion though man)

Sean-Der: @daenney that’s super interesting I had no idea about memguard. 

I will play with it/understand the trade offs! I don’t have a security background/I am pretty unaware of what the community is doing :/

client_handlers.go
```diff
@@ -332,6 +354,27 @@ func clientFlightHandler(c *Conn) (bool, *alert, error) {
...
+			if err != nil {
+				return false, alertPtr, err
+			}
+
+			if alertPtr, err := initalizeCipherSuite(c, &handshakeMessageServerKeyExchange{}); err != nil {
+				return false, alertPtr, err
+			}
+		}
+
+		// handshakeMessageServerKeyExchange is optional for PSK
```
Sean-Der: Move comment up/remove conditional?

Sean-Der:
@RyanGordon this LGTM!

I am approving, feel free to merge w/e you are happy with it and all the tests pass :) after it lands I am going to do do a cleanup pass (I will submit it!). When reading the code during the review I saw a lot of things that I did that make me unhappy :(

PR #94, opened 2019-08-19
Adding connection timeout

### LINKED ISSUE #13: Implement re-try backoff (and eventual timeouts)
The handshake takes quite some time now. Maybe we can trigger the TimerThread when changing flight?

### LINKED ISSUE #89: DTLS with PSK stuck at handshake progress
Once send packet over DTLS with wrong PSK, the DTLS server will be stuck at handshaking and cannot handle other request anymore event use right PSK.

CipherSuites: TLS_PSK_WITH_AES_128_CCM_8

RyanGordon:
#### Description
This change adds a connection timeout (effectively a handshake timeout) that allows handshakes that never complete for any reason to be closed and not take up resources or retry indefinitely. Typically this logic should not be hit unless the client or server really does not complete the handshake in time.

This technically addressed #89 because once the timeout occurs, processing will be able to occur again, however further improvements are needed so that a single handshake does not block the progress of other handshakes, even if any of them get stuck or error out. This additional improvements will be covered in subsequent PRs

#### Reference issue
Partially fixes #89
Partially fixes #13

### DIFF
--- a/config.go
+++ b/config.go
@@ -73,6 +73,17 @@ type Config struct {
 	ServerName string
 
 	LoggerFactory logging.LoggerFactory
+
+	// ConnectTimeout is the timeout threshold for new connection handshakes
+	// to complete (default is 30 seconds)
+	ConnectTimeout *time.Duration
+}
+
+const defaultConnectTimeout = 30 * time.Second
+
+// ConnectTimeoutOption simply provides a wrapper for creating a *time.Duration
+func ConnectTimeoutOption(timeout time.Duration) *time.Duration {
+	return &timeout
 }
 
 // PSKCallback is called once we have the remote's PSKIdentityHint.
--- a/conn.go
+++ b/conn.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rand"
 	"crypto/x509"
 	"fmt"
+	"math"
 	"net"
 	"sync"
 	"sync/atomic"
@@ -41,6 +42,8 @@ type Conn struct {
 	workerTicker   *time.Ticker
 
 	state State // Internal state
+
+	connectTimeout time.Duration
 
 	remoteRequestedCertificate bool // Did we get a CertificateRequest
 
@@ -111,6 +114,15 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 		loggerFactory = logging.NewDefaultLoggerFactory()
 	}
 
+	connectTimeout := defaultConnectTimeout
+	if config.ConnectTimeout != nil {
+		connectTimeout = *config.ConnectTimeout
+	}
+
+	if connectTimeout <= 0 {
+		connectTimeout = math.MaxInt64 * time.Nanosecond
+	}
+
 	c := &Conn{
 		nextConn:                    nextConn,
 		currFlight:                  newFlight(isClient),
@@ -118,6 +130,7 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 		handshakeCache:              newHandshakeCache(),
 		handshakeMessageHandler:     handshakeMessageHandler,
 		flightHandler:               flightHandler,
+		connectTimeout:              connectTimeout,
 		localCertificate:            config.Certificate,
 		localPrivateKey:             config.PrivateKey,
 		clientAuth:                  config.ClientAuth,
@@ -170,8 +183,13 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 	// Handle inbound
 	go c.inboundLoop()
 
-	<-c.handshakeCompletedSignal
-	err = c.getConnErr()
+	select {
+	case <-c.handshakeCompletedSignal:
+		err = c.getConnErr()
+	case <-time.After(c.connectTimeout):
+		err = errConnectTimeout
+	}
+
 	if err == nil {
 		c.setHandshakeCompletedSuccessfully()
 	}
--- a/conn_test.go
+++ b/conn_test.go
@@ -311,6 +311,30 @@ func TestPSKHintFail(t *testing.T) {
 	}
 }
 
+func TestClientTimeout(t *testing.T) {
+	// Limit runtime in case of deadlocks
+	lim := test.TimeOut(time.Second * 20)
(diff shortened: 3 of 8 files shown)


config.go
```diff
@@ -73,6 +73,10 @@ type Config struct {
 	ServerName string
 
 	LoggerFactory logging.LoggerFactory
+
+	// ConnectTimeoutMs is the timeout threshold for new connection handshakes
+	// to complete (default is no timeout)
```
daenney: I don't think the timeout should be no timeout by default. I realise that's the current behaviour but if we're going to introduce a way to control it I'd also prefer we set a value, 30s for example. No timeout feels like just leaving a gun for someone to shoot themselves with. If people really want no timeout they can explicitly set the timeout to 0.

RyanGordon: That works for me but how do you think we should default it? Golang by default initializes it to zero so its difficult to tell if its explicitly set to zero or implicitly initialized to zero. I could turn it into a pointer with nil as default or move config to a constructor based system

daenney: I don't think the -1/0 is what we should do. So far in all Go libraries I've seen dealing with networks, the value 0 w.r.t timeouts has meant no timeout. We should stick to that, -1 would be surprising.

I would be inclined to make it a pointer but it's annoying when using the library since you now need a `IntP` helper or something to give you an integer pointer.

@Sean-Der @igolaizola thoughts?

RyanGordon: I'm thinking `*time.Duration` or `*int` might be the least of all evils/road-blocky. We can have a little function for people to use:

```
func ConnectTimeoutOption(time.Duration) *time.Duration
```

Any of the other options (variadic functions, constructors) are larger breaking changes

daenney: Ah yes, that works for me.

RyanGordon: Okay updated! :+1:

igolaizola: In my opinion no timeout by default doesn't seems like a bad idea.
- A default value of 30 sec might not fit every user.
- Using a `time.Duration` is much cleaner than `*time.Duration`.

Some examples that use a `Timeout time.Duration` property and "no timeout" as default:
- `http.Client` of the standard library https://golang.org/pkg/net/http/#Client 
- `http2.Server` https://godoc.org/golang.org/x/net/http2#Server

conn.go
```diff
@@ -169,7 +177,16 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 	// Handle inbound
 	go c.inboundLoop()
 
-	<-c.handshakeCompleted
+	if c.connectTimeout.Seconds() > 0.0 {
+		select {
+		case <-c.handshakeCompleted:
+		case <-time.After(c.connectTimeout):
+			c.log.Errorf("%s", errConnectTimeout)
```
Sean-Der: (annoying nit) I would drop the log line. I try to only log when we have things happening async we can't handle.

conn.go
```diff
@@ -169,7 +177,16 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 	// Handle inbound
 	go c.inboundLoop()
 
-	<-c.handshakeCompleted
+	if c.connectTimeout.Seconds() > 0.0 {
```
Sean-Der: I would do `math.MaxInt64` if the user didn't specify a timeout. That way we can only have one code path.

Might be worth confirming it isn't expensive, but I always worry the complexity could hide bugs.

config.go
Sean-Der: Fantastic feature @RyanGordon, we are going to do a little bike shedding. But I think this is for sure a merge!

----
Oh nice, thanks for the data @igolaizola. I would say copy stdlib w/e we can. Feels like the source of truth in all cases :)

Does anyone know why `crypto/tls` doesn't have a timeout? 

I have never had a case where I needed timeouts with crypto/tls but it looks like Deadline is the way to do it. Should we copy them and use `Deadline` to cover this (instead of timeout?) then the deadline would apply to every read as well?

daenney: The no timeout as a default is dangerous. It makes this library a danger to use in production if people aren't aware of it. http.Server, http.Client etc not having a timeout is b/c nobody considered it before it landed in Go 1 and they can't change it now. It's why not safe to use http.Server and why there's articles like https://blog.cloudflare.com/exposing-go-on-the-internet/ that try to teach people to set timeouts.

Sean-Der: Thanks for the push back @daenney, you have convinced me :) 

My vote is 30 seconds as well, we should also update the examples to maybe be 30 seconds explicitly so people expect it?

igolaizola: @daenney you've also convinced me. Thanks for the cloudflare's article link btw, it's very interesting.

The only way to avoid `*time.Duration` will be using `MaxInt` for "no timeout" and reinterpreting the zero value as default, but it might be confusing as you've already said.

conn.go
RyanGordon: Fixed :+1:

conn.go
RyanGordon: Fixed :+1:

config.go
RyanGordon: Kept at 30 seconds and updated examples :+1:

conn.go
```diff
@@ -111,13 +114,22 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 		loggerFactory = logging.NewDefaultLoggerFactory()
 	}
 
+	connectTimeout := defaultConnectTimeout
+	if config.ConnectTimeout != nil {
+		connectTimeout = *config.ConnectTimeout
+		if connectTimeout <= 0 {
+			connectTimeout = math.MaxInt64 * time.Nanosecond
+		}
```
daenney: I would move the nested `if` out of it and one up, so you end up with:

```
connectTimeout := defaultConnectTimeout
if config.ConnectTimeout != nil {
	connectTimeout = *config.ConnectTimeout
}
if connectTimeout <= 0 {
    connectTimeout = math.MaxInt64 * time.Nanosecond
}
```

This keeps the flow a bit simpler while achieving the same thing.

RyanGordon: Yep, that does feel cleaner, thanks! Fixed up

daenney:
:shipit:

PR #95, opened 2019-08-20
Improve alerting during handshake

### LINKED ISSUE #5: Send alerts + errors on invalid input

### LINKED ISSUE #89: DTLS with PSK stuck at handshake progress

RyanGordon:
#### Description
Previously, alerts would not be sent during handshake. This change adds unencrypted alerting during the handshake and encrypted alerts once the handshake completes. This allows the client to know to close and not rely on a given timeout.

Separately, if the connection was closed in the middle of a handshake (unlikely), a panic would occur because alerts were previously only sent as encrypted alerts even if the cipher suite was not yet initialized, leading to a panic.

This fixes issue #89 because the server will now properly send an alert to the client (or vice versa) indicating that there was a problem and that the connection should be closed immediately.

If there are any missing cases, or if the alert gets dropped during flight, then the connection timeout is the safety net (https://github.com/pion/dtls/pull/94).

#### Reference issue
Fixes #89
Fixes #5

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -28,8 +28,8 @@ var invalidKeyingLabels = map[string]bool{
 	"key expansion":   true,
 }
 
-type handshakeMessageHandler func(*Conn) error
-type flightHandler func(*Conn) (bool, error)
+type handshakeMessageHandler func(*Conn) (*alert, error)
+type flightHandler func(*Conn) (bool, *alert, error)
 
 // Conn represents a DTLS connection
 type Conn struct {
@@ -69,9 +69,10 @@ type Conn struct {
 	rootCAs               *x509.CertPool
 	serverName            string
 
-	handshakeMessageHandler handshakeMessageHandler
-	flightHandler           flightHandler
-	handshakeCompleted      chan bool
+	handshakeMessageHandler        handshakeMessageHandler
+	flightHandler                  flightHandler
+	handshakeCompletedSignal       chan bool
+	handshakeCompletedSuccessfully atomic.Value
 
 	connErr atomic.Value
 	log     logging.LeveledLogger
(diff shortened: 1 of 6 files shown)


conn.go
```diff
@@ -69,9 +69,11 @@ type Conn struct {
 	rootCAs               *x509.CertPool
 	serverName            string
 
-	handshakeMessageHandler handshakeMessageHandler
-	flightHandler           flightHandler
-	handshakeCompleted      chan bool
+	handshakeMessageHandler             handshakeMessageHandler
+	flightHandler                       flightHandler
+	handshakeCompletedSignal            chan bool
+	handshakeCompletedSuccessfully      bool
```
Sean-Der: mind using an atomic instead, just one less member and we don't need to worry about invalid access!

conn.go
```diff
@@ -169,9 +171,16 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 	// Handle inbound
 	go c.inboundLoop()
 
-	<-c.handshakeCompleted
-	c.log.Trace("Handshake Completed")
```
Sean-Der: What do you think of `c.log.Trace(fmt.Sprintf("Handshake Completed %s", c.getConn()))` instead?

Probably a nit, but adding complexity for logging is something would like to avoid. Sorry to be annoying, but I just worry as we grow `pion/dtls` it will be harder for others to jump in

Sean-Der:
@RyanGordon I think this also fixes #5 ! Just triaging tickets

If you want to resolve that one with your commit message you will be working overtime :)

conn.go
RyanGordon: Yes, good idea :+1:

conn.go
RyanGordon: Hmm, not sure I follow. do you mean `getConnErr()` or add a new function `getConn()` for the log?

Sean-Der: sorry yes `getConnErr()` my mistake!

RyanGordon: Fixed up, does that look right?

Sean-Der: Oh yea that LGTM! Just nice to not worry about multiple branches

RyanGordon:
@Sean-Der Just double checking, does this look good to merge?

PR #98, opened 2019-08-24
Fix handshake race condition

RyanGordon:
#### Description
Currently cipherSuite can be created (non-nil) but not "initialized" (flight2 vs flight4). If this race happens the program tries to process the packet and fails in the decrypt stage, rather than _slightly more gracefully_ dropping it at the correct spot.

This race condition happens from time to time during the lossy E2E test, when client application data gets received by the server before the final client handshake packets are received.

(Pulled this fix out of https://github.com/pion/dtls/pull/93)


Sean-Der:
Nice! Thanks for picking stuff out into solo commits :)

PR #111, opened 2019-08-28
CI: Redo the pipeline

daenney:
* Use Travis build stages: stages run sequentially but jobs within
  a stage can be parametrized to run things in parallel. This would allow us to run the `test` stage with multiple versions of Go
* Do all the linting first, tests after, it's really annoying to have
  the tests pass and then the commit linter etc fail you
* Move tooling installation into before_script
* Only upload code coverage on test success, if tests fail codecov
  really can't tell us anything useful
* Add caching of Go module download and build results to speed up subsequent runs of tests for the same PR


Sean-Der:
Wow this is fantastic, yes we should get this everywhere!

PR #112, opened 2019-08-28
Fragmenting sent handshake messages

RyanGordon:
#### Description
Previously handshake messages being sent out would not be fragmented. Most handshake messages are pretty small but a couple of exceptions, such as Certificate, mean the packet size can be greater than the MTU, which then causes some packet loss and the handshake would never be able to complete. This changes the `internalSend` function to fragment handshakes to the set MTU.

We set `MTU` as a config option, which defaults to `1200`

### DIFF
--- a/client_handlers.go
+++ b/client_handlers.go
@@ -3,6 +3,7 @@ package dtls
 import (
 	"bytes"
 	"fmt"
+	"sync/atomic"
 )
 
 func initalizeCipherSuite(c *Conn, h *handshakeMessageServerKeyExchange) (*alert, error) {
@@ -171,7 +172,7 @@ func clientHandshakeHandler(c *Conn) (*alert, error) {
 			if alertPtr, err := handleSingleHandshake(expectedMessages[0].data); err != nil {
 				return alertPtr, err
 			}
-			c.state.localSequenceNumber++
+			c.handshakeMessageSequence++
 		case expectedMessages[1] != nil:
 			if alertPtr, err := handleSingleHandshake(expectedMessages[1].data); err != nil {
 				return alertPtr, err
@@ -220,7 +221,7 @@ func clientHandshakeHandler(c *Conn) (*alert, error) {
 			}
 		}
 
-		c.state.localSequenceNumber++
+		c.handshakeMessageSequence++
 		c.currFlight.set(flight5)
 	case flight5:
 		expectedMessages := c.handshakeCache.pull(
@@ -234,7 +235,8 @@ func clientHandshakeHandler(c *Conn) (*alert, error) {
 		}
 
 		c.setLocalEpoch(1)
-		c.state.localSequenceNumber = 1
+		c.handshakeMessageSequence = 1
+		atomic.StoreUint64(&c.state.localSequenceNumber, 1)
 		c.handshakeDoneSignal.Close()
 	default:
 		return &alert{alertLevelFatal, alertUnexpectedMessage}, fmt.Errorf("client asked to handle unknown flight (%d)", c.currFlight.get())
@@ -289,13 +291,11 @@ func clientFlightHandler(c *Conn) (bool, *alert, error) {
 
 		c.internalSend(&recordLayer{
 			recordLayerHeader: recordLayerHeader{
-				sequenceNumber:  c.state.localSequenceNumber,
 				protocolVersion: protocolVersion1_2,
 			},
 			content: &handshake{
-				// sequenceNumber and messageSequence line up, may need to be re-evaluated
 				handshakeHeader: handshakeHeader{
-					messageSequence: uint16(c.state.localSequenceNumber),
+					messageSequence: uint16(c.handshakeMessageSequence),
 				},
 				handshakeMessage: &handshakeMessageClientHello{
 					version:            protocolVersion1_2,
@@ -313,23 +313,21 @@ func clientFlightHandler(c *Conn) (bool, *alert, error) {
 			return true, nil, nil
 		}
 
-		sequenceNumber := c.state.localSequenceNumber
+		messageSequence := c.handshakeMessageSequence
 		if c.remoteRequestedCertificate {
 			c.internalSend(&recordLayer{
 				recordLayerHeader: recordLayerHeader{
-					sequenceNumber:  c.state.localSequenceNumber,
 					protocolVersion: protocolVersion1_2,
 				},
 				content: &handshake{
-					// sequenceNumber and messageSequence line up, may need to be re-evaluated
 					handshakeHeader: handshakeHeader{
-						messageSequence: uint16(c.state.localSequenceNumber),
+						messageSequence: uint16(messageSequence),
 					},
 					handshakeMessage: &handshakeMessageCertificate{
 						certificate: c.localCertificate,
 					}},
 			}, false)
-			sequenceNumber++
+			messageSequence++
 		}
 
 		clientKeyExchange := &handshakeMessageClientKeyExchange{}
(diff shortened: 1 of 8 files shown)


RyanGordon:
This is ready for review now. A couple of notes:

I don't _love_ the `Marshal()` interface, but it didn't really make sense to fit it anywhere else. I can't fragment before `packet.Marshal()`, because the handshake is just an in-memory struct still; we have no idea the full size of the handshake message itself. And I can't do it after because we need to manipulate the record and handshake headers to support fragmentation properly to spec. It might take a larger refactor to have a cleaner interface. Open to thoughts here.

I also realizing I needed to tackle the
`// sequenceNumber and messageSequence line up, may need to be re-evaluated`
tech debt because the possibility of fragmentation means these will not align anymore, so there is slightly more code tweaked to fix this case.

One last change was a small improvement to the way the "handshake completed signaling" works so that we don't end up in a situation where we could cause a panic calling an already-closed channel. I can split this out if desired / I may move this change to https://github.com/pion/dtls/pull/113

Sean-Der:
I am going to spend some time with this, but I am not a big fan of how the MTU is split out all over the code base.

I need to think over it, but it would be really nice if we didn't have to change `Marshal()` but just had a single function that took care of fragmenting for us.

 I don't think there is much danger to double parse/allocate since this isn't a hot path. I do think this could start to introduce substantial code complexity. Maybe this is the only way we can do it, but I will think it over and come back!

Sean-Der:
@RyanGordon re handshake complete stuff, merge away! Thank you so much for putting it in a smaller PR, it makes it so much easier to just worry about a single thing.

Sean-Der:
What do you think of keeping something in memory that maps `original sequence number` -> `fragmented`

Then only when it comes time to encrypt and send we can say `We are sending N`, but `N-1`  was actually these three so we need to account for that.

This adds more state/complexity to that function. But I just think in the long run it will pay off to have something that is very contained/testable.

RyanGordon:
I agree it would be nicer if we could keep `Marshal()` the same. I am going to think more on this but the basic complexity is the way the inheritance works with marshaling:

recordHeader.Marshal -> handshakeHeader.Marshal -> handshake.Marshal

If we reversed it, then it becomes much easier to keep the fragmenting and other stateful logic out of Marshal

i.e. very rough-cut:
```
internalSend(message) {
  var packets [][]byte
  if message.type == handshake {
    handshakes := fragment_handshake(message.handshake)
    for handshake in handshakes {
       record = ...
       packets = append(packets, record.Marshal())
   }
  } else {
   packets = [message.Marshal()]
  }

  // send packets
}

fragment_handshake(handshake) {
  content := handshake.content.Marshal()
  origHeader := handshake.header
  fragments = split(content, MTU_SIZE)
  
  handshakes = []
  for fragment in fragments {
     fragmentHeader = copy(origHeader)
     fragmentHeader.offset = ...
     fragmentHeader.len = ...
     fragmentHeader.content = fragment
     handshakes = append(handshakes, fragmentHeader)
  }
  
  return handshakes
}

```

Sean-Der:
This is fantastic @RyanGordon!

I am really impressed how fast you rewrote/fixed up this PR. Reading through the code I like it a lot more, it is just nice that it is only in one place.

I think that is the hardest part about OpenSSL. A lot of the TLS code duplicates logic/ a lot of things are interwoven so it is hard to follow.

This is a merge from me!

Sean-Der:
So checking out the PR I think that if we fire messages for a flight multiple times this will cause issues.

It looks like we increment the sequenceNumber every time we send a fragment. I think this a good design decision.

However if we re-run a flight multiple times the SequenceNumber will not account for that. I think the only change we need to make is to record `sequenceNumber at start of flight` and then we should be good! Then every time we re-run that flight we need to reset back.

Happy to take care of it also!

Sean-Der:
Another thing I realized, we currently might send different packets with the same sequenceNumber.

If this flight handler has run multiple times https://github.com/pion/dtls/blob/master/server_handlers.go#L481 we don't increment the global sequenceNumber counter until we have moved onto the next flight. 

If we hit an error during handshaking https://github.com/pion/dtls/blob/master/conn.go#L391 we will reuse a sequence number that has already been used. The global sequenceNumber will just be lower than what has actually gone out.

We need to handle sequenceNumbers and replaying flights better in general. The code is confusing error prone. Maybe we should 'replay flights' by pulling them from the local handshake cache? Then we can increment the global sequenceNumber cache w/e we want.

RyanGordon:
> So checking out the PR I think that if we fire messages for a flight multiple times this will cause issues.
> 
> It looks like we increment the sequenceNumber every time we send a fragment. That works great for the happy case.
> 
> However if we re-run a flight multiple times the SequenceNumber will not account for that. I think the only change we need to make is to record `sequenceNumber at start of flight` and then we should be good. Then every time we re-run that flight we need to reset back.
> 
> Happy to take care of it also!

Hmm, I think this is actually how it works because `c.handshakeMessageSequence` is used for the handshake header `messageSequence` field and `c.state.localSequenceNumber` is only used for the record header `sequenceNumber` field

`c.handshakeMessageSequence` is assigned based on the old logic that would correctly set the handshake header `messageSequence` to the right value for the state of the sequence, including replays.

And now `c.state.localSequenceNumber` is only used for the record and is incremented every time a record is sent so it is always unique for each sent record (even if the record contains retransmission content). At least according to RFC6347 4.1.2.6 seems to say that the record layer sequence number should always be unique or duplicate packets could be dropped by the receiver depending on the library implementation.

Sean-Der:
@RyanGordon 

(These are just assumptions, please correct me where I am wrong!)

* https://github.com/pion/dtls/blob/bc94af06cd09f3ec7fb246e5e0a264382da4e19e/conn.go#L379 always increments the SequenceNumber. We will never have a message with the same sequence number twice.

* This means that each call to internal send will use new sequenceNumbers https://github.com/pion/dtls/blob/bc94af06cd09f3ec7fb246e5e0a264382da4e19e/client_handlers.go#L292

`pion/dtls` itself is unable to handle gaps in sequence numbers. It is very explicit about the SequenceNumber that is used for certain messages https://github.com/pion/dtls/blob/bc94af06cd09f3ec7fb246e5e0a264382da4e19e/server_handlers.go#L188

RyanGordon:
Hmm okay, that is interesting. I need to do more research into this to make sure I am understanding the spec correctly and that this doesn't introduce breaking behavior. (Currently) My understanding is that the record "sequence number" (I believe) should always be incrementing even on re-transmit. The handshake "message sequence" is the one that should have the specific ordering and be re-used depending on the state of the handshake and used to know where in the handshake the sender is.

I will be double checking what the spec behavior _should be_, and what impact these changes  would have and what would need to be changed to avoid any impact

Sean-Der:
Fantastic, sorry for this review being a doozy! I really appreciate how much you are doing to land this.

I would see what OpenSSL does on retransmissions (if it increments the SequenceNumber), and honestly if E2E Lossy passes I think we are in great shape. 

I really need to make a nice way for us to do E2E Lossy against OpenSSL.

daenney:
> I really need to make a nice way for us to do E2E Lossy against OpenSSL.
I'm working on that. Once I've recovered from a 9hr timezone shift I'll submit

Sean-Der:
> > I really need to make a nice way for us to do E2E Lossy against OpenSSL.
> > I'm working on that. Once I've recovered from a 9hr timezone shift I'll submit

Well that is just a fantastic surprise. Thanks for grabbing that!

RyanGordon:
Ok, this is ready for a final review @Sean-Der, et al. Tests should pass but golangci-lint seems to be doing something weird (related to Go 1.13 release?)

Sean-Der:
Nice @RyanGordon !

Mind rebasing this off of you 1.12 pin I will make sure I review this tonight! If it passes everything that is for sure going to be a merge from me.

really nice work

e2e/e2e_test.go
```diff
@@ -232,3 +232,33 @@ func TestPionE2ESimplePSK(t *testing.T) {
 		assertE2ECommunication(cfg, cfg, serverPort, t)
 	}
 }
+
+func TestPionE2EMTUs(t *testing.T) {
```
Sean-Der: Can you introduce loss to this test as well? I want to see how it handles gaps in sequence numbers as well.

RyanGordon: Sure thing!

RyanGordon: Done!

client_handlers.go
```diff
@@ -171,7 +172,7 @@ func clientHandshakeHandler(c *Conn) (*alert, error) {
 			if alertPtr, err := handleSingleHandshake(expectedMessages[0].data); err != nil {
 				return alertPtr, err
 			}
-			c.state.localSequenceNumber++
+			c.handshakeMessageSequence++
```
Sean-Der: I think we can easily get this out of the global state, could help readability a bit I think.

we can put `handshakeMessageSequence` on the stack, and then look at the local cache for the last flight. Or just look at what features are enabled

Sean-Der: Don't worry about doing this in this PR, merge away!

PR #113, opened 2019-08-28
Fix connect timeout doesn't stop outgoing handshake routine

RyanGordon:
### Description
Recently a ConnectTimeout option was added which waits for the handshake to complete in the given timeframe or otherwise moves on. However, the outbound handshake goroutine was never cancelled when a timeout occurs. This PR refactors the previous signaling mechanism and adds a call to `c.handshakeDoneSignal.Close()` at timeout, so that the outgoing handshake goroutine receives the signal and stops.

#### Reference issue
Fixes regression from #94

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -75,7 +75,7 @@ type Conn struct {
 
 	handshakeMessageHandler        handshakeMessageHandler
 	flightHandler                  flightHandler
-	handshakeCompletedSignal       chan bool
+	handshakeDoneSignal            *Closer
 	handshakeCompletedSuccessfully atomic.Value
 
 	connErr atomic.Value
@@ -126,6 +126,8 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 		connectTimeout = math.MaxInt64 * time.Nanosecond
 	}
 
+	handshakeDoneSignal := NewCloser()
+
 	c := &Conn{
 		nextConn:                    nextConn,
 		currFlight:                  newFlight(isClient, logger),
@@ -149,10 +151,10 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 		localPSKCallback:     config.PSK,
 		localPSKIdentityHint: config.PSKIdentityHint,
 
-		decrypted:                make(chan []byte),
-		workerTicker:             time.NewTicker(workerInterval),
-		handshakeCompletedSignal: make(chan bool),
-		log:                      logger,
+		decrypted:           make(chan []byte),
+		workerTicker:        time.NewTicker(workerInterval),
+		handshakeDoneSignal: handshakeDoneSignal,
+		log:                 logger,
 	}
 
 	// Use host from conn address when serverName is not provided
@@ -188,10 +190,11 @@ func createConn(nextConn net.Conn, flightHandler flightHandler, handshakeMessage
 	go c.inboundLoop()
 
 	select {
-	case <-c.handshakeCompletedSignal:
+	case <-c.handshakeDoneSignal.Done():
 		err = c.getConnErr()
 	case <-time.After(c.connectTimeout):
 		err = errConnectTimeout
+		c.handshakeDoneSignal.Close()
 	}
 
 	if err == nil {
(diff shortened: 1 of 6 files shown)


RyanGordon:
Makes sense :+1: I will add a test tomorrow

RyanGordon:
Okay, I added a regression test (~400ms after optimizing the flight interval and timeout) that fails without the regression fix

I also switched the `handshakeCompletedSignal` from a channel to a `context.WithCancel()` which tends to be safer than calling close on a channel to signal a goroutine shutdown

Sean-Der:
Great work @RyanGordon 

We should move everything from `close` -> `Context`. Doing `close` is a very common pattern in the code bases, so I think it would be a great cleanup across the board. If you feel strongly about making this change I say we commit to doing it everywhere, just want to avoid the lava pattern where some code bases are more modern then others!

```
seaduboi@38f9d359441f:~/go/src/github.com/pion$ rg -l --fixed-strings 'close(' | grep -v dart | grep -v .js
stun/client.go
stun/client_test.go
mdns/conn.go
ice/gather.go
ice/agent.go
ice/agent_test.go
ice/candidate_base.go
ice/transport_test.go
ice/candidate_relay.go
ice/connectivity_vnet_test.go
ice/candidate.go
turn/server.go
turn/client_test.go
dtls/resume_test.go
dtls/conn.go
sctp/ack_timer_test.go
sctp/association_test.go
sctp/rtx_timer.go
sctp/rtx_timer_test.go
sctp/ack_timer.go
srtp/stream_srtp_test.go
srtp/session_srtp.go
srtp/session.go
sctp/association.go
srtp/session_srtcp.go
transport/test/bridge.go
transport/test/stress.go
quic/transport_test.go
transport/packetio/buffer_test.go
transport/packetio/buffer.go
webrtc/peerconnection_close_test.go
webrtc/sctptransport.go
webrtc/peerconnection_media_test.go
webrtc/quictransport_test.go
webrtc/datachannel.go
webrtc/icegathererstate.go
webrtc/datachannel_go_test.go
transport/vnet/router_test.go
transport/vnet/net_test.go
transport/vnet/stress_test.go
transport/vnet/conn_test.go
transport/vnet/conn.go
transport/vnet/router.go
webrtc/peerconnection.go
webrtc/rtpsender.go
webrtc/datachannel_ortc_test.go
webrtc/icetransport_test.go
webrtc/dtlstransportstate.go
webrtc/rtpreceiver.go
webrtc/peerconnection_go_test.go
dtls/e2e/e2e_lossy_test.go
stun/cmd/stun-traversal/main.go
turn/internal/client/transaction.go
turn/internal/client/trylock_test.go
turn/internal/client/periodic_timer.go
turn/internal/client/conn.go
turn/internal/allocation/allocation.go
dtls/internal/udp/conn.go
webrtc/internal/mux/mux.go
webrtc/internal/mux/endpoint.go
```

conn.go
```diff
@@ -75,7 +76,8 @@ type Conn struct {
 
 	handshakeMessageHandler        handshakeMessageHandler
 	flightHandler                  flightHandler
-	handshakeCompletedSignal       chan bool
+	handshakeCtx                   context.Context
```
Sean-Der: Mind choosing a name that is a little more descriptive and putting these members together? On my first read through I was confused if `handshakeCtx` and `handshakeDoneSignal` were related.

Maybe `handshakeCtx` and `handshakeCtxCancel` (or anything like that) just want to make it clear to the reader!

RyanGordon:
Personally I find it a bit dangerous to use a chan for shutdown because its not idempotent and the failure case is a panic. I.e. `close()` is the only way to guarantee precedence if multiple channels are being listened to in a `select` and if `close()` ends up being called twice by the application for whatever reason that leads to the panic.

I do agree chan is a common pattern to use for signaling closing, I know there are a few opinions on the best approach out there but generally, even if it looks a little worse in implementation (handling `ctx` and `func()` from `context.WithCancel(...)` in practice its much safer than closing channels.

Happy to have more discussion on this, but if we have consensus on the approach, I can open up PRs for the additional places you've found. We could even create a wrapper struct to abstract it, like this:

```
type Closer struct {
   ctx context.Context
   shutdownFunc func()
}

NewCloser() *Closer {
   ctx, shutdownFunc := context.WithCancel(context.Background())
   return &Closer{
      ctx: ctx,
      shutdownFunc: shutdownFunc,
   }
}

func (c *Closer) Closed() <-chan struct{}  {
   return c.ctx.Done()
}

func (c *Closer) Close() {
   c.shutdownFunc()
}
```

Then we can use the abstraction like this in the various libraries:

```
...
a.handshakeCloser := NewCloser()
defer a.handshakeCloser.Close()
go a.someFunc()
...
``` 

```
func (a ...) someFunc() {
  for { 
     select {
     case <-a.handshakeCloser.Closed():
        return
     case ...
     }
  }
}
```

RyanGordon:
Okay, I updated this, mind reviewing again when you get a chance?

Sean-Der:
@RyanGordon oh yea that is great, merge away!

I think the old `close` method is a 'clever hack' I don't think if I saw that for the first time I would understand the intent. I will be copying this into `pion/webrtc` when I get the chance for sure :)

PR #114, opened 2019-08-28
Don't reuse the Config struct

daenney:
~~In the benchmarks and TestSimpleE2E we used to create a Config struct once and then pass
it to both the client and the server. This can cause a concurrent access
to members of the Config struct triggering the race detector.~~ Literature is being annoying and unclear about concurrent read access to struct members but I can't trigger a race with a simple test on my side. It makes sense that this would be fine too.

Aside from that, we're now passing options into client that it has no
business needing (i.e the private key and cert) and the server doesn't
do anything with InsecureSkipVerify.

This change ensures that the server config is created inside the
goroutine we spin off to run the server, so every server gets its own,
and we only set the options the server needs. We similarly generate a
separate config with only the options the client should have.

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -15,16 +15,14 @@ func TestSimpleReadWrite(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	config := &Config{
-		Certificate:        certificate,
-		PrivateKey:         privateKey,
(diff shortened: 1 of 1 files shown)


daenney:
Eh, how did this end up affecting code coverage?

PR #117, opened 2019-09-03
Version lock CI to Go 1.12

RyanGordon:
#### Description
All CI builds are currently failing. Locking CI to 1.12 until golangci-lint is fixed or we replace it with a different recommendation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ branches:
   - master
 
 go:
-  - "1.x" # use the latest Go release
+  - "1.12" # use the latest Go release
 
 env:
   - GO111MODULE=on


Sean-Der:
Ugh sorry about that headache :/

I hope we can stick with `golangci-lint` it seems to be the standard as far as linting currently! If something else is out there always ready to try it out

RyanGordon:
I'm not too familiar with the go lint scene, but someone in #pion suggested "staticcheck + golint". 🤷‍♂ Anyway, just something to consider for the future

daenney:
It would be good in general to do something like "we support current Go -1", so in our case that would've been 1.11 and 1.12 and have a build in the matrix that can fail that uses Go's tip/master so we can catch issues around that earlier.

PR #118, opened 2019-09-11
Compact handshake records into less datagrams

RyanGordon:
#### Description
Currently each record during the handshake process is sent as a separate datagram.

The dTLS spec allows for combining records into a single datagram so long as the datagram fits within the MTU.

Here we implement this improvement, compacting records during the handshake process into less packets. The number of packets sent for a complete handshake before this change was 12. With this change the number is reduced to 7.

While improvements are likely hard to measure, this should theoretically reduce tail handshake latency in lossy networks.

Note: This PR is more easily viewed with `?w=1`

### DIFF
--- a/client_handlers.go
+++ b/client_handlers.go
@@ -289,23 +289,26 @@ func clientFlightHandler(c *Conn) (bool, *alert, error) {
 			})
 		}
 
-		c.internalSend(&recordLayer{
-			recordLayerHeader: recordLayerHeader{
-				protocolVersion: protocolVersion1_2,
-			},
-			content: &handshake{
-				handshakeHeader: handshakeHeader{
-					messageSequence: uint16(c.handshakeMessageSequence),
+		c.internalSend(&packet{
+			record: &recordLayer{
+				recordLayerHeader: recordLayerHeader{
+					protocolVersion: protocolVersion1_2,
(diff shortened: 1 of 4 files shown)


Sean-Der:
@RyanGordon 

This is great! I can imagine this is super helpful in lossy networks. If you can group together everything in a flight makes things a lot easier.

To cut down on complexity can we just get rid of `internalSend` and always call `bufferPacket` and then `flushBufferPacket` at the very end of `clientFlightHandler` and `serverFlightHandler`. I think that is just easier for the reader to understand. 

I see we call `flushBufferPacket` multiple times explicitly, any advantage to that?

Sean-Der:
@RyanGordon 

I can't think of any additional tests I would want to add either. Since we test with different MTU sizes we already have great coverage! This is a merge from me, I would just drop `internalSend` since it just makes more for people to understand.

great work.

Sean-Der:
@RyanGordon 

No rush on getting this in, but if there is anything I can do here to help :) once we land this I would love to do `/v2` so we can support others.

thanks

Sean-Der:
I am moving this to a branch I have access to! Will re-open and merge with @daenney, thanks for all the work on this @RyanGordon :)

PR #119, opened 2019-09-13
Update golangci-lint to 1.18.0

Sean-Der:
Adds 1.13 compatibility, don't pin to 1.12 anymore


daenney:
I'm not super fond of immediately forcing Go 1.13 on everyone.

Sean-Der:
@daenney 

Pulled that back! No idea what I did to add that, but only wanted to bump golangci and start testing with v1.13

PR #129, opened 2019-10-14
Compact handshake records into less packets

Sean-Der:
Currently each record during the handshake process is sent as a
separate packet. The dTLS spec allows for combining records into a
signal packet so long as the packet fits within the MTU. Here we
implement this improvement, compacting records during the handshake
process into less packets. The number of packets sent for a complete
handshake before this change was 12. With this change the number is
reduced to 7. While improvements are likely hard to measure, this
should reduce tail handshake latency in lossy networks.

### DIFF
--- a/client_handlers.go
+++ b/client_handlers.go
@@ -289,23 +289,26 @@ func clientFlightHandler(c *Conn) (bool, *alert, error) {
 			})
 		}
 
-		c.internalSend(&recordLayer{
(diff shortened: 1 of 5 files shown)


daenney:
@Sean-Der Do you still want to merge this?

Sean-Der:
Yes! I am going to do it this weekend, but wanted to dig through it fully before it landed!

After that will get the other one and we will be down to 0 PRs :)

Issue #133, opened 2019-10-29
Only ECDSA keys supported


hjames9:
Hello, it appears in the code that only ECDSA keys are supported.
https://github.com/pion/dtls/blob/master/conn.go#L100

However in the documentation (mainly README.md), it mentions the following ciper suites as being supported:
TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA

Is the documentation incorrect or is there another way to bypass that certificate type check?

Sean-Der:
@hjames9 I *believe* RSA works if the remote offers it, but you are right locally we only support ECDSA.

I am not against supporting it though, just need to see the complexity vs value trade off. I don't want to add bugs for other things in the process of supporting it.

rumpelsepp:
Is adding support for (local) ed25519 keys possible in the current codebase? I mean, are there only  a few lines that need to be changed, or is some larger refactoring needed?

Sean-Der:
@rumpelsepp I *think* just a few lines need changing. 

We have a few cases where we switch off the type of the key. Someone just needs to support the cases where that fails and we should be good!

rumpelsepp:
ed25519 is implemented as of today.

hjames9:
@Sean-Der I was thinking of maybe tackling RSA certificate support.   Is there anything in particular you were weary about with this implementation?

Sean-Der:
@hjames9 if you copy/change what @rumpelsepp did it should be perfect!

So excited you are contributing, are you interested in being involved long term? I can add you to the Pion organization, then you can merge your own PRs after one approval (and you can approve others!)

Also if you want quicker responses I am on [slack](pion.ly/slack) 24/7 :)

hjames9:
@Sean-Der Great!  I'll submit a PR adding RSA support modeling what @rumpelsepp has done, which can be used to close out this ticket.

Would love to join Pion, and I will also hop on Slack.  Thanks!

rumpelsepp:
I think rsa is not that easy as ed25519, since there are multiple signature formats available and these need to be marshalled with asn1. I need to read the rfc though…

daenney:
I'm going to close this issue. If people want to support additional key types then PRs are always welcome. If people want to submit a PR to clarify that we only support the RSA in certain situations, that's also fine and appreciated.

GregorioDiStefano:
Reopening this instead of creating a new issue.

[...]

PR #136, opened 2019-10-31
Add support for ed25519 keys

rumpelsepp:
see discussion here:
https://github.com/pion/dtls/issues/133

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Lander Noterman](https://github.com/LanderN)
 * [Aleksandr Razumov](https://github.com/ernado) - *Fuzzing*
 * [Ryan Gordon](https://github.com/ryangordon)
+* [Stefan Tatschner](https://rumpelsepp.org/contact.html)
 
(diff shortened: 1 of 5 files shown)


rumpelsepp:
tests added

rumpelsepp:
The tests fail due to false positives… How to proceed?

rumpelsepp:
I also added a helper method to remove the code duplication. `SelfSign()` can be used to create self signed certs from both ecdsa and ed25519 keys.

rumpelsepp:
The linter *is* really annoying. Here you are. I disabled the security check on the random number generation, since it is obviously a false positive. There is only `crypto/rand` imported…

Sean-Der:
Merged, no this was perfect thank you so much @rumpelsepp!

Would you be interested in joining the Pion org and helping out with `pion/dtls`? Always looking for people who are interested :)

I love the refactor PRs, and changes like that would be great (anything goes IMO as long as it doesn't change the public API or break tests!)

rumpelsepp:
> Would you be interested in joining the Pion org and helping out with pion/dtls? Always looking for people who are interested :)

Sure! I can try my best.

PR #140, opened 2019-11-07
Validate DTLS config object before listening on port

### LINKED ISSUE #139: DTLS listener should fail immediately on invalid configuration
This is related to an issue I had using DTLS mode in the go-coap library (https://github.com/go-ocf/go-coap/issues/69).  If I attempt to instantiate a DTLS listener, with an invalid configuration (in my case attempting to use an RSA key), the listener will start successfully, but all incoming messages will fail.

### What did you expect?
If provided an invalid DTLS config object, the listener should return an error object with details on the configuration failure.

### What happened?
The DTLS listener successfully started, but all messages on the session fail

I've added a PR with the above changes.  Please review and let me know if it makes sense.
https://github.com/pion/dtls/pull/140
Thanks!

hjames9:
Before successfully listening on a UDP port to receive DTLS messages,
we should validate that the configuration is in working condition to avoid
deferring any failures on incoming messages.

#### Reference issue
Fixes #139

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -1,7 +1,6 @@
 package dtls
 
 import (
-	"errors"
 	"net"
 	"time"
 
@@ -10,9 +9,10 @@ import (
 
 // Listen creates a DTLS listener
 func Listen(network string, laddr *net.UDPAddr, config *Config) (*Listener, error) {
-	if config == nil {
-		return nil, errors.New("no config provided")
+	if err := validateConfig(config); err != nil {
+		return nil, err
 	}
+
 	parent, err := udp.Listen(network, laddr)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


daenney:
This looks mostly fine but:

- Commit linter is unhappy with you :wink: 
- It's missing tests. The validation method is new, so I'd like to see a few tests that exercise it

hjames9:
@daenney Thank you for the review.  I've fixed the lint failures and added test cases for the new validate function.

hjames9:
@daenney I cleaned up the git author to match the commit, but the changes are the same.  Should be ready to merge unless you have any other items.  Thanks!

hjames9:
Thank you.  How do I go about getting this merged?

listener.go
```diff
@@ -10,9 +9,11 @@ import (
 
 // Listen creates a DTLS listener
 func Listen(network string, laddr *net.UDPAddr, config *Config) (*Listener, error) {
-	if config == nil {
-		return nil, errors.New("no config provided")
+	err := validateConfig(config)
```
rumpelsepp: Personally I like one liners for this better:

``` go
if err := validateConfig(config); err != nil {
    //…
}
```

daenney: Mmm yes, since err isn't used any further that's kinda nicer. @hjames9 could you address this one, then I'll merge it.

hjames9:
@rumpelsepp @daenney Updated based on feedback

Sean-Der:
Merged with d0f3d4c64ee6a8768c860932f7e6dcee3df0cc31 🎉

PR #142, opened 2019-11-11
V2 Release! First breaking change certificate chains.

Sean-Der:
Thank you so much @jkralik for this patch, sorry it took me so long :( but we are going to land this now and get in some other stuff!

-----

Breaking changes:
* certificate, privateKey was replaced
  by certificate (tls.Certificate)
* verifyPeerCertificates uses array of bytes for chain
  certificate instead of certificate(*x509.Certificate)

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,9 +1,9 @@
 package dtls
 
 import (
-	"crypto"
 	"crypto/ecdsa"
 	"crypto/ed25519"
+	"crypto/tls"
 	"crypto/x509"
 	"time"
 
@@ -16,11 +16,7 @@ type Config struct {
 	// Certificates contains certificate chain to present to the other side of the connection.
 	// Server MUST set this if PSK is non-nil
 	// client SHOULD sets this so CertificateRequests can be handled if PSK is non-nil
-	Certificate *x509.Certificate
-
-	// PrivateKey contains matching private key for the certificate
-	// only ECDSA is supported
-	PrivateKey crypto.PrivateKey
+	Certificate tls.Certificate
 
 	// CipherSuites is a list of supported cipher suites.
 	// If CipherSuites is nil, a default list is used
@@ -67,7 +63,7 @@ type Config struct {
 	// setting InsecureSkipVerify, or (for a server) when ClientAuth is
 	// RequestClientCert or RequireAnyClientCert, then this callback will
 	// be considered but the verified flag will always be false.
-	VerifyPeerCertificate func(cer *x509.Certificate, verified bool) error
+	VerifyPeerCertificate func(rawCert [][]byte, verified bool) error
 
 	// RootCAs defines the set of root certificate authorities
 	// that one peer uses when verifying the other peer's certificates.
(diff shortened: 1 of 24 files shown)


config.go
```diff
@@ -130,14 +126,14 @@ func validateConfig(config *Config) error {
 	switch {
 	case config == nil:
 		return errNoConfigProvided
-	case config.Certificate != nil && config.PSK != nil:
+	case config.Certificate.PrivateKey != nil && config.PSK != nil:
```
hjames9: I notice this in a few places, shouldn't we check both Certificate.Certificate and Certificate.PrivateKey?  Maybe write one validation function that's reused every place to make sure both those fields are set properly?

crypto.go
```diff
@@ -139,27 +155,58 @@ func verifyCertificateVerify(handshakeBodies []byte, hashAlgorithm HashAlgorithm
 	return errKeySignatureVerifyUnimplemented
 }
 
-func verifyClientCert(cert *x509.Certificate, roots *x509.CertPool) error {
+func loadCerts(rawCertificates [][]byte) ([]*x509.Certificate, error) {
+	certs := make([]*x509.Certificate, 0, 16)
```
hjames9: Seems like the size of this slice should be based off the length of rawCertificates

handshake_message_certificate.go
```diff
@@ -1,28 +1,34 @@
...
+	for _, r := range h.certificate {
+		certTotalLen += len(r)
 	}
-
-	out := make([]byte, 6)
-	putBigEndianUint24(out, uint32(len(raw))+3)
-	putBigEndianUint24(out[3:], uint32(len(raw)))
-
-	return append(out, raw...), nil
+	outLen := len(h.certificate)*3 + certTotalLen
```
hjames9: A few magic numbers in this function (i.e. 3 and 6 mainly).  Should this be replaced with const integers that illustrate what the actual values mean?

Sean-Der: Fixed!

crypto.go
Sean-Der: Fixed!

config.go
Sean-Der: I double checked and this should be good! If we are a Client `Certificate.Certificate` may be `nil`.

We do `Config` validation in three different places `Server` `Client` and `validateConfig` I think it would be a great improvement to merge these all into `validateConfig`!

daenney:
I'm a bit puzzled by something. In a few places I now see this show up:

```go
func xxx(... rawCertificates [][]byte ...) ... {
if len(rawCertificates) == 0 {
		return errLengthMismatch
	}
certificate, err := x509.ParseCertificate(rawCertificates[0])
```

Since it seems we only ever care/check for/validate the first one, what is the idea behind accepting a `[][]byte`? And if `len(rawCertificates) > 1`, what happens to those?

Sean-Der:
@jkralik hey (no rush though)! any thoughts on @daenney question, I was just fixing merge conflicts haven't really dove in yet.

@hjames9 mind giving it another look over? I tried to use constants everywhere/simplify stuff as much as I possible.

crypto.go
```diff
@@ -139,27 +155,58 @@ func verifyCertificateVerify(handshakeBodies []byte, hashAlgorithm HashAlgorithm
 	return errKeySignatureVerifyUnimplemented
 }
 
-func verifyClientCert(cert *x509.Certificate, roots *x509.CertPool) error {
+func loadCerts(rawCertificates [][]byte) ([]*x509.Certificate, error) {
+	certs := make([]*x509.Certificate, 0, len(rawCertificates))
```
hjames9: Minor issue, but you might want to move this below the size check to avoid the unnecessary allocation

hjames9: Changes look good besides this one minor issue.  I'll approve afterwards

Sean-Der: Fixed!

jkralik:
@Sean-Der Sorry for late response but I was very busy. I created PR https://github.com/pion/dtls/pull/152 for this branch to fix issues. And I updated Config  -> Now it is similar as is crypto/tls. But for now it supports only one certificate.

PR #154, opened 2019-11-25
Replacing crypto/ed25519 with compatible wrapper

epes:
https://golang.org/doc/go1.13#crypto/ed25519 added
the standalone crypto/ed25519 package but it does not
exist below Go 1.13. This change uses the previous /x/
package that in 1.13+ acts as a wrapper for the new
crypto/ed25519 package.

#### Reference issue
Fixes #...

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Stefan Tatschner](https://rumpelsepp.org/contact.html)
 * [Hayden James](https://github.com/hjames9)
 * [Jozef Kralik](https://github.com/jkralik)
+* [Robert Eperjesi](https://github.com/epes)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/config.go
+++ b/config.go
@@ -2,10 +2,11 @@ package dtls
 
 import (
 	"crypto/ecdsa"
-	"crypto/ed25519"
 	"crypto/tls"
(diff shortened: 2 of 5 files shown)


rumpelsepp:
What's the situation with go 1.14? They use conditional compiling stuff for only 1.13. This might prevent us from upgrading.

daenney:
Could you elaborate a bit @rumpelsepp? I don't understand what you're asking.

daenney:
Could you elaborate a bit @rumpelsepp? I don't understand what you're asking.

rumpelsepp:
This package uses conditional compiling for go1.13, see here: https://github.com/golang/crypto/tree/master/ed25519

Once there is 1.14 out, it won't use the stdlib anymore if the wrapper itself is not updated. I would recommend to rely only on the stdlib here. Also, this change is pretty much useless, since the crypto/x509 methods do not support ed25519 keys below go 1.13:

* https://pkg.go.dev/crypto/x509@go1.12.13?tab=doc#MarshalPKCS8PrivateKey
* https://pkg.go.dev/crypto/x509?tab=doc#MarshalPKCS8PrivateKey

What's the minimum go version for this package anyway? I recommend bumping it to go1.13.

daenney:
CI runs with "1.x", meaning whatever the latest stable release is. So currently though the package may work with 1.12 and earlier, we don't test/guarantee anything for it. The `go.mod` also specifies 1.13, so you should be good to raise a PR with the proposed change.

PR #156, opened 2019-12-03
Add CCM cipher suits for ECDHE_ECDSA, PSK

jkralik:
#### Feature
Add ciphers:
* TLS_ECDHE_ECDSA_WITH_AES_128_CCM
* TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8
* TLS_PSK_WITH_AES_128_CCM

#### Description
By OCF  Specification(https://openconnectivity.org/specs/OCF_Security_Specification_v2.1.0.pdf) only few ciphers are supported - mainly for certificates(TLS_ECDHE_ECDSA_WITH_AES_128_CCM, TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8). This commit implements the ciphers.


Sean-Der:
nice, this LGTM @jkralik

Not your problem, but we should fine a way to cut down on all the duplication when defining a crypto suite. I could see a bug being found in one, and not fix in the other ones.

jkralik:
@Sean-Der thx for comment. I merged them to the one struct.

crypto_ccm.go
```diff
@@ -9,23 +9,27 @@ import (
 	"github.com/pion/dtls/v2/internal/crypto/ccm"
 )
 
+type cryptoCCMTagLen int
```
daenney: We can change this later, but why is this its own type? Looking at the code, everywhere it's used we do `int(cryptoCCMTagLen)`. This seems a bit pointless, if all we ever do with it is convert it back to `int` before we can use it.

jkralik: I want to provides meaningful values. Int type nothing says which values are supported.

jkralik:
Hi. Pls could I merge it or I need to wait ? Thx

daenney:
Once it has the minimal approvals required you can always go ahead and merge it.

Sean-Der:
Yes please merge away @jkralik! w/e that button is green feel free to merge :)

Maybe we should more stringent, but would rather see people move quickly/unblocked and then we can go back and fix things later.

PR #157, opened 2019-12-06
prf: Use X25519() instead of ScalarMult

daenney:
In a recent update to x/crypto ScalarMult() has been deprecated in favour
of X25519(). Because of this staticcheck will start failing on prf.go.

This updates our x/crypto dependency as well as changing the necessary
code to use X25519() instead. It also happens to simplify the code a
bit, which is a nice benefit.

Closes #155

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/pion/dtls/v2
 require (
 	github.com/pion/logging v0.2.2
 	github.com/pion/transport v0.8.10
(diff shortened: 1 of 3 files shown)


daenney:
Related:
* https://github.com/golang/go/issues/32670
* https://github.com/golang/crypto/commit/f4817d981bb690635456c5c1c6aa0585e5d45891

Sean-Der:
That is awesome! Nice work @daenney 

When I saw renovate failing I was really worried it was going to be a big thing :/

PR #158, opened 2019-12-09
Revert "Replacing crypto/ed25519 with compatible wrapper"

rumpelsepp:
This reverts commit 0958495f8bfba4c6c2ce23cd851cda89deb972c5.

As discussed here: https://github.com/pion/dtls/pull/154#issuecomment-561936942

> This package uses conditional compiling for go1.13, see here:
> https://github.com/golang/crypto/tree/master/ed25519
>
> Once there is 1.14 out, it won't use the stdlib anymore if the wrapper
> itself is not updated. I would recommend to rely only on the stdlib
> here. Also, this change is pretty much useless, since the crypto/x509
> methods do not support ed25519 keys below go 1.13:
>
> https://pkg.go.dev/crypto/x509@go1.12.13?tab=doc#MarshalPKCS8PrivateKey
> https://pkg.go.dev/crypto/x509?tab=doc#MarshalPKCS8PrivateKey

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ We would love contributes that fall under the 'Planned Features' and fixing any 
 
 ### Using
 
+Go 1.13 is required as the minimum Go version.
+
 #### Pion DTLS
 For a DTLS 1.2 Server that listens on 127.0.0.1:4444
 ```sh
--- a/config.go
+++ b/config.go
@@ -2,11 +2,10 @@ package dtls
 
 import (
 	"crypto/ecdsa"
+	"crypto/ed25519"
 	"crypto/tls"
 	"crypto/x509"
 	"time"
-
-	"golang.org/x/crypto/ed25519"
 
 	"github.com/pion/logging"
 )
--- a/crypto.go
+++ b/crypto.go
@@ -3,6 +3,7 @@ package dtls
 import (
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/ed25519"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/sha256"
@@ -11,8 +12,6 @@ import (
 	"encoding/binary"
 	"math/big"
 	"time"
-
-	"golang.org/x/crypto/ed25519"
 )
 
 type ecdsaSignature struct {
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -1,6 +1,7 @@
 package e2e
 
 import (
+	"crypto/ed25519"
 	"crypto/rand"
 	"crypto/tls"
 	"errors"
@@ -10,8 +11,6 @@ import (
 	"sync/atomic"
 	"testing"
 	"time"
-
-	"golang.org/x/crypto/ed25519"
 
 	"github.com/pion/dtls/v2"
 	"github.com/pion/transport/test"
--- a/util.go
+++ b/util.go
@@ -3,6 +3,7 @@ package dtls
 import (
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/ed25519"
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/tls"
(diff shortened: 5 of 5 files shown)


README.md
```diff
@@ -131,7 +133,6 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Stefan Tatschner](https://rumpelsepp.org/contact.html)
 * [Hayden James](https://github.com/hjames9)
 * [Jozef Kralik](https://github.com/jkralik)
-* [Robert Eperjesi](https://github.com/epes)
```
rumpelsepp: I used `git revert`, sorry.

rumpelsepp: If this is about to be merged, I will remove this line from the commit.

daenney: Please remove it now, no need to wait.

rumpelsepp: done.

rumpelsepp:
> Aside from that, 50-52 characters is the expected length of the subject:
>
> The defaults are 50 characters for the summary and 72 characters for the description

The 50 char subject length is annoying, since git generates sometimes longer subjects, e.g. `git revert` or `git merge`. Anyway, the commit is now dropped.

rumpelsepp:
Removed the commit description. Hope it builds now, I hate this check.

Sean-Der:
@rumpelsepp really sorry about the commit message stuff, it is garbage (and my fault :(  )

I want to move to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) but just haven't had the time. I want to generate changelogs and give users nice commit logs to audit. People have asked for it multiple times.

We do need to have the linting because people would push commit with messages like `fix issue` it was very frustrating at the time!

-----

Is having this code in blocking anything/anyone? Currently we have a few users on 1.12, and reverting this will break them again.

rumpelsepp:
> @rumpelsepp really sorry about the commit message stuff, it is garbage (and my fault :( )

Never mind. It was a bit annoying, but I finally made it. :-)

> Is having this code in blocking anything/anyone? Currently we have a few users on 1.12, and reverting this will break them again.

We could also delay this a bit and e.g. make a policy that the two most recent go versions should be supported. Right now this would mean go 1.13 and 1.12. As go has a pretty well defined release cycle it is also predictable for folks using this library.

daenney:
The problem with users currently on Go 1.12 is that we don't actually support them. `go.mod` says Go 1.13, which means we can already be using things introduced in Go 1.13 that wouldn't work for Go 1.12 users, and that'll blow up at compile time.

The reason for #154 as far as I can see is #136, which introduced support for ED25519 but broke Go 1.12 compatibility. CI never caught it, since we don't test on anything but latest stable. This now reverts #154, essentially bringing us back to the state of #136 as far as possible Go 1.12 support goes, basically one month ago.

I think this makes it clear we need to adopt an actual policy here on what to do, b/c we can't keep going back and forth and breaking the library on people like this.

The suggestion of supporting both Go 1.12 and Go 1.13, and merging this once Go 1.14 comes out makes sense to me. If people are happy with that we can leave this one open. I'll submit a separate PR to ensure we run CI for the 2 Go versions we support so we can catch this kind of issue before it gets merged in the future.

rumpelsepp:
From: https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance

> Minor releases to address non-security problems for Go 1.x stop once Go 1.x+1 is released.
>
> Minor releases to address security problems for Go 1.x stop once Go 1.x+2 is released. For more about security updates, see the security policy.

That being said, even the three most recent releases are feasible. For now this is 1.11, 1.12, 1.13. Personally, I think Go 1.x+1 is enough, the decision is yours.

Sean-Der:
nothing to add here :) thanks @rumpelsepp and @daenney for taking time out to maintain/work on this stuff I really appreciate it! It is a good way to end the night reading this thread.

I am going to explore moving to GitHub actions (and fixing some of our tech debt around CI) this weekend! I will make a list on the wiki of everything we want. The big ones are a better commit message linter and testing on windows.

PR #161, opened 2019-12-15
Fix Close error affected by closed nextConn

at-wat:
Stored error in Conn was overwritten by inboundLoop() after Close()
set ErrConnClosed. It caused EOF error of Close() against healthy
connection especially on CI environment.

Investigation of this problem was on https://github.com/pion/webrtc/pull/943.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


conn_test.go
```diff
@@ -73,6 +74,46 @@ func stressDuplex(t *testing.T) {
 	}
 }
 
+func TestStressClose(t *testing.T) {
```
at-wat: This test can reproduce original problem not only on Travis but local.
It doesn't detect in 100%, so `go test . -count=N` would help.

Sean-Der:
@at-wat I am so happy you figured out that issue, it has been driving me crazy for months! I have been playing with your fix and had some questions/thoughts https://github.com/pion/dtls/tree/fix-close-retval-2

#### Add a test that can reproduce this every time
I can do this! Basically make it so you can `Close()` both sides and block the `Read()` and only have it return when we are ready. Then you can call `Close` and `Write` to get the error.

#### Store the first `stopWithErr` call
Right now each call to `stopWithError` will overwrite the previous values, I think we should take the first value only. So if we close with `ErrConnClosed` all the other calls too `stopWithErr` should be discarded.

at-wat:
@Sean-Der 

> Add a test that can reproduce this every time

Yeah, that's way better.
My version of the test consumes long to detect the problem in high probability.

> Store the first `stopWithErr` call

Sounds good!
Returning the first error would be more helpful for the library user as the first error can be assumed as a root of the problem.

Sean-Der:
@at-wat great! I can do all this work and send you a PR (if you don't mind waiting one more day)

Sean-Der:
Moved to #162

PR #162, opened 2019-12-17
Only set connErr once

Sean-Der:
Don't store subsequent calls to stopWithError.
During a graceful shutdown we would overwrite the
original value with the value from the nextConn

Co-authored-by: Atsushi Watanabe <atsushi.w@ieee.org>

[0] https://github.com/pion/webrtc/pull/943#issuecomment-565801043

### DIFF
--- a/atomic_error.go
+++ b/atomic_error.go
@@ -1,0 +1,22 @@
+package dtls
+
+import "sync"
+
+type atomicError struct {
+	mu  sync.Mutex
+	val error
+}
+
+func (a *atomicError) store(err error) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	a.val = err
+}
+
+func (a *atomicError) load() error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+
+	return a.val
+}
--- a/client_handlers.go
+++ b/client_handlers.go
@@ -291,7 +291,7 @@ func clientFlightHandler(c *Conn) (bool, *alert, error) {
 			})
 		}
 
-		c.bufferPacket(&packet{
+		if err := c.bufferPacket(&packet{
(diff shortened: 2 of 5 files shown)


Sean-Der:
@at-wat This fixes the root cause, and catches the error everytime!

If you revert the `conn.go` and `resume.go` change the error from `nextConn.Read()` incorrectly bubbles up everytime.

daenney:
```
--- FAIL: TestRoutineLeakOnClose (0.41s)
    conn_test.go:97: io: read/write on closed pipe
```

at-wat:
also this one
```
=== RUN   TestStressDuplex
--- FAIL: TestStressDuplex (0.86s)
    conn_test.go:61: io: read/write on closed pipe
```

at-wat:
https://github.com/pion/dtls/blob/486ad1d14fc4932346dd3620f075b2cc32d64b4b/conn_test.go#L54-L63
I guess `ca.Close()` causes `cb.Read()` `ErrClosedPipe`.

at-wat:
@Sean-Der So, I think this error is caused by the mixture of read errror and close error.
It would be better to separately store `nextConn.Read` error, `nextConn.Write` error, and `nextConn.Close` error, then return them from `conn.Read`, `conn.Write`, and `conn.Close` respectively.

Sean-Der:
@at-wat Great idea! I went and looked at `crypto/tls` and that is what they do as well.

[reading an error only for writing](https://github.com/golang/go/blob/master/src/crypto/tls/conn.go#L1079-L1081)

Sean-Der:
@at-wat I rebased this off your changes, will keep working on this tonight when I get home from work! Thanks for all your work :)

PR #166, opened 2020-01-08
Update assets to v0.0.0

pionbot:
Update lint scripts and CI configs.

PR #167, opened 2020-01-08
Disable disallowed-func-lint line by line

at-wat:
instead of excluding whole file.

This change allow to unify lint script among the repositories.

ref: https://github.com/pion/dtls/pull/166#discussion_r364023754


at-wat:
rebasing

PR #168, opened 2020-01-10
Define internal/crypto/ccm errors as var

at-wat:
To make it easy to unit test failure cases.
Tests to check panics and errors on invalid inputs are added.

(This increases patch cover ratio of #167.)


Sean-Der:
That's fantastic! Thank you so much @at-wat :)

I am just announcing `pion/turn` v2 right now, and then will be finishing up a `/v2` here!

PR #170, opened 2020-01-10
Update assets to v0.0.2

pionbot:


.github/lint-filename.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
at-wat: not yet called from `.travis.yml`

PR #173, opened 2020-01-12
Support WASM build/test

at-wat:
Some tests using net.Conn connection are disabled for now
as it is not fully supported by WASM MVP stage.

preparation for https://github.com/pion/.goassets/pull/4

### DIFF
--- a/e2e/e2e_lossy_test.go
+++ b/e2e/e2e_lossy_test.go
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	"crypto/tls"
+	"fmt"
 	"math/rand"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


e2e/e2e_lossy_test.go
```diff
@@ -100,78 +101,88 @@ func TestPionE2ELossy(t *testing.T) {
...
-			cfg := &dtls.Config{
-				FlightInterval:     flightInterval,
-				CipherSuites:       test.CipherSuites,
-				InsecureSkipVerify: true,
-				MTU:                test.MTU,
+		for _, ciph := range test.CipherSuites {
+			name += "_With" + ciph.String()
+		}
+		test := test
+		t.Run(name, func(t *testing.T) {
```
at-wat: Just run each test parameter as a sub-test.

e2e/e2e_lossy_test.go
```diff
@@ -100,78 +101,88 @@ func TestPionE2ELossy(t *testing.T) {
...
+					serverConn = serverResult.dtlsConn
+				case clientResult := <-clientDone:
+					if clientResult.err != nil {
+						t.Fatalf("Fail, clientError: clientComplete(%t) serverComplete(%t) LossChance(%d) error(%v)", clientConn != nil, serverConn != nil, chosenLoss, clientResult.err)
+					}
+
+					clientConn = clientResult.dtlsConn
+				case <-testTimer.C:
+					t.Fatalf("Test expired: clientComplete(%t) serverComplete(%t) LossChance(%d)", clientConn != nil, serverConn != nil, chosenLoss)
+				case <-time.After(10 * time.Millisecond):
```
at-wat: In JS/WASM environment, `select` loop without sleep looks causing deadlock.

e2e/e2e_lossy_test.go
at-wat: (added to figure out what caused deadlock)

PR #174, opened 2020-01-13
Update CI configs to v0.1.0

pionbot:
Update lint scripts and CI configs.

PR #175, opened 2020-01-13
Fix go.sum

at-wat:
Add missing golang.org/x/crypto v0.0.0-20191227163750-53104e6ec876 entry.

ref: https://github.com/pion/dtls/pull/174


daenney:
How did this happen? A `go get` invocation should've done this automatically.

at-wat:
Seems caused by Renovate Bot.
https://github.com/pion/dtls/blame/master/go.sum

PR #176, opened 2020-01-14
Update CI configs to v0.1.1

pionbot:


.github/lint-filename.sh
```diff
@@ -9,13 +9,13 @@
 set -e
 
 SCRIPT_PATH=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
-GO_REGEX="^[a-zA-Z_]+\.go$"
+GO_REGEX="^[a-zA-Z0-9_]+\.go$"
```
daenney: I'm not sure if this is more convention than required, but I've never seen `.go` files that start with a number, which this would allow. I would prefer a regex like `^[a-zA-Z][a-zA-Z0-9]?\.go$`

daenney: @at-wat Any thoughts on this?

at-wat: I found some files starts with a number in Golang's code.
```
lint/testdata/5_test.go
lint/testdata/4.go
go/src/cmd/compile/internal/ssa/gen/386Ops.go
go/src/cmd/compile/internal/x86/387.go
go/test/64bit.go
go/test/235.go
```
It looks not very usual, so it can be restricted.

Could you fix it at https://github.com/pion/.goassets?

at-wat: Probably, I'm getting flu and going to sleep soon.
Feel free to close current PRs and update .goassets!

daenney: Thanks for chiming in! Do rest up and get better, and thanks for all the work you've done on this. I'll send a PR to goassets

daenney:
As discussed, closing in favor of https://github.com/pion/.goassets/pull/10 and syncing again after.

Issue #148, opened 2019-11-12
Bring us closer to crypto/tls


daenney:
As part of v2 I'd like us to take some time and see if there's more of crypto/tls we can reuse. Mostly want to ensure we don't diverge from interafces and behaviour in crypto/tls unless mandated by the DTLS RFCs. This should help newcomers to immediately get started with pion/dtls as it's hopefully just swapping out crypto/tls for pion/dtls.

at-wat:
I have take a look around `crypto/tls`.
Followings are my suggestion to make `pion/dtls` closer to `crypto/tls`

## Unexpose or move non-DTLS things
### type Closer
It is a package internal utility. It would be better to be unexposed.

### func GenerateSelfSigned(), func SelfSign()
They are for testing and not DTLS specific. It would be better to be unexposed or moved it to somewhere.

### func Fingerprint()
It looks be a part of SDP more than DTLS. Moving it to `pion/sdp` would be better.

## Interfacing
### type Listener
`type Listener` would better be implement `net.Listener`.
`Close(shutdownTimeout time.Duration)` differs from `net.Listener`. Putting `shutdownTimeout` in `Config` struct looks better.

## Misc.
### Clean-up connection timeout config interface
In `crypto/tls`, TCP connection timeout is configured via `Timeout` field in `net.Dialer`. I think adding `dtls.DialWithDialer()` which is compatible with `tls.DialWithDialer()` and making default dialer having 30 seconds of default timeout would be clean and good at compatibility with `crypto/tls`.

at-wat:
> ### func Fingerprint()

Maybe better to move `Fingerprint` and also `HashAlgorithm` under `pkg/crypto`?

at-wat:
## Behavior
### type Listener (#181) ✅ 
- Accepted connections should be available after `Listener.Close()`.

### type Conn (#184) ✅
Pass `nettest.ConnTest`.
- `Read` should return already buffered data even after the connection is closed.
- `SetDeadline` should be implemented.

at-wat:
@daenney @rumpelsepp @jkralik @Sean-Der I have created some PRs to make public APIs more closer to `crypto/tls`. Please add comments if you have any ideas on them.

daenney:
I noticed something in the Go 1.14 releases:

* [ ]: The new CipherSuites and InsecureCipherSuites functions return a list of currently implemented cipher suites
* [ ]: The new CipherSuiteName function returns a name for a cipher suite ID

[...]

PR #181, opened 2020-02-02
Make Listener compatible with net.Listener

at-wat:
Breaking changes:

1. Remove timeout argument from `listener.Close`
Already accepted connections should be active after `listener.Close()`.
Previously, the connections were broken after the timeout.
In the new implementation, `readLoop` routine is remained after
`listener.Close()` and automatically stopped after closing
all connections. The final `conn.Close()` returns after stopping
the routine.

2. ~Return "use of closed connection" error~
~Double `Close()` calls returned `nil`, and `Write()` to closed connection
returned `io.EOF`. Return `"use of closed connection"` error on them to
make the behaviour close to `tls.Conn`.~

3. Unexpose listener implementation
Return as `net.Listener` interface instead of `*dtls.Listener`.

### Reference issue
Part of #148

### DIFF
--- a/internal/udp/conn.go
+++ b/internal/udp/conn.go
@@ -6,11 +6,11 @@ import (
 	"io"
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
 const receiveMTU = 8192
-const closeRetryInterval = 100 * time.Millisecond
 
 var errClosedListener = errors.New("udp: listener closed")
 
@@ -18,20 +18,24 @@ var errClosedListener = errors.New("udp: listener closed")
 type Listener struct {
 	pConn *net.UDPConn
 
-	lock      sync.RWMutex
-	accepting bool
+	accepting atomic.Value // bool
 	acceptCh  chan *Conn
 	doneCh    chan struct{}
 	doneOnce  sync.Once
 
-	conns map[string]*Conn
+	connLock sync.Mutex
+	conns    map[string]*Conn
+	connWG   sync.WaitGroup
+
+	readWG   sync.WaitGroup
+	errClose atomic.Value // error
 }
 
 // Accept waits for and returns the next connection to the listener.
-// You have to either close or read on all connection that are created.
 func (l *Listener) Accept() (*Conn, error) {
 	select {
 	case c := <-l.acceptCh:
+		l.connWG.Add(1)
 		return c, nil
 
 	case <-l.doneCh:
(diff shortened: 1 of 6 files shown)


at-wat:
`Accept()`ed connection looks be allowed to be double closed in `net.TCPConn`.
Relaxing this part.

internal/udp/conn.go
```diff
@@ -91,14 +78,20 @@ func Listen(network string, laddr *net.UDPAddr) (*Listener, error) {
...
+		doneCh:   make(chan struct{}),
 	}
+	l.accepting.Store(true)
+	l.connWG.Add(1)
+	l.readWG.Add(1)
 
 	go l.readLoop()
+	go func() {
+		l.connWG.Wait()
+		_ = l.pConn.Close()
```
Sean-Der: Still reading the code, but this jumps out at me. Anyway we can not ignore this error?

at-wat: It would be better to be returned from the final `Close()` call of `Listener` and `Conn`s. (underlying `Conn` are shared between them.)
https://github.com/pion/dtls/blob/7ee6cc76ae90b3b66c25f9313c21e7c7f725c9ed/internal/udp/conn.go#L58-L61
https://github.com/pion/dtls/blob/7ee6cc76ae90b3b66c25f9313c21e7c7f725c9ed/internal/udp/conn.go#L187-L190

internal/udp/conn.go
```diff
@@ -41,41 +44,25 @@ func (l *Listener) Accept() (*Conn, error) {
 
 // Close closes the listener.
 // Any blocked Accept operations will be unblocked and return errors.
-func (l *Listener) Close(shutdownTimeout time.Duration) error {
-	l.lock.Lock()
-	defer l.lock.Unlock()
-
-	var err error
+func (l *Listener) Close() error {
+	err := errClosedListener
```
at-wat: Our dtls implementation calls `Listener.Close()` multiple times.
It should allow multiple calls for now.

internal/udp/conn.go
at-wat: Fixed to store this error and return from the final `Close()`.

at-wat:
Adding a test to check Close error propagation.

at-wat:
@Sean-Der please take another look when you have time,

Sean-Der:
Fantastic work! We should have been more strict about implementing stdlib interfaces.

I really love in the other PR that we are using the in-tree tests now :)

PR #183, opened 2020-02-02
Unexpose package internal utils

at-wat:
1. Unexpose `Closer`
Make `Closer` internal as it is a package internal utility.
2. Move self sign helper and `CCM` under `pkg/`
They are not parts of DTLS, but useful enough to be exposed.

### Reference issue
Part of #148


Sean-Der:
Great changes! Maybe someone else can find the `CCM` implementation helpful.

If you get a chance I would update `pion/webrtc` to use the new self-signed stuff. Lots of devs just do `master` for all their libs and then hit breakage when stuff doesn't work :/

daenney:
The reason I put CCM in internal is b/c we didn't really write or review the implementation much, though I did add a whole battery of tests. I'm not sure we should make something part of the public interface of this package that wasn't written by us, even though we do use it.

at-wat:
Oh, I didn't see the header well.
https://github.com/pion/dtls/blob/0b80bd9abfa64f92945b81c8acd9407d741a50ed/internal/crypto/ccm/ccm.go#L6-L12
Sounds reasonable to keep unexposed.
Users can import the original one if they want.

daenney:
I'm fine with exposing it to be honest, it's just more of a question of what our stance on this is given we didn't really write the code. The tests are lifted from the RFC that defines CCM, so from that point of view the implementation is at least correct.

The other issue with the original source is that it hasn't been updated to use Go modules, so it might be nice if we do provide it to the larger ecosystem.

at-wat:
@Sean-Der any thoughts?

Sean-Der:
@at-wat I don't feel strongly either way, but I support w/e decision you make :)

IMO a healthy project should have churn/changes, so I don't want you to worry too much about changing things!

Sean-Der:
@daenney that is awesome I completely forgot you added more tests. 

That feels like a life time you got all the PSK stuff in, it is amazing the impact it caused (making all the CoAP stuff happen in pure Go) :)

daenney:
Seems like nobody has strong feelings about CCM, so lets just expose it for now.

at-wat:
@daenney @Sean-Der thank you for your comments!
Let me merge this!

PR #185, opened 2020-02-03
Unexpose HashAlgorithm

at-wat:
Move `Fingerprint` under `pkg/crypto/fingerprint/` and
use `crypto.Hash` instead of `HashAlgorithm`.
Unexpose `HashAlgorithm` representation.

### Reference issue
part of #148


Sean-Der:
:O I love this! 

The more code we hide/delete the better. I really love what you are doing with `pion/dtls`. Sorry I have been distracted, I really want to push Add/Remove (Re-signaling) this week so close :)

Sean-Der:
When this lands would you mind tagging + updating `pion/webrtc` as well? I just want to make sure it doesn't have any side effects, and we don't block people not using modules yet.

pkg/net/fingerprint/fingerprint.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Fingerprint creates a fingerprint for a certificate using the specified hash algorithm
+func Fingerprint(cert *x509.Certificate, algo crypto.Hash) (string, error) {
+	if !algo.Available() {
+		return "", errHashUnavailable
+	}
+	h := algo.New()
+	for i := 0; i < len(cert.Raw); {
+		n, _ := h.Write(cert.Raw[i:])
+		// Hash.Writer never returns an error.
```
Sean-Der: Is this a current implementation detail? It might be worth checking the error + returning, this isn't in the hot path so should be fine!

at-wat: It is defined in https://golang.org/pkg/hash/#Hash `hash.Hash` interface comment. I'll add a link to it.

at-wat:
> When this lands would you mind tagging + updating pion/webrtc as well?

Sure, these are used only at a few places; it will be quick.

daenney:
I'm a bit confused as to why this moves into `pkg/net`? If I look at the Go stdlib, things in `net` directly deal with network related concerns, whereas this is creating a TLS cert hash.

at-wat:
Makes sense. `pkg/crypto` maybe proper location.

at-wat:
Fixed sub-package path.

at-wat:
Rebased.
I will merge it, after the CI passed.

PR #186, opened 2020-02-06
Add fingerprint.StringFromHash

at-wat:
Inverse function of fingerprint.HashFromString.

#### Reference issue
https://github.com/pion/webrtc/pull/1004#discussion_r375470812

### DIFF
--- a/pkg/crypto/fingerprint/hash.go
+++ b/pkg/crypto/fingerprint/hash.go
@@ -23,3 +23,13 @@ func HashFromString(s string) (crypto.Hash, error) {
 	}
 	return 0, errInvalidHashAlgorithm
 }
+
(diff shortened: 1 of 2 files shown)


pkg/crypto/fingerprint/hash.go
```diff
@@ -16,10 +16,26 @@ var nameToHash = map[string]crypto.Hash{
 	"sha-512": crypto.SHA512, // [RFC4055]
 }
 
+var hashToName = make(map[crypto.Hash]string)
+
+func init() {
```
at-wat: gochecknoinits

Sean-Der: I think it is fine to run this every time! I would just have `StringToHash` be more expensive.

PR #187, opened 2020-02-08
Mirror new exported functions from crypto/tls in Go 1.14

daenney:


daenney:
The annoying bit here is that due to `// +build go1.14` CI isn't running the new tests. They pass locally for me with Go 1.14rc1.

We'll need to keep that build tag while we support Go 1.13 and Go 1.14, but at least once Go 1.14 is out we can have CI test this too. Once Go 1.15 hits and we move over we can merge the files together and remove the build constraint.

cipher_suite.go
```diff
@@ -28,6 +28,10 @@ const (
 	TLS_PSK_WITH_AES_128_GCM_SHA256 CipherSuiteID = 0x00a8
 )
 
+var (
+	_ = allCipherSuites() // Necessary until this function isn't only used by Go 1.14
```
daenney: This makes me unhappy but otherwise CI currently breaks since it considers `allCipherSuites()` to be dead code

cipher_suite_test.go
```diff
@@ -21,3 +21,28 @@ func TestDecodeCipherSuites(t *testing.T) {
...
+
+	for _, testCase := range testCases {
+		res := CipherSuiteName(testCase.suite)
+		if res != testCase.expected {
+			t.Fatalf("Expected: %s, got %s", testCase.expected, res)
+		}
+	}
+}
+
+func TestAllCipherSuites(t *testing.T) {
```
daenney: This is an extremely silly test. Doesn't seem like it would ever catch anything, but it keeps the coverage up :man_shrugging:

cipher_suite.go
```diff
@@ -68,6 +72,16 @@ type cipherSuite interface {
 	decrypt(in []byte) ([]byte, error)
 }
 
+// CipherSuiteName provides the same functionality as tls.CipherSuiteName
+// that appeared first in Go 1.14
+func CipherSuiteName(id uint16) string {
```
daenney: I kept the signature identical to stdlib, but I'm thinking that it might make more sense to make this `func CipherSuiteName(id CipherSuiteID) string` since we require the use of a `CipherSuiteID` everywhere instead of a `uint16`.

cipher_suite_go114.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"crypto/tls"
+)
+
+// Convert from our cipherSuite interface to a tls.CipherSuite struct
+func toTLSCipherSuite(c cipherSuite) *tls.CipherSuite {
+	return &tls.CipherSuite{
+		ID:                uint16(c.ID()),
+		Name:              c.String(),
+		SupportedVersions: []uint16{0xfefd},
```
at-wat: Would be nice to have public const like https://github.com/golang/go/blob/a52821569310ca3807f74993af4019dfcaa74cfe/src/crypto/tls/common.go#L29-L38.

cipher_suite_test.go
at-wat: `if actual == 0 { t.Fatalf() }` may be enough for this purpose?

cipher_suite_go114_test.go
```diff
@@ -0,0 +1,49 @@
...
+	ours := allCipherSuites()
+	theirs := CipherSuites()
+
+	if len(ours) != len(theirs) {
+		t.Fatalf("Expected %d CipherSuites, got %d", len(ours), len(theirs))
+	}
+
+	for i, s := range ours {
+		t.Run(s.String(), func(t *testing.T) {
+			c := theirs[i]
```
at-wat: golangci-lint on Go1.14 seems saying:
```
$ golangci-lint run
cipher_suite_go114.go:23:2: Consider preallocating `res` (prealloc)
	var res []*tls.CipherSuite
	^
cipher_suite_go114_test.go:27:16: Using the variable on range scope `i` in function literal (scopelint)
			c := theirs[i]
			            ^
cipher_suite_go114_test.go:28:22: Using the variable on range scope `s` in function literal (scopelint)
			if c.ID != uint16(s.ID()) {
			                  ^
cipher_suite_go114_test.go:29:49: Using the variable on range scope `s` in function literal (scopelint)
				t.Fatalf("Expected ID: 0x%04X, got 0x%04X", s.ID(), c.ID)
				                                            ^
cipher_suite_go114_test.go:32:17: Using the variable on range scope `s` in function literal (scopelint)
			if c.Name != s.String() {
```
scopelint errors are not actual problem on non Parallel tests.

cipher_suite.go
at-wat: I don't have strong opinion on this, but using `CipherSuiteID` with a comment about type difference maybe fine.

cipher_suite_go114.go
daenney: So we have a `protocolVersion`, but it's a struct with `dtlsMajor` and `dtlsMinor`. Also unexported.

I suppose we could add a `VersionDTLS12` to match stdlib?

cipher_suite_go114_test.go
daenney: I believe I've fixed them all, for good measure.

cipher_suite_test.go
daenney: Yap, seems fair. Changed it to that now.

daenney:
@at-wat Thanks for the review! I think I managed to address all of the comments.

cipher_suite.go
daenney: Yap, I've changed it. I think it makes more sense that way.

PR #191, opened 2020-02-11
Support context.Context based cancellation

### LINKED ISSUE #189: context.Context based cancellation
Proposed at https://github.com/pion/dtls/issues/147.

https://github.com/pion/dtls/issues/147#issuecomment-580080132
> @Sean-Der:
> What do you think of updating Client/Server to take a context? I need a way to be notified that an error has occurred, and it looks like taking a `Context` is the proper way to have a async way to bubble up an error. I talk about it in #151 a bit.

at-wat:
`DialWithContext()`, `ClientWithContext()` and `ServerWithContext()`
are added. Instead, `ConnectTimeout` is dropped from `Config`.

### Reference issue
Closes #189

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,6 +1,7 @@
 package dtls
 
 import (
+	"context"
 	"crypto/ecdsa"
 	"crypto/tls"
 	"crypto/x509"
@@ -83,13 +84,30 @@ type Config struct {
 
 	LoggerFactory logging.LoggerFactory
 
+	// ConnectContextMaker is a function to make a context used in Dial(),
+	// Client(), Server(), and Accept(). If nil, the default ConnectContextMaker
+	// is used. It can be implemented as following.
+	//
+	// 	func ConnectContextMaker() (context.Context, func()) {
+	// 		return context.WithTimeout(context.Background(), 30*time.Second)
+	// 	}
+	ConnectContextMaker func() (context.Context, func())
+
 	// MTU is the length at which handshake messages will be fragmented to
 	// fit within the maximum transmission unit (default is 1200 bytes)
 	MTU int
 }
 
-// DefaultConnectTimeout is a timeout duration used in Dial(), Client() and Server().
-const DefaultConnectTimeout = 30 * time.Second
+func defaultConnectContextMaker() (context.Context, func()) {
+	return context.WithTimeout(context.Background(), 30*time.Second)
+}
+
+func (c *Config) connectContextMaker() (context.Context, func()) {
+	if c.ConnectContextMaker == nil {
+		return defaultConnectContextMaker()
+	}
+	return c.ConnectContextMaker()
+}
 
 const defaultMTU = 1200 // bytes
 
--- a/config.go
+++ b/config.go
@@ -83,21 +83,13 @@ type Config struct {
 
 	LoggerFactory logging.LoggerFactory
 
-	// ConnectTimeout is the timeout threshold for new connection handshakes
-	// to complete (default is 30 seconds)
-	ConnectTimeout *time.Duration
-
 	// MTU is the length at which handshake messages will be fragmented to
 	// fit within the maximum transmission unit (default is 1200 bytes)
 	MTU int
 }
 
-const defaultConnectTimeout = 30 * time.Second
-
-// ConnectTimeoutOption simply provides a wrapper for creating a *time.Duration
-func ConnectTimeoutOption(timeout time.Duration) *time.Duration {
-	return &timeout
-}
+// DefaultConnectTimeout is a timeout duration used in Dial(), Client() and Server().
+const DefaultConnectTimeout = 30 * time.Second
 
 const defaultMTU = 1200 // bytes
 
--- a/errors.go
+++ b/errors.go
@@ -1,6 +1,11 @@
 package dtls
 
-import "errors"
+import (
+	"context"
+	"errors"
+
+	"golang.org/x/xerrors"
+)
 
 // Typed errors
 var (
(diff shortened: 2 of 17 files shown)


config.go
```diff
@@ -83,20 +84,25 @@ type Config struct {
 
 	LoggerFactory logging.LoggerFactory
 
-	// ConnectTimeout is the timeout threshold for new connection handshakes
-	// to complete (default is 30 seconds)
-	ConnectTimeout *time.Duration
+	// function to make a context used in Dial(), Client(), Server(), and Accept().
+	// If nil, DefaultConnectContextMaker is used.
+	ConnectContextMaker func() (context.Context, func())
```
at-wat: This function type suggest user to use cancelable contexts.

errors.go
```diff
@@ -50,11 +54,27 @@ var (
...
 	errServerNoMatchingSRTPProfile       = errors.New("dtls: Client requested SRTP but we have no matching profiles")
 	errServerRequiredButNoClientEMS      = errors.New("dtls: Server requires the Extended Master Secret extension, but the client does not support it")
 	errClientRequiredButNoServerEMS      = errors.New("dtls: Client required Extended Master Secret extension, but server does not support it")
 	errInvalidCertificate                = errors.New("dtls: No certificate provided")
+
+	// Wrapped errors
+	errConnectTimeout = &wrappedError{"dtls: The connection timed out during the handshake", context.DeadlineExceeded}
 )
+
+type wrappedError struct {
```
at-wat: Use `fmt.Errorf("...: %w")` instead after Go 1.12 EOL.

daenney: Perhaps we could use https://godoc.org/golang.org/x/xerrors over our own type?

at-wat: Definitely! I'll switch to use xerrors.

at-wat:
@daenney @Sean-Der @jkralik please take a look when you have time!

at-wat:
Increased context related test coverage.

config.go
```diff
@@ -83,20 +84,25 @@ type Config struct {
...
 
 	// MTU is the length at which handshake messages will be fragmented to
 	// fit within the maximum transmission unit (default is 1200 bytes)
 	MTU int
 }
 
-const defaultConnectTimeout = 30 * time.Second
+// DefaultConnectContextMaker is a default ConnectContextMaker used in Dial(), Client(), Server(), and Accept().
+var DefaultConnectContextMaker = func() (context.Context, func()) {
```
Sean-Der: Mind making this a function? I don't believe modification is anything we will ever need!

config.go
Sean-Der: I am fine removing this entirely! ping @daenney make sure I am not making a bad decision :)

* `Context` with a deadline if they still need deadlines.
* We don't support Dialer, so not like we are breaking any APIs (Dialer also doesn't support server so will not work for us)

Sean-Der:
With the next context API is it possible to be notified of a Close now without doing a `Read`? This is something WebRTC needs. I have a case where there is no SCTP traffic, so the DTLS alert is the fastest way to kill the connection. I was hoping I could do.

```
ctx, _ := context.WithCancel(context.Background())
c, err := dtls.Client(...)
if err != nil {
  return err
}

go func() {
   <-ctx.Done()
   if ctx.Err() != nil {
     fmt.Println("DTLS Session has ended in error!")
   }
}()


```

config.go
at-wat: In stdlib, default things are often defined as global variable, e.g. `net.DefaultResolver`: https://golang.org/pkg/net/#pkg-variables

One of my intention of this was providing an example of ConnectContextMaker implementation on godoc.
I'll make it a function, unexpose, and explicitly add an example as a comment.

at-wat:
@Sean-Der unfortunately, https://github.com/pion/dtls/pull/191#issuecomment-586484133 is not possible.
`context.Context` interface doesn't provide a way to cancel itself.
`ctx, _ := context.WithCancel(context.Background())` is never been cancelled and just causes context leak.

I think it's better to add a state callback for this purpose.

config.go
Sean-Der: Sorry I am wrong! Wasn't aware this was a Go pattern. I just always avoid globals and the possibility of people modifying things. 

 If you want to revert and go back to the way you had it I am happy with that :)

at-wat: updated

at-wat: But yea, it may allow something bad. If one of the library does
```go
func init() {
    net.DefaultResolver = &net.Resolver{
        Dial: EvilDial,
    }
}
```
It's weird.

Sean-Der:
@at-wat I am happy with merging! We can always improve things as we go, such a shame about the leakage :( 

I will add the state callback soon, need it for good WebRTC behavior

config.go
at-wat: I added this to make it possible to customize the behavior of accepted connection in `dtls.listener`.

As `Listen()` blocks until incoming connection and return established `dtls.Conn`, deadline timer must be initialized at the time the first packet, that triggers dtls handshake, arrives.

daenney: Thanks for the ping. I think this makes sense, and the fact that when not explicitly provided we inject a default context with a proper timeout makes me happy.

config.go
daenney: So I agree this is a pattern in stdlib, but I think this is a place where we should deviate from stdlib, and I'm happy to see the `defaultConnectContextMaker` instead.

I understand why it's historically shown up in stdlib, net/http does something similar with `DefaultHTTPClient` so you can call a naked `http.$Method` without needing to create your own `http.Client` first, but it's proven to be quite the footgun.

at-wat:
Thank you for reviewing!

PR #194, opened 2020-02-18
Add Extension Server Name Indication

juliens:
#### Description
Add Server Name Indication extension following the [RFC 6066 section 3](https://tools.ietf.org/html/rfc6066#section-3).

I change the way with choose the Certificate to use by comparing the server name with the CN and SANs.

This PR is based on the implementation in TLS in the standard lib.


conn.go
```diff
@@ -278,6 +301,44 @@ func ServerWithContext(ctx context.Context, conn net.Conn, config *Config) (*Con
...
+		// There's only one choice, so no point doing any work.
+		return &c.localCertificates[0], nil
+	}
+
+	if len(serverName) == 0 {
+		return &c.localCertificates[0], nil
+	}
+
+	name := strings.ToLower(serverName)
+	for len(name) > 0 && name[len(name)-1] == '.' {
```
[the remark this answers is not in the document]
at-wat: BTW, do we need to fix the server name automatically?

https://tools.ietf.org/html/rfc6066#page-7
>"HostName" contains the fully qualified DNS hostname of the server,
> as understood by the client.  The hostname is represented as a byte
> string using ASCII encoding without a trailing dot.

I think the hostname with trailing dot is an error.

conn.go
```diff
@@ -128,6 +131,25 @@ func createConn(ctx context.Context, nextConn net.Conn, flightHandler flightHand
...
+		x509Cert := cert.Leaf
+		if x509Cert == nil {
+			var parseErr error
+			x509Cert, parseErr = x509.ParseCertificate(cert.Certificate[0])
+			if parseErr != nil {
+				continue
+			}
+		}
+		if len(x509Cert.Subject.CommonName) > 0 {
+			nameToCertificate[x509Cert.Subject.CommonName] = cert
```
at-wat:
```suggestion
			nameToCertificate[strings.ToLower(x509Cert.Subject.CommonName)] = cert
```

conn.go
```diff
@@ -128,6 +131,25 @@ func createConn(ctx context.Context, nextConn net.Conn, flightHandler flightHand
...
+		}
+		for _, san := range x509Cert.DNSNames {
+			nameToCertificate[san] = cert
```
at-wat:
```suggestion
			nameToCertificate[strings.ToLower(san)] = cert
```

conn.go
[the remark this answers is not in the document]
juliens: You're right, the serverName cannot have a trailing dot in the server part, but as the `getCertificate` is used in the client part too, and as the serverName came from the configuration in the client part, it's safer to remove it. But maybe we can reject the config in the validation if there is a trailing dot? WDYT?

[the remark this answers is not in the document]
daenney: I'm inclined to say it's fine in this case. The with or without trailing dot is a technicality from my point of view. If your code can make the user not have to care/deal with it, I'd prefer that over having to throw an error.

Additionally, it seems relatively unlikely to me someone would configure it with a trailing dot.

extension_server_name.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	for !nameList.Empty() {
+		var nameType uint8
+		var serverName cryptobyte.String
+		if !nameList.ReadUint8(&nameType) ||
+			!nameList.ReadUint16LengthPrefixed(&serverName) ||
+			serverName.Empty() {
+			return errInvalidSNIFormat
+		}
+		if nameType != 0 {
```
at-wat: This 0 seems also `extensionServerNameTypeDNSHostName`.

juliens: done :+1:

daenney:
For some weird reason the code coverage bot seems to not want to show up on this PR. Here's the current status of this PR: https://codecov.io/gh/pion/dtls/pull/194/changes

conn.go
```diff
@@ -778,7 +836,7 @@ func (c *Conn) startHandshakeOutbound() {
 
 func (c *Conn) close() error {
 	if c.connectionClosed.Err() == nil && c.handshakeErr.load() == nil {
-		c.notify(alertLevelWarning, alertCloseNotify) //nolint
+		c.notify(alertLevelWarning, alertCloseNotify) // nolint
```
daenney: Need to drop this change, causes a merge conflict b/c the code on master has been updated.

jkralik:
Great job 👍

PR #196, opened 2020-02-20
Send close_notify as warning level

### LINKED ISSUE #195: Send close_notify as warning level
`pion/dtls` sends `close_notify` as fatal level at now, but OpenSSL sends `close_notify` as warning level. OpenSSL shows `close_notify` from `pion/dtls` as an error log.

`close_notify` was specified as warning level in TLS 1.0 RFC.
Later TLS RFC doesn't explicitly specify the level of `close_notify`.

https://tools.ietf.org/html/rfc2246#section-7.2.1
> ```
>    close_notify
>       This message notifies the recipient that the sender will not send
>       any more messages on this connection. The session becomes
>       unresumable if any connection is terminated without proper
>       close_notify messages with level equal to warning.
> ```

TLS 1.1 RFC
https://tools.ietf.org/html/rfc4346#section-7.2.1
> ```
>    close_notify
>       This message notifies the recipient that the sender will not send
>       any more messages on this connection.  Note that as of TLS 1.1,
>       failure to properly close a connection no longer requires that a
>       session not be resumed.  This is a change from TLS 1.0 to conform
>       with widespread implementation practice.
> ```

TLS 1.2 RFC
https://tools.ietf.org/html/rfc5246#section-7.2.1
> ```
>    close_notify
>       This message notifies the recipient that the sender will not send
>       any more messages on this connection.  Note that as of TLS 1.1,
>       failure to properly close a connection no longer requires that a
>       session not be resumed.  This is a change from TLS 1.0 to conform
>       with widespread implementation practice.
> ```

It sounds that TLS 1.0's spec about `close_notify` level is not denied in the latest specs.

at-wat:
OpenSSL sends close_notify as warning level and expects that incoming close_notify is warning level, so, OpenSSL showed close_notify from pion/dtls as an error.

close_notify was specified as warning level in TLS 1.0 RFC2246. Later TLS 1.1/1.2 RFC4346/5246 doesn't explicitly specify the level of close_notify, but they don't deny TLS 1.0's spec.

### Reference issue
Fixes #195

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -778,7 +778,7 @@ func (c *Conn) startHandshakeOutbound() {
 
 func (c *Conn) close() error {
 	if c.connectionClosed.Err() == nil && c.handshakeErr.load() == nil {
-		c.notify(alertLevelFatal, alertCloseNotify) //nolint
+		c.notify(alertLevelWarning, alertCloseNotify) //nolint
 	}
 
 	c.workerTicker.Stop()


Sean-Der:
:+1: from me! Thank you @at-wat :)

at-wat:
I found why it is warning level.
In RFC 5246, server/client must respond to a incoming close_notify with a close_notify of its own. As both parties immediately close the connection on fetal alert, sending close_notify can't adapt the protocol spec.
https://github.com/pion/dtls/issues/197

at-wat:
I'll update the commit message and then merge.

PR #198, opened 2020-02-21
Respond to close_notify with close_notify

### LINKED ISSUE #197: Respond to close_notify with close_notify
https://tools.ietf.org/html/rfc5246#section-7.2.1
>   ```
>   Unless some other fatal alert has been transmitted, each party is
>   required to send a close_notify alert before closing the write side
>   of the connection.  The other party MUST respond with a close_notify
>   alert of its own and close down the connection immediately,
>   discarding any pending writes.  It is not required for the initiator
>   of the close to wait for the responding close_notify alert before
>   closing the read side of the connection.
>   ```

`The other party MUST respond with a close_notify`

https://github.com/pion/dtls/blob/34ebfdba22d430240630610d70edf371d8b121fc/conn.go#L667-L672

at-wat:
RFC5246 Section 7.2.1 specifies that one received a close_notify alert
MUST respond with a close_notify alert and close down the connection
immediately.

### Reference issue
Fixes #197

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -667,6 +667,8 @@ func (c *Conn) handleIncomingPacket(buf []byte) (*alert, error) {
 	case *alert:
 		c.log.Tracef("<- %s", content.String())
 		if content.alertDescription == alertCloseNotify {
+			// Respond with a close_notify [RFC5246 Section 7.2.1]
+			_ = c.notify(alertLevelWarning, alertCloseNotify)
 			return nil, c.Close()
 		}
 		return nil, fmt.Errorf("alert: %v", content)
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -667,6 +667,8 @@ func (c *Conn) handleIncomingPacket(buf []byte) (*alert, error) {
 	case *alert:
 		c.log.Tracef("<- %s", content.String())
 		if content.alertDescription == alertCloseNotify {
+			// Respond with a close_notify [RFC5246 Section 7.2.1]
+			c.notify(alertLevelWarning, alertCloseNotify) //nolint
```
igolaizola: are you adding `//nolint` to skip the error processing? You can do `_ = c.notify` instead

at-wat: agree with this.
I'll also fix one in `close()` where I copied it from.

Issue #199, opened 2020-02-21
CCM ciphersuite causes a lot of CPU load


cohosh:
### Your environment.
  - Version: v2.1.18
  - Browser: N/A (Snowflake)
  - Other Information: CPU profiling output given below

### What did you do?
We've been running Snowflake with the pion/webrtc library for a while now. We noticed that our Snowflake proxies were using a lot more CPU than expected in sending and receiving WebRTC messages.

More info on snowflake: https://snowflake.torproject.org

### What did you expect?

We expected the CPU usage to be around 10-20% for proxying a single client's traffic.

### What happened?

Our CPU usage for a single client was over 60%.

After some investigations, we tracked a lot of CPU heavy operations coming from the CCM ciphersuite. Changing the ciphersuite to GCM resulted in the more expected usage of 10-20%.

See our ticket here: https://trac.torproject.org/projects/tor/ticket/33211#comment:14

I'd suggest
1) allowing a way for application develoeprs to change the ciphersuite (as filed here: https://github.com/pion/webrtc/issues/1043)

2) taking a look at the CCM implementation to see if there are some optimizations available there, and

3) using a different default ciphersuite and update the priority listing of available ciphersuites. GCM is supposed to be slightly more efficient, and is much more commonly seen in practice.

Sean-Der:
My guess is [this list](https://github.com/pion/dtls/blob/master/cipher_suite.go#L116) is used backwards, and we are preferring the wrong CipherSuites. 

I will look into that, and make sure we get a test in. Thanks for filing this @cohosh! I doubt we will ever get close to beating GCM Go, the stdlib use AES-NI and has a bunch of scrutiny :)

cohosh:
Yep, that seems plausible. This is pion-to-pion traffic so the ciphersuite used **should** be the highest priority one listed there.

cohosh:
It could also be that [https://github.com/pion/dtls/blob/master/cipher_suite.go#L127](this list of all cipher suites) is being used instead.

Sean-Der:
@cohosh Fixed! Mind reviewing that PR, I will merge + tag across.

Sean-Der:
![Screen Shot 2020-02-21 at 2 26 26 PM](https://user-images.githubusercontent.com/1302304/75076978-50869600-54b6-11ea-8607-a246340531f0.png)

Sean-Der:
[...]

PR #200, opened 2020-02-21
Fix encodeCipherSuites ordering

Sean-Der:
Before encodeCipherSuites would improperly encode backwards, putting
our least preferred suite first. Fix this function so it properly encodes
and update TestHandshakeMessageClientHello so it has multiple CipherSuites
to make sure this doesn't regress.

Relates to #199


cipher_suite.go
```diff
@@ -115,10 +115,10 @@ func cipherSuiteForID(id CipherSuiteID) cipherSuite {
 // CipherSuites we support in order of preference
 func defaultCipherSuites() []cipherSuite {
 	return []cipherSuite{
-		&cipherSuiteTLSEcdheRsaWithAes256CbcSha{},
-		&cipherSuiteTLSEcdheEcdsaWithAes256CbcSha{},
 		&cipherSuiteTLSEcdheRsaWithAes128GcmSha256{},
 		&cipherSuiteTLSEcdheEcdsaWithAes128GcmSha256{},
```
at-wat: https://github.com/pion/dtls/issues/199#issuecomment-589863610
In Chromium, ECDSA is prior to RSA?

Sean-Der: :+1: will put ECDSA first.

PR #201, opened 2020-02-22
Remove CCM from default cipher suites

### LINKED ISSUE #199: CCM ciphersuite causes a lot of CPU load

Sean-Der:
If users want CCM they need to explicitly request it.

Resolves #199

### DIFF
--- a/cipher_suite.go
+++ b/cipher_suite.go
@@ -119,8 +119,6 @@ func defaultCipherSuites() []cipherSuite {
 		&cipherSuiteTLSEcdheRsaWithAes128GcmSha256{},
 		&cipherSuiteTLSEcdheEcdsaWithAes256CbcSha{},
 		&cipherSuiteTLSEcdheRsaWithAes256CbcSha{},
-		newCipherSuiteTLSEcdheEcdsaWithAes128Ccm(),
-		newCipherSuiteTLSEcdheEcdsaWithAes128Ccm8(),
 	}
 }
 


Sean-Der:
@jkralik just want to make sure this isn't a breaking change/doesn't effect you negatively.

I think it is in users best interested to have GCM -> CCM as the defaults.

jkralik:
It's change default behavior but I'm ok with that.

PR #193, opened 2020-02-17
Add end to end tests using OpenSSL

igolaizola:
End to end tests are now executed in the folowing modes:

 - `default`: pion client, pion server
 - `openssl_client`: openssl client, pion server
 - `openssl_server`: pion client, openssl server

OpenSSL is launched using `exec.Command`

TODO:

- [x] Changes to CI to install OpenSSL
- [x] Make OpenSSL server work with PSK tests
- [ ] Make OpenSSL work with ed25519 test

PR #203, opened 2020-02-24
Add key usage cert-sign to selfsigned certs

igolaizola:
Without this key usage the selfsigned certificate cannot be used to validate itself.

OpenSSL E2E tests https://github.com/pion/dtls/pull/193 are throwing the following messages because of this:

```
verify error:num=20:unable to get local issuer certificate
```


daenney:
This seems fine by me. The SelfSign helpers are really just meant for testing purposes and ease of local development, so we can be a bit more lax here.

PR #209, opened 2020-03-05
Refactor errors

at-wat:
Wrap all errors to implement net.Error.
Caller of Read/Write can check `err.(net.Error).Temporary()` to determine whether the error is fatal or not, as with raw `net.UDPConn`.

Split `errInvalidPacketLength` from `errLengthMismatch`.
`errInvalidPacketLength` is used for errors caused by incoming data, which must be discarded, and `errLengthMismatch` is used for errors due to the data generated by the library, which is a fatal internal error.

### DIFF
--- a/errors_errno.go
+++ b/errors_errno.go
@@ -1,0 +1,25 @@
+// +build aix darwin dragonfly freebsd linux nacl nacljs netbsd openbsd solaris windows
+
+// For systems having syscall.Errno.
+// Update build targets by following command:
+// $ grep -R ECONN $(go env GOROOT)/src/syscall/zerrors_*.go \
+//     | tr "." "_" | cut -d"_" -f"2" | sort | uniq
+
+package dtls
+
+import (
+	"os"
+	"syscall"
+)
+
+func isOpErrorTemporary(err *os.SyscallError) bool {
+	if ne, ok := err.Err.(syscall.Errno); ok {
+		switch ne {
+		case syscall.ECONNREFUSED:
+			return true
+		default:
+			return false
+		}
+	}
+	return false
+}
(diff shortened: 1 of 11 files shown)


errors_errno.go
```diff
@@ -0,0 +1,22 @@
+// +build aix darwin dragonfly freebsd linux nacl nacljs netbsd openbsd solaris windows
```
daenney: It feels like we're likely to forget to update this along the way, given how big the build tag is. Can we do something shorter? It seems like it almost builds for everything.

record_layer.go
```diff
@@ -77,12 +77,12 @@ func unpackDatagram(buf []byte) ([][]byte, error) {
...
 	for offset := 0; len(buf) != offset; {
 		if len(buf)-offset <= recordLayerHeaderSize {
-			return nil, errDTLSPacketInvalidLength
+			return nil, errInvalidPacketLength
 		}
 
 		pktLen := (recordLayerHeaderSize + int(binary.BigEndian.Uint16(buf[offset+11:])))
 		if offset+pktLen > len(buf) {
-			return nil, errLengthMismatch
+			return nil, errInvalidPacketLength
```
daenney: Why `errInvalidPackageLength` instead of `errLengthMismatch`?

at-wat: There were two usage of `errLengthMismatch`.
One is caused by internal implementation, which must be treated as a fatal error.
Another one is caused by incoming data and DTLS must ignore error due to incoming data as written in RFC. https://tools.ietf.org/html/rfc6347#section-4.1.2.7

errors_errno.go
at-wat: It's actual list that have specific errno. For example js doesn't.

daenney: I realise that. But wouldn't it be easier to do `!js !z390` or something?

at-wat: I don't think there is a safe way to make it shorter.
Added a comment to generate the list.

at-wat:
rebased

PR #210, opened 2020-03-05
Limit retransmit interval of last flight

at-wat:
Also fix retransmit interval parameter in TestHandshaker. It sometimes caused WASM test failure.

### DIFF
--- a/handshaker_test.go
+++ b/handshaker_test.go
@@ -12,6 +12,8 @@ import (
 	"github.com/pion/transport/test"
 )
 
+const nonZeroRetransmitInterval = 100 * time.Millisecond
+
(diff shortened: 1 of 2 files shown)


handshaker_test.go
```diff
@@ -51,6 +52,7 @@ func TestHandshaker(t *testing.T) {
 					cancelCli()
 				}
 			},
+			retransmitInterval: 100 * time.Millisecond,
```
daenney: Is there a reason why it's 100? Should this maybe be a constant so the name can document its purpose?

at-wat: Added const. It just must be non-zero.

PR #211, opened 2020-03-06
Add replay attack protection

### LINKED ISSUE #146: Add sliding window replay detection for ApplicationData
Credit to @ic0ns @pfg666 and @jurajsomorovsky for this report, thank you!

---

Implement https://tools.ietf.org/html/rfc6347#section-4.1.2.6

There is no threat during the handshake (random data protects against replays) but currently we are depending on the layer above (SCTP in the WebRTC case for this)

at-wat:
Implements RFC 6347 Section 4.1.2.6.
Set `config.ReplayProtectionWindow` to change the size of the protection window.
Default is 64.

### Reference issue
Fixes #146


at-wat:
CI failure on WASM will be fixed by #210.

conn.go
```diff
@@ -89,13 +92,22 @@ func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient
...
+	logger := loggerFactory.NewLogger(loggerName)
 
 	mtu := config.MTU
 	if mtu <= 0 {
 		mtu = defaultMTU
 	}
 
+	replayProtectionWindow := config.ReplayProtectionWindow
+	if replayProtectionWindow <= 0 {
+		replayProtectionWindow = 64
```
Sean-Der: Mind making the 64 a const? Just so people can easily find.

conn.go
```diff
@@ -89,13 +92,22 @@ func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient
 		loggerFactory = logging.NewDefaultLoggerFactory()
 	}
 
-	logger := loggerFactory.NewLogger("dtls")
+	loggerName := "dtls server"
+	if isClient {
+		loggerName = "dtls client"
+	}
+	logger := loggerFactory.NewLogger(loggerName)
```
Sean-Der: I don't have a problem letting people pass this in, but probably shouldn't make this change now. People collect logs off of `dtls` right now, so would break them.

conn.go
```diff
@@ -623,24 +641,57 @@ func (c *Conn) handleQueuedPackets(ctx context.Context) error {
...
+		return false, nil, nil
+	}
+
+	// Anti-replay protection
+	for len(c.state.replayDetector) <= int(h.epoch) {
+		c.state.replayDetector = append(c.state.replayDetector,
+			replaydetector.New(c.replayProtectionWindow, maxSequenceNumber),
+		)
+	}
+	accept, ok := c.state.replayDetector[int(h.epoch)].Check(h.sequenceNumber)
```
Sean-Der: mind renaming `accept` so it makes it more clear it is accepting it for the relayDetector? All the calls to `accept` below might confuse people who aren't familiar.

conn.go
```diff
@@ -623,24 +641,57 @@ func (c *Conn) handleQueuedPackets(ctx context.Context) error {
...
+		}
+		if enqueue {
+			c.log.Debug("received packet of next epoch, queuing packet")
+			c.encryptedPackets = append(c.encryptedPackets, buf)
+		}
+		return false, nil, nil
+	}
+
+	// Anti-replay protection
+	for len(c.state.replayDetector) <= int(h.epoch) {
```
Sean-Der: Could an attack DDoS/Overflow us by sending a very large epoch?

internal/replaydetector/replaydetector_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func TestReplayDetector(t *testing.T) {
+	cases := map[string]struct {
+		windowSize uint
+		maxSeq     uint64
+		input      []uint64
+		expected   []uint64
+	}{
+		"Continuous": {16, 0x0000FFFFFFFFFFFF,
```
Sean-Der: Is it worth adding a test for the window being moved forward?

```
   If the received record falls within the window and is new, or if the
   packet is to the right of the window, then the receiver proceeds to
   MAC verification.  If the MAC validation fails, the receiver MUST
   discard the received record as invalid.  The receive window is
   updated only if the MAC verification succeeds.
```

conn.go
at-wat: I'll just revert log related changes. I added this to check trace on CI as it is very hard to see the trace log, but it is not needed in usual applications.

internal/replaydetector/replaydetector.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+}
+
+func (d *slidingWindowDetector) Check(seq uint64) (accept func(), ok bool) {
+	if seq+d.windowSize < d.latestSeq || seq > d.maxSeq {
+		// Older than window size or exceeding upper limit.
+		return func() {}, false
+	}
+	if seq > d.latestSeq {
+		// Update the head of the window.
```
at-wat: Window shift must be done after validation.

internal/replaydetector/replaydetector.go
```diff
@@ -0,0 +1,59 @@
...
+		shift := seq - d.latestSeq
+		for i := len(d.mask) - 1; i > 0; i-- {
+			d.mask[i] <<= shift
+			d.mask[i] |= d.mask[i-1] >> (64 - shift)
```
at-wat: This loose some bits if window size is larger than 128 and window shift is larger than 64.

at-wat:
Addressed review feedbacks and rebased.

at-wat:
rebased

PR #212, opened 2020-03-07
Fix flaky test on WASM environment

at-wat:
In JS/WASM environment, select loop without sleep seems causing deadlock. Add sleep to trigger context switch.

### Reference issue
Same fix as #173


at-wat:
Re-ran CI 10 times. The error doesn't appear.

PR #214, opened 2020-03-08
Close on outgoing sequence number overflow

at-wat:
Implementation must either abandon an association or rehandshake. RFC 6347 Section 4.1.
Close connection for now.


at-wat:
rebased

Sean-Der:
@at-wat sorry this PR has been sitting so long!

IMO if you are left waiting more then 3 days, I am fine with you merging. I would rather see progress forward then you get block on other people :(

PR #215, opened 2020-03-08
Verify protocol version

### LINKED ISSUE #3: Implement strict version checking
We should check the version of the remote we are communicating with (and throw an error) right now we depend on the other side blowing up if they are `v1.0`

at-wat:
Check version in handshake message.
Silently discard records with unsupported version in record layer header.

~Branched from #209 in order to add a new error.~ rebased

### Reference issue
Fixes #3


daenney:
This PR is a bit impossible to review due to it also containing all changes from #209. I'll take a look once that's been rebased.

at-wat:
Rebased

PR #225, opened 2020-03-13
Docs/example listen cast

vidavidorra:
#### Description

Add a comment about `dtls.Conn` cast in listen example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Robert Eperjesi](https://github.com/epes)
 * [Atsushi Watanabe](https://github.com/at-wat)
 * [Julien Salleyron](https://github.com/juliens) - *Server Name Indication*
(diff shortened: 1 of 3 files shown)


Sean-Der:
@vidavidorra mind adding the same to `listen-psk` then LGTM!

Thanks for making things easier for people in the future :)

Sean-Der:
Don't worry about the linter for this commit. I will squash everything for you

vidavidorra:
> Thanks for making things easier for people in the future :)

No worries, happy to help!

> Don't worry about the linter for this commit. I will squash everything for you

I think the problem is the git hooks not being set up. How are those supposed to be set up locally? I'm not familiar with this since I've always used npm packages like husky to handle hooks for me.

Sean-Der:
Merged with d5cce09442d3c5db4a18e6e170de3c09f20508f2 fantastic work!

Sean-Der:
If you run `.github/install-hooks.sh` it *should* set them up.

You can also see everything we do in the `.travis.yml`. I am gone for 8 hours (bed time for me) but excited to get more people involved :)

PR #226, opened 2020-03-15
Concurrently read and accept UDP packet

### LINKED ISSUE #190: Can't accept new conn if first conn do not call Read()
https://github.com/pion/dtls/blob/34883a71275c351e804377df69a318fe4c2edc10/internal/net/udp/conn.go#L129

can't accept new conn if first conn do not call Read()

https://github.com/pion/dtls/blob/34883a71275c351e804377df69a318fe4c2edc10/internal/net/udp/conn.go#L180

### LINKED ISSUE #216: Move replaydetector and deadline internal packages to pion/transport
pion/srtp also need to have replay protection. As DTLS and SRTP is completely independent, `replaydetector` should be in the other package.
https://github.com/pion/transport/pull/59

`deadline` internal package helps to implement `net.Conn.SetDeadline` and makes Conn compatible with `net.Conn`. `pion/transport/test.bridgeConn` and `pion/transport/packetio.Buffer` would also be better to be compatible with `net.Conn`. `pion/webrtc/internal/mux` backended by `packetio.Buffer`. It will make multiplexed DTLS and SRTP connection properly deadline controllable.
https://github.com/pion/transport/pull/58

at-wat:
Queue accept and buffer read.
Read performance optimization is not included in this PR.

### Reference issue
Fixes #190 
Closes #216 
~Requires https://github.com/pion/transport/pull/58~ merged


at-wat:
@adwpc @Sean-Der please take a look.

internal/net/udp/conn.go
```diff
@@ -74,20 +89,35 @@ func (l *Listener) Close() error {
...
 // Addr returns the listener's network address.
-func (l *Listener) Addr() net.Addr {
+func (l *listener) Addr() net.Addr {
 	return l.pConn.LocalAddr()
 }
 
-// Listen creates a new listener
-func Listen(network string, laddr *net.UDPAddr) (*Listener, error) {
+// ListenConfig stores options for listening to an address.
+type ListenConfig struct {
```
at-wat: This config is not accessible at now. I'd like to expose `internal/net/udp` by moving it to `pion/transport` package.
Then we'll be able to use custom inner listener using https://github.com/pion/dtls/pull/226/files#diff-804732da4aad8a84ebedfb545342f187R26.

listener.go
```diff
@@ -22,10 +22,22 @@ func Listen(network string, laddr *net.UDPAddr, config *Config) (net.Listener, e
 	}, nil
 }
 
+// NewListener creates a DTLS listener which accepts connections from an inner Listener.
+func NewListener(inner net.Listener, config *Config) (net.Listener, error) {
```
at-wat: ref: https://godoc.org/crypto/tls#NewListener

adwpc:
> Queue accept and buffer read.
> Read performance optimization is not included in this PR.
> 
> ### Reference issue
> Fixes #190
> ~Requires [pion/transport#58](https://github.com/pion/transport/pull/58)~ merged

Awsome！Concurrently read fix #190 and buffer maybe reduce cpu cost？

at-wat:
@adwpc `packetio.Buffer` is introduced to asynchronize inner connection and accepted connections. Read syscall is still called per packet. so, I guess it doesn't make large performance change.

PR #231, opened 2020-03-22
Disable insecure hash algorithms by default

daenney:
We add a new option on the configuratio which toggles the use of hash
algorithms that are deemed insecure. We default to disallowing insecure
hash algorithms, MD2 (which we don't support at all), MD5 and SHA-1.

When parsing signature algorithms we check whether it's being passed an
insecure hash algorithm, and whether we are configured to allow it or
not. If we don't allow it, we omit adding it, to ensure we never
advertise this capability.

### DIFF
--- a/config.go
+++ b/config.go
@@ -57,6 +57,10 @@ type Config struct {
 	// This should be used only for testing.
 	InsecureSkipVerify bool
 
+	// InsecureHashes allows the use of hashing algorithms that are known
+	// to be vulnerable.
+	InsecureHashes bool
+
 	// VerifyPeerCertificate, if not nil, is called after normal
 	// certificate verification by either a client or server. It
 	// receives the certificate provided by the peer and also a flag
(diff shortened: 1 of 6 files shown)


daenney:
Not sure if this is the best approach to solving this, but it's one and it seems to have the intended effect.

config.go
```diff
@@ -57,6 +57,10 @@ type Config struct {
 	// This should be used only for testing.
 	InsecureSkipVerify bool
 
+	// InsecureHashes allows the use of hashing algorithms for which
+	// insecure() call returns true.
```
at-wat: `insecure()` is unexposed and not shown in godoc.
Would be better to be explained without unexposed function.

signature_hash_algorithm.go
```diff
@@ -71,6 +71,9 @@ func parseSignatureSchemes(sigs []tls.SignatureScheme) ([]signatureHashAlgorithm
 				xerrors.Errorf("SignatureScheme %04x: %w", ss, errInvalidHashAlgorithm),
 			}
 		}
+		if h.insecure() && !insecureHashes {
+			continue
```
at-wat: Need to assert `len(out) != 0` after `for` loop, or immediately return error.

daenney:
@at-wat Addressed your comments, added an extra test too.

PR #237, opened 2020-03-27
Add ConnectionState to Conn to return State

jdbruijn:
The RemoteCertificate function is now deprecated and the
PeerCertificates property of State can be used to get the peer
certificates.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -335,11 +335,12 @@ func (c *Conn) Close() error {
 	return err
 }
 
-// RemoteCertificate exposes the remote certificate
-func (c *Conn) RemoteCertificate() [][]byte {
+// ConnectionState returns basic DTLS details about the connection.
+// Note that this replaced the `Export` function of v1.
+func (c *Conn) ConnectionState() State {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
-	return c.state.remoteCertificate
+	return *c.state.clone()
 }
 
 // SelectedSRTPProtectionProfile returns the selected SRTPProtectionProfile
--- a/conn_test.go
+++ b/conn_test.go
@@ -810,7 +810,8 @@ func TestClientCertificate(t *testing.T) {
 				if res.err != nil {
 					t.Errorf("Client failed(%v)", res.err)
 				}
-				actualClientCert := server.RemoteCertificate()
+
+				actualClientCert := server.ConnectionState().PeerCertificates
 				if tt.serverCfg.ClientAuth == RequireAnyClientCert || tt.serverCfg.ClientAuth == RequireAndVerifyClientCert {
 					if actualClientCert == nil {
 						t.Errorf("Client did not provide a certificate")
@@ -826,7 +827,7 @@ func TestClientCertificate(t *testing.T) {
 					}
 				}
 
-				actualServerCert := res.c.RemoteCertificate()
+				actualServerCert := res.c.ConnectionState().PeerCertificates
 				if actualServerCert == nil {
 					t.Errorf("Server did not provide a certificate")
 				}
--- a/examples/listen/psk/main.go
+++ b/examples/listen/psk/main.go
@@ -58,7 +58,7 @@ func main() {
 
 			// `conn` is of type `net.Conn` but may be casted to `dtls.Conn`
 			// using `dtlsConn := conn.(*dtls.Conn)` in order to to expose
-			// functions like `RemoteCertificate` etc.
+			// functions like `ConnectionState` etc.
 
 			// Register the connection with the chat hub
 			if err == nil {
--- a/examples/listen/selfsign/main.go
+++ b/examples/listen/selfsign/main.go
@@ -59,7 +59,7 @@ func main() {
 
 			// `conn` is of type `net.Conn` but may be casted to `dtls.Conn`
 			// using `dtlsConn := conn.(*dtls.Conn)` in order to to expose
-			// functions like `RemoteCertificate` etc.
+			// functions like `ConnectionState` etc.
 
 			// Register the connection with the chat hub
 			if err == nil {
--- a/examples/listen/verify/main.go
+++ b/examples/listen/verify/main.go
@@ -68,7 +68,7 @@ func main() {
 
 			// `conn` is of type `net.Conn` but may be casted to `dtls.Conn`
 			// using `dtlsConn := conn.(*dtls.Conn)` in order to to expose
-			// functions like `RemoteCertificate` etc.
+			// functions like `ConnectionState` etc.
 
 			// Register the connection with the chat hub
 			hub.Register(conn)
--- a/flight3handler.go
+++ b/flight3handler.go
@@ -64,7 +64,7 @@ func flight3Parse(ctx context.Context, c flightConn, state *State, cache *handsh
 	}
 
 	if h, ok := msgs[handshakeTypeCertificate].(*handshakeMessageCertificate); ok {
-		state.remoteCertificate = h.certificate
+		state.PeerCertificates = h.certificate
 	}
 
 	if h, ok := msgs[handshakeTypeServerKeyExchange].(*handshakeMessageServerKeyExchange); ok {
--- a/flight4handler.go
+++ b/flight4handler.go
@@ -23,11 +23,11 @@ func flight4Parse(ctx context.Context, c flightConn, state *State, cache *handsh
 	}
 
 	if h, hasCert := msgs[handshakeTypeCertificate].(*handshakeMessageCertificate); hasCert {
-		state.remoteCertificate = h.certificate
+		state.PeerCertificates = h.certificate
 	}
 
 	if h, hasCertVerify := msgs[handshakeTypeCertificateVerify].(*handshakeMessageCertificateVerify); hasCertVerify {
-		if state.remoteCertificate == nil {
+		if state.PeerCertificates == nil {
 			return 0, &alert{alertLevelFatal, alertNoCertificate}, errCertificateVerifyNoCertificate
 		}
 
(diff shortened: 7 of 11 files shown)


daenney:
I think this makes sense. It brings our API a bit closer to crypto/tls, and the naming is more correct. RemoteCertificate() suggests there's only one, but that's usually not the case.

jdbruijn:
Yeah, just a little step at a time :wink: 
I'm not sure what wyh there is a patch diff. I don't think this should have changed anything in coverage.

at-wat:
>  I'm not sure what wyh there is a patch diff. I don't think this should have changed anything in coverage.

Patch coverage is a test coverage of the edited lines in the PR.
If-statements tested only one condition is treated as partially covered (not counted as hit).

conn.go
```diff
@@ -335,11 +335,11 @@ func (c *Conn) Close() error {
...
 }
 
-// RemoteCertificate exposes the remote certificate
-func (c *Conn) RemoteCertificate() [][]byte {
+// ConnectionState returns basic DTLS details about the connection.
+func (c *Conn) ConnectionState() State {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
-	return c.state.remoteCertificate
+	return c.state
```
at-wat: I think it's better to return `clone()`-ed state to avoid any future data race.

jdbruijn:
Yeah, but my changes (mosty rename of `remoteCertificate` to `PeerCertificates` doesn't change anything in whether or not a branch of an if is triggered.

Let's take `flight4handler` line 31.
- Previous commit doesn't hit it: https://codecov.io/gh/pion/dtls/src/3caecff8298b6b091c45a1914e9bf54f3b2a7e7e/flight4handler.go#L31
- This commit also doesn't hit it: https://codecov.io/gh/pion/dtls/commit/41ecacd6288b464ad44c0679c6a3944b4806b4a5#D2-31

I understand what the patch coverage represents, but don't get why it wouldnt be a higher percentage. Maybe I'm just not seeing it.

conn.go
jdbruijn: Yeah, but that would mean that there is an error that should be checked as well and if I return `(State, error)` this is not the same as crypto/tls.

I've looked at what crypto/tls does, and that also returns a copy of the state, so I think this is good practice to return a copy!

Ideally I think we can do something similar to what crypto/tls is doing, but our state contains other information (e.g. `localRandom`) which is serialised using `.Marshal()`, which could return an error.
So I think eventually we should strip down the connection state to include less stuff so we can copy it like crypto/tls is doing, without possibility of an error. (I'm not sure whether that is possible due to differences between TLS and DTLS, but it would be very nice to have a similar state IMHO.)

So for now, do we return state and error or ignore a possible error?

```go
func (c *Conn) ConnectionState() ConnectionState {
	c.handshakeMutex.Lock()
	defer c.handshakeMutex.Unlock()

	var state ConnectionState
	state.HandshakeComplete = c.handshakeComplete()
	state.ServerName = c.serverName

	if state.HandshakeComplete {
		state.Version = c.vers
		state.NegotiatedProtocol = c.clientProtocol
		state.DidResume = c.didResume
		state.NegotiatedProtocolIsMutual = !c.clientProtocolFallback
		state.CipherSuite = c.cipherSuite
		state.PeerCertificates = c.peerCertificates
		state.VerifiedChains = c.verifiedChains
		state.SignedCertificateTimestamps = c.scts
		state.OCSPResponse = c.ocspResponse
		if !c.didResume && c.vers != VersionTLS13 {
			if c.clientFinishedIsFirst {
				state.TLSUnique = c.clientFinished[:]
			} else {
				state.TLSUnique = c.serverFinished[:]
			}
		}
		if c.config.Renegotiation != RenegotiateNever {
			state.ekm = noExportedKeyingMaterial
		} else {
			state.ekm = c.ekm
		}
	}

	return state
}
```

at-wat: It seems better to rewrite `State.clone()`.
Logically, this clone method should never fail as it is just an equivalent of memory deep copy.
I think adding `handshakeRandom.clone()` resolves it.

at-wat: I'll refactor `State.clone()` quickly.

at-wat: https://github.com/pion/dtls/pull/238

jdbruijn: Was busy yesterday with other stuff, but I can continue with this today. Thanks for refactoring!

jdbruijn:
This is now rebased on master to include #238 and I'm returning the cloned state, thanks again for refactoring that @at-wat .

As discussed I've tried removing the Export (https://github.com/pion/dtls/blob/refactor/peer-certificates/resume.go#L9), but I'm a bit overwhelmed with the test output errors I got from that.

I saw the only usage was in [resume_test](https://github.com/pion/dtls/blob/refactor/peer-certificates/resume_test.go#L114), so I replaced that with `ConnectionState` and tried to add `local.Close()` as well, but still got a some errors. I honestly don't know where to start to figure out what is going happening here, other than it has to do with `DoTestResume`.

@at-wat could you please help me figure out what is going on here?

```
$ CGO_ENABLED=0 go test
--- FAIL: TestResumeClient (10.41s)
    resume_test.go:53: EOF
    util.go:38: Unexpected routines on test end: 
        goroutine 40619 [chan send]:
        github.com/pion/dtls/v2.DoTestResume.func2(0x6fae08, 0xc00015ee40, 0xc00009a000, 0xc000056900)
                /home/jdebruijn/projects/dtls/resume_test.go:82 +0x1ee
        created by github.com/pion/dtls/v2.DoTestResume
                /home/jdebruijn/projects/dtls/resume_test.go:61 +0x393
--- FAIL: TestResumeServer (10.20s)
    util.go:38: Unexpected routines on test startup: 
        goroutine 40619 [chan send]:
        github.com/pion/dtls/v2.DoTestResume.func2(0x6fae08, 0xc00015ee40, 0xc00009a000, 0xc000056900)
                /home/jdebruijn/projects/dtls/resume_test.go:82 +0x1ee
        created by github.com/pion/dtls/v2.DoTestResume
                /home/jdebruijn/projects/dtls/resume_test.go:61 +0x393
FAIL
exit status 1
FAIL    github.com/pion/dtls/v2 32.005s
```

This is the diff I used to create above test output.
```diff
+++ b/resume.go
@@ -5,13 +5,7 @@ import (
        "net"
 )
 
-// Export extracts dtls state and inner connection from an already handshaked dtls conn
-func (c *Conn) Export() (*State, net.Conn, error) {
-       state := c.state.clone()
-       return state, c.nextConn.Conn(), nil
-}
-
-// Resume imports an already stablished dtls connection using a specific dtls state
+// Resume imports an already established dtls connection using a specific dtls state
 func Resume(state *State, conn net.Conn, config *Config) (*Conn, error) {
        if err := state.initCipherSuite(); err != nil {
                return nil, err
diff --git a/resume_test.go b/resume_test.go
index 81c5a0c..b207124 100644
--- a/resume_test.go
+++ b/resume_test.go
@@ -108,18 +108,14 @@ func DoTestResume(t *testing.T, newLocal, newRemote func(net.Conn, *Config) (*Co
                fatal(t, errChan, fmt.Errorf("messages missmatch: %s != %s", message, recv[:n]))
        }
 
-       // Export dtls connection
-       var state *State
+       // // Export dtls connection
        var innerConn net.Conn
-       state, innerConn, err = local.Export()
-       if err != nil {
-               fatal(t, errChan, err)
-       }
        if err = innerConn.Close(); err != nil {
                fatal(t, errChan, err)
        }
 
        // Serialize and deserialize state
+       state := local.ConnectionState()
        var b []byte
        b, err = state.MarshalBinary()
        if err != nil {
```

at-wat:
```diff
diff --git a/resume_test.go b/resume_test.go
index 81c5a0c..e72aad9 100644
--- a/resume_test.go
+++ b/resume_test.go
@@ -108,18 +108,12 @@ func DoTestResume(t *testing.T, newLocal, newRemote func(net.Conn, *Config) (*Co
                fatal(t, errChan, fmt.Errorf("messages missmatch: %s != %s", message, recv[:n]))
        }
 
-       // Export dtls connection
-       var state *State
-       var innerConn net.Conn
-       state, innerConn, err = local.Export()
-       if err != nil {
-               fatal(t, errChan, err)
-       }
-       if err = innerConn.Close(); err != nil {
+       if err = localConn1.Close(); err != nil {
                fatal(t, errChan, err)
        }
 
        // Serialize and deserialize state
+       state := local.ConnectionState()
        var b []byte
        b, err = state.MarshalBinary()
        if err != nil {
```
should work.

jdbruijn:
Thanks! pushed that,

What do I need to do about the codecov/patch issue?

conn.go
```diff
@@ -335,11 +335,11 @@ func (c *Conn) Close() error {
 	return err
 }
 
-// RemoteCertificate exposes the remote certificate
-func (c *Conn) RemoteCertificate() [][]byte {
+// ConnectionState returns basic DTLS details about the connection.
```
at-wat: Could you add a comment to indicate that `Export()` in v1 has been replaced by this?

at-wat:
Patch coverage state is for reference. No problem for this PR.

igolaizola:
Do you mind not merging this until monday? I have some projects depending on the current `Resume` `Export` APIs that I would like to review to give you some feedback.

jdbruijn:
Just as a reference: see #238 for discussion on removal/renaming of `Export`.

igolaizola:
> Just as a reference: see #238 for discussion on removal/renaming of `Export`.

Following the discussion...

Previous implementation of `Export` (now `ConnectionState`) returned inner `net.Conn`. This was useful for example to close the inner connection (UDP or other) and save the current DLTS state at the same time without closing the DTLS part. However, this can still be done even if `ConnectionState` doesn't return the inner connection. The caller just needs to maintain a reference to this inner `net.Conn` instead of depending on `Export` returning it.
The only inconvenient will be if the caller uses `dtls.Dial` because he won't have access to the inner connection. This can be easily fixed using `dtls.Client` instead and creating the UDP connection outside.

In short, the change looks good to me.

daenney:
Thanks for the extensive review on the API and the consequences of the changes @igolaizola, it's much appreciated. Good to know we can move forward.

conn.go
jdbruijn: Fixed in the latest version.

jdbruijn:
> Could you move ExportKeyingMaterial from *Conn to *State?
Exposing other members of State will not break API, but moving ExportKeyingMaterial breaks, so, it should be done before v2.0.0 release.

@at-wat I believe this is the last piece open in this PR. I feel like a complete seperate change so propose I create a sperate PR for that once I've merged this PR. I can work on it right away when merged and doesn't look like too much work so can get it done today!

jdbruijn:
If you're okay with that and the changes in this PR, I'll merged it and will create a PR for moving `ExportKeyingMaterial`.

PR #238, opened 2020-03-28
Refactor State.clone()

at-wat:
Remove `error` from return value of `State.clone()`.
Fix `TestResume` which didn't used deserialized state.

Note: `State.MarshalBinary()` result is incompatible with the previous version.

Related to #237

### DIFF
--- a/handshake_random.go
+++ b/handshake_random.go
@@ -15,23 +15,18 @@ type handshakeRandom struct {
 	randomBytes [randomBytesLength]byte
 }
 
-func (h *handshakeRandom) Marshal() ([]byte, error) {
-	out := make([]byte, handshakeRandomLength)
+func (h *handshakeRandom) marshalFixed() [handshakeRandomLength]byte {
+	var out [handshakeRandomLength]byte
 
 	binary.BigEndian.PutUint32(out[0:], uint32(h.gmtUnixTime.Unix()))
 	copy(out[4:], h.randomBytes[:])
 
-	return out, nil
+	return out
 }
 
-func (h *handshakeRandom) Unmarshal(data []byte) error {
-	if len(data) != handshakeRandomLength {
-		return errBufferTooSmall
-	}
+func (h *handshakeRandom) unmarshalFixed(data [handshakeRandomLength]byte) {
 	h.gmtUnixTime = time.Unix(int64(binary.BigEndian.Uint32(data[0:])), 0)
 	copy(h.randomBytes[:], data[4:])
-
-	return nil
 }
 
 // populate fills the handshakeRandom with random values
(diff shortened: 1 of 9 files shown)


resume.go
```diff
@@ -7,15 +7,15 @@ import (
 
 // Export extracts dtls state and inner connection from an already handshaked dtls conn
 func (c *Conn) Export() (*State, net.Conn, error) {
```
at-wat: Exporting inner `net.Conn` is useful for testing but should be unnecessary for the real use.
Replacing `Export` by `ConnectionState` maybe better.

handshake_random.go
```diff
@@ -15,13 +15,13 @@ type handshakeRandom struct {
...
+	var out [handshakeRandomLength]byte
 
 	binary.BigEndian.PutUint32(out[0:], uint32(h.gmtUnixTime.Unix()))
 	copy(out[4:], h.randomBytes[:])
 
-	return out, nil
+	return out
 }
 
 func (h *handshakeRandom) Unmarshal(data []byte) error {
```
jdbruijn: The `Marshal` is replaced with `marshalFixed` but `Unmarshal` is still here. Does some code still use `Unmarshal`? And if so, why doesn't it need to use the new `unmarshalFixed` func?

resume.go
jdbruijn: Yea, I agree. I propose I'll change that in #237 too

handshake_random.go
at-wat: https://github.com/pion/dtls/blob/3caecff8298b6b091c45a1914e9bf54f3b2a7e7e/handshake_message_client_hello.go#L68-L70
https://github.com/pion/dtls/blob/3caecff8298b6b091c45a1914e9bf54f3b2a7e7e/handshake_message_server_hello.go#L69-L71
these two functions unmarshal random into Slice, and Go Slice can't be converted to Array without copy.

at-wat: These functions are not called frequently, so copy may be fine.

at-wat: Removed `handshakeRandom.Unmarshal`

resume.go
igolaizola: `crypto/tls` doesn't have a `Export` method but it is a requested feature. See https://github.com/golang/go/issues/25351#issuecomment-415357066.

Both @Sean-Der and me agreed that it was an interesting feature to add to `pion/dtls` and it was added here: https://github.com/pion/dtls/pull/41

I don't mind renaming/refactoring `Export` by `ConnectionState` even if this is a breaking change but just saying that the feature of exporting and resuming is there for a reason and shouln't be taken out, at least without discussing it.

jdbruijn: I didn't know about that!
The main difference between the `Export` and `ConnectionState` is that `Export` also exposes `net.Conn`. I'm not super familiar with the DTLS resume;  is the `ConnectionState` not sufficient for resuming (i.e. is `net.Conn` also required to resume a connection)?

igolaizola: Answered here: https://github.com/pion/dtls/pull/237#issuecomment-605506634

PR #239, opened 2020-03-28
Rework #158

daenney:
* Move to stdlib imports for ed25519
* Clarify minimum Go version in README
* Minor other fixes to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,11 @@
 </p>
 <br>
 
-Go DTLS 1.2 implementation. The original user is pion-WebRTC, but we would love to see it work for everyone.
+Native [DTLS 1.2][rfc6347] implementation in the Go programming language.
 
 A long term goal is a professional security review, and maye inclusion in stdlib.
+
+[rfc6347]: https://tools.ietf.org/html/rfc6347
 
 ### Goals/Progress
 This will only be targeting DTLS 1.2, and the most modern/common cipher suites.
@@ -67,7 +69,8 @@ We would love contributes that fall under the 'Planned Features' and fixing any 
 
 ### Using
 
-This library needs at least Go 1.13.
+This library needs at least Go 1.13, and you should have [Go modules
+enabled](https://github.com/golang/go/wiki/Modules).
 
 #### Pion DTLS
 For a DTLS 1.2 Server that listens on 127.0.0.1:4444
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -67,6 +69,8 @@ We would love contributes that fall under the 'Planned Features' and fixing any
 
 ### Using
 
+This library needs at least Go 1.13.
```
at-wat: Go modules should be enabled.

daenney:
> Merging #239 into master will decrease coverage by 0.07%.

It's beyond me how only swapping out the import could result in this. But :man_shrugging:.

README.md
```diff
@@ -18,10 +18,12 @@
 </p>
 <br>
 
-Go DTLS 1.2 implementation. The original user is pion-WebRTC, but we would love to see it work for everyone.
+Native [DTLS 1.2](rfc6347) implementation in the Go programming language.
```
at-wat: Link is not correctly recognized.
See: https://github.com/pion/dtls/blob/ed25519/README.md

Maybe
```suggestion
Native [DTLS 1.2][rfc6347] implementation in the Go programming language.
```

daenney: Ah crap, thanks for catching that. I missed it b/c GitHub still renders it, but does the wrong thing :man_facepalming:

README.md
daenney: That's the case by default on Go 1.13, so that doesn't seem like something we need to explicitly call out.

daenney: https://golang.org/doc/go1.13#modules

> The GO111MODULE environment variable continues to default to auto, but the auto setting now activates the module-aware mode of the go command whenever the current working directory contains, or is below a directory containing, a go.mod file — even if the current directory is within GOPATH/src. This change simplifies the migration of existing code within GOPATH/src and the ongoing main

at-wat: Go 1.13 default is `auto`, so it depends on the directory structure.
(I recently asked by some people about build problem due to this.)

daenney: Sure, but that'll only go wrong if you're using 1.13 and you don't have a `go.mod` in the package you're building. Which on Go 1.13 means you're kinda doing it wrong. Things should have a go.mod by now, which will activate module mode by default.

at-wat: I think something like `Go modules should be enabled.` and link to https://github.com/golang/go/wiki/Modules is enough.

daenney: I think that's overdoing it, but fine. Made the change.

PR #245, opened 2020-04-02
Fix the RemoteAddr() err

bjdgyc:
#### Description
The connctx has a bug.
```go
func (c *connCtx) RemoteAddr() net.Addr {
	return c.nextConn.LocalAddr()
}
```

I fix it.
```go
func (c *connCtx) RemoteAddr() net.Addr {
	return c.nextConn.RemoteAddr()
}
```
#### Reference issue
Fixes #...

### DIFF
--- a/internal/net/connctx/connctx.go
+++ b/internal/net/connctx/connctx.go
@@ -148,7 +148,7 @@ func (c *connCtx) LocalAddr() net.Addr {
 }
 
 func (c *connCtx) RemoteAddr() net.Addr {
-	return c.nextConn.LocalAddr()
+	return c.nextConn.RemoteAddr()
 }
 
 func (c *connCtx) Conn() net.Conn {


daenney:
Eh. Ouch. Thanks for catching that. Could you please add a test to make sure this doesn't happen again? That's very not good.

// ping @at-wat for visibility

bjdgyc:
Sorry, I don't know how to test it.
The `net.Pipe` always say `pipe`. 
```go
ca,_ := net.Pipe()
c := New(ca)
addr := c.RemoteAddr()
fmt.Println(addr)
```

at-wat:
@bjdgyc good catch!

Adding a mock like:
```go
type stringAddr struct{
  network string
  addr string
}
func (a stringAddr) Network() string { return a.network }
func (a stringAddr) String() string  { return a.addr }

type connAddrMock struct {}
func (*connAddrMock) RemoteAddr() net.Addr { return stringAddr{"remote_net", "remote_addr"} }
func (*connAddrMock) LocalAddr() net.Addr { return stringAddr{"local_net", "local_addr"} }
func (*connAddrMock) Read(b []byte) (n int, err error) { panic("unimplemented") }
func (*connAddrMock) Write(b []byte) (n int, err error) { panic("unimplemented") }
func (*connAddrMock) Close() error { panic("unimplemented") }
func (*connAddrMock) SetDeadline(t time.Time) error { panic("unimplemented") }
func (*connAddrMock) SetReadDeadline(t time.Time) error { panic("unimplemented") }
func (*connAddrMock) SetWriteDeadline(t time.Time) error { panic("unimplemented") }
```
should work.

bjdgyc:
@at-wat 
Hi, I think this don't cover the `connCtx struct`.

at-wat:
Test code would be like:
```go
c := New(&connAddrMock{})
al := c.LocalAddr()
ar := c.RemoteAddr()
// Check al.Network(), al.String(), ar.Network(), ar.String(), 
```

bjdgyc:
@at-wat 
Thank you, I have pull the testing.
pr: https://github.com/pion/dtls/pull/246

at-wat:
merged as https://github.com/pion/dtls/commit/513a0e3eef80175e02831380c136f75cb1e84d49

PR #248, opened 2020-04-29
Don't retransmit HelloVerifyRequest

at-wat:
HelloVerifyRequests must not be retransmitted.
[RFC 6347 Section-3.2.1]

### DIFF
--- a/flight0handler.go
+++ b/flight0handler.go
@@ -6,14 +6,12 @@ import (
 )
 
 func flight0Parse(ctx context.Context, c flightConn, state *State, cache *handshakeCache, cfg *handshakeConfig) (flightVal, *alert, error) {
-	// HelloVerifyRequest can be skipped by the server,
-	// so allow ServerHello during flight1 also
-	seq, msgs, ok := cache.fullPullMap(state.handshakeRecvSequence,
+	seq, msgs, ok := cache.fullPullMap(0,
 		handshakeCachePullRule{handshakeTypeClientHello, cfg.initialEpoch, true, false},
 	)
 	if !ok {
 		// No valid message received. Keep reading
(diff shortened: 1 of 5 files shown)


at-wat:
Client will retransmit the first ClientHello and the server must handle it.

flight0handler.go
```diff
@@ -6,14 +6,12 @@ import (
 )
 
 func flight0Parse(ctx context.Context, c flightConn, state *State, cache *handshakeCache, cfg *handshakeConfig) (flightVal, *alert, error) {
-	// HelloVerifyRequest can be skipped by the server,
-	// so allow ServerHello during flight1 also
-	seq, msgs, ok := cache.fullPullMap(state.handshakeRecvSequence,
+	seq, msgs, ok := cache.fullPullMap(0,
```
at-wat, line 9: cache must have a flag indicates the message was already processed.
otherwise, any non-ClientHello packet may trigger HelloVerifyRequest.

at-wat: Only handshake packet can triggering it.
One response per one packet is OK in point of DDoS prevention.

Sean-Der:
@at-wat LGTM! Great work :)

Do you think it would be possible to write a test for this? Maybe we just cut a TODO, but I could see this regressing easily in the future (and we wouldn't notice)

at-wat:
@Sean-Der added a test to verify HelloVerifyRequest retransmit condition.

at-wat:
fixed lint error and test failure on WASM

Sean-Der:
@at-wat really love that test you wrote!

I need to think about it, but I bet we have lots of tests we could add now that we have that :)

PR #258, opened 2020-05-28
Make handshake errors from Accept() non-fatal

### LINKED ISSUE #255: Closing connection on the server causes fatal error from Accept()
### Your environment.
  - Version: v2.0.0
  - Browser: n/a

### What did you do?

- Start a server with an Accept() loop
- Have a client connect to the server and just Read() on the connection
- The server actively closes the connection

Server:
```go
package main

import (
	"context"
	"fmt"
	"net"
	"time"

	"github.com/pion/dtls/v2"
	"github.com/pion/dtls/v2/examples/util"
)

func main() {
	addr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 4444}

	config := &dtls.Config{
		PSK: func(hint []byte) ([]byte, error) {
			return []byte{0xAB, 0xC1, 0x23}, nil
		},
		PSKIdentityHint:      []byte("Pion DTLS Client"),
		CipherSuites:         []dtls.CipherSuiteID{dtls.TLS_PSK_WITH_AES_128_CCM_8},
		ExtendedMasterSecret: dtls.RequireExtendedMasterSecret,
		ConnectContextMaker: func() (context.Context, func()) {
			return context.WithTimeout(context.Background(), 2*time.Second)
		},
	}

	listener, err := dtls.Listen("udp", addr, config)
	util.Check(err)

	for {
		fmt.Println("waiting for connection")
		conn, err := listener.Accept()
		if err != nil {
			if netErr, ok := err.(net.Error); ok && netErr.Temporary() {
				fmt.Println("tempoary error, continuing")
				continue
			}
			fmt.Println("permanent error")
			panic(err)
		}

		go func() {
			fmt.Println("closing connection in 1s")
			time.Sleep(time.Second)
			conn.Close()
		}()
	}
}
```

Client:
```go
package main

import (
	"fmt"
	"io"
	"net"
	"os"

	"github.com/pion/dtls/v2"
	"github.com/pion/dtls/v2/examples/util"
)

func main() {
	addr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 4444}

	config := &dtls.Config{
		PSK: func(hint []byte) ([]byte, error) {
			return []byte{0xAB, 0xC1, 0x23}, nil
		},
		PSKIdentityHint:      []byte("Pion DTLS Server"),
		CipherSuites:         []dtls.CipherSuiteID{dtls.TLS_PSK_WITH_AES_128_CCM_8},
		ExtendedMasterSecret: dtls.RequireExtendedMasterSecret,
	}

	conn, err := dtls.Dial("udp", addr, config)
	util.Check(err)

	fmt.Println("Connected")
	io.Copy(os.Stdout, conn)
}
```

### What did you expect?

Expected the connection to be closed and the server to keep accepting future (new) connections.

### What happened?

The connection is closed, but by doing do, the server's Accept() receives a confirmation from the client which triggers a new handshake. Since this isn't a real handshake, it times out and becomes a Fatal error from listener.Accept() which usually fails the server.

If you let me know how best to address this, I can try to make a PR for it. Perhaps the response from the client shouldn't trigger another handshake, and it's probably best to not make handshake timeouts fatal errors anyway.

folbricht:
#### Description

Makes handshake errors non-fatal in Accept() to avoid servers stopping when one client fails to complete the handshake and open a connection.

First attempt, not exactly an elegant solution.

#### Reference issue
Fixes #255

### DIFF
--- a/cipher_suite.go
+++ b/cipher_suite.go
@@ -170,7 +170,7 @@ func parseCipherSuites(userSelectedSuites []CipherSuiteID, excludePSK, excludeNo
 		for _, id := range ids {
 			c := cipherSuiteForID(id)
 			if c == nil {
-				return nil, fmt.Errorf("CipherSuite with id(%d) is not valid", id)
+				return nil, &invalidCipherSuite{id}
 			}
 			cipherSuites = append(cipherSuites, c)
 		}
--- a/errors.go
+++ b/errors.go
@@ -36,7 +36,6 @@ var (
 	errCompressionMethodUnset           = &FatalError{errors.New("server hello can not be created without a compression method")}
 	errCookieMismatch                   = &FatalError{errors.New("client+server cookie does not match")}
 	errCookieTooLong                    = &FatalError{errors.New("cookie must not be longer then 255 bytes")}
-	errHandshakeTimeout                 = &FatalError{xerrors.Errorf("the connection timed out during the handshake: %w", context.DeadlineExceeded)}
 	errIdentityNoPSK                    = &FatalError{errors.New("PSK Identity Hint provided but PSK is nil")}
 	errInvalidCertificate               = &FatalError{errors.New("no certificate provided")}
 	errInvalidCipherSpec                = &FatalError{errors.New("cipher spec invalid")}
@@ -100,6 +99,27 @@ type TimeoutError struct {
 	Err error
 }
 
+// HandshakeError indicates that the handshake failed.
+type HandshakeError struct {
+	Err error
+}
+
+// invalidCipherSuite indicates an attempt at using an unsupported cipher suite.
+type invalidCipherSuite struct {
+	id CipherSuiteID
+}
+
+func (e *invalidCipherSuite) Error() string {
+	return fmt.Sprintf("CipherSuite with id(%d) is not valid", e.id)
+}
+
+func (e *invalidCipherSuite) Is(err error) bool {
+	if other, ok := err.(*invalidCipherSuite); ok {
+		return e.id == other.id
+	}
+	return false
+}
+
 // Timeout implements net.Error.Timeout()
 func (*FatalError) Timeout() bool { return false }
 
@@ -144,6 +164,27 @@ func (e *TimeoutError) Unwrap() error { return e.Err }
 
 func (e *TimeoutError) Error() string { return fmt.Sprintf("dtls timeout: %v", e.Err) }
 
+// Timeout implements net.Error.Timeout()
+func (e *HandshakeError) Timeout() bool {
+	if netErr, ok := e.Err.(net.Error); ok {
+		return netErr.Timeout()
+	}
+	return false
+}
+
+// Temporary implements net.Error.Temporary()
+func (e *HandshakeError) Temporary() bool {
+	if netErr, ok := e.Err.(net.Error); ok {
+		return netErr.Temporary()
+	}
+	return false
+}
+
+// Unwrap implements Go1.13 error unwrapper.
+func (e *HandshakeError) Unwrap() error { return e.Err }
+
+func (e *HandshakeError) Error() string { return fmt.Sprintf("handshake error: %v", e.Err) }
+
 // errAlert wraps DTLS alert notification as an error
 type errAlert struct {
 	*alert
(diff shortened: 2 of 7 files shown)


folbricht:
Apparently I broke a bunch of tests that expect fatal errors from Server(). Will fix those up if you want to go with this solution.

at-wat:
I think this solution is reasonable.
The tests should be updated to compare errors by `errors.Is()`.

folbricht:
Fixed up the impacted tests to use errors.Is()

cipher_suite.go
```diff
@@ -170,7 +170,7 @@ func parseCipherSuites(userSelectedSuites []CipherSuiteID, excludePSK, excludeNo
 		for _, id := range ids {
 			c := cipherSuiteForID(id)
 			if c == nil {
-				return nil, fmt.Errorf("CipherSuite with id(%d) is not valid", id)
+				return nil, &invalidCipherSuite{0}
```
at-wat, line 173:
```suggestion
				return nil, &invalidCipherSuite{id}
```

errors.go
```diff
@@ -36,7 +36,7 @@ var (
 	errCompressionMethodUnset           = &FatalError{errors.New("server hello can not be created without a compression method")}
 	errCookieMismatch                   = &FatalError{errors.New("client+server cookie does not match")}
 	errCookieTooLong                    = &FatalError{errors.New("cookie must not be longer then 255 bytes")}
-	errHandshakeTimeout                 = &FatalError{xerrors.Errorf("the connection timed out during the handshake: %w", context.DeadlineExceeded)}
+	errHandshakeTimeout                 = &FatalError{&HandshakeError{xerrors.Errorf("the connection timed out during the handshake: %w", context.DeadlineExceeded)}}
```
at-wat, line 39: I don't have strong opinion, but wrapping at https://github.com/pion/dtls/pull/258/files#diff-a2ddca1600f34abb91244424f8431c2cL889 as same as the other errors would have better uniformity.

folbricht: It'd be nice, but whichever way I go there, it'll look inconsistent. The code-path there used to return a FatalError and I didn't want to change that while getting a HandshakeError inside it. Could move that piece of code but it'll look equally out of place. Any suggestions of how to arrange it better without changing the semantics?

errors.go
```diff
@@ -100,6 +100,27 @@ type TimeoutError struct {
 	Err error
 }
 
+// HandshakeError indicates that the handshake failed.
+type HandshakeError struct {
```
at-wat, line 104: HandshakeError should provide `Temporary()` and `Timeout()` which transparently returns a value from underlying error (`Err.Temporary()` and `Err.Timeout()`) or `false` if the error doesn't implement it.

folbricht: done

errors.go
at-wat: Other handshake errors except `errHandshakeTimeout` are now `&HandshakeError{&FatalError{error}}` https://github.com/pion/dtls/pull/258/files#diff-a2ddca1600f34abb91244424f8431c2cR887-R892.
Only `errHandshakeTimeout` is `&FatalError{&HandshakeError{error}}`.
I think it's better to unify them.

folbricht: Ideally I'd like to change it to something like this:
```go
func (c *Conn) translateHandshakeCtxError(err error) error {
	if err == nil {
		return nil
	}
	if errors.Is(err, context.Canceled) && c.isHandshakeCompletedSuccessfully() {
		return nil
	}
	return &HandshakeError{err}
}
```
This looks correct to me since it returns a HandshakeError when a handshake fails which can be temporary or permanent (depending on what error wrapped). However that'll mean Server(), Dial(), Client() do no longer always return a FatalError like currently. If that works for you, I can make the necessary changes. Alternatively, I can start wrapping every error that comes out of Server(), Dial(), etc in a FatalError in their respective functions to keep the current behavior (which doesn't seem quite right).
To me, FatalError means "this operation failed and there's no hope this could ever work" which is correct for things like bad config, certs and the like. A timeout in the handshake doesn't really fall into this category, even for a Dial() operation.

at-wat: IMO, `FatalError` is not explicitly telling `operation is failed`, but just both `.Temporary()` and `.Timeout()` (that implements `net.Error`) return false.
If returning `&HandshakeError{context.DeadlineExceeded}` without using `errHandshakeTimeout`, the error doesn't implement `net.Error` and user can't know it is a timeout error by unified way.

BTW, `errHandshakeTimeout` should be `TimeoutError` error instead of `FatalError`.

folbricht: Thanks. The HandshakeError implements net.Error now and passes Temporary() and Timeout() to the inner error.

folbricht:
Thank you. Rebased and conflict resolved

at-wat:
Thank your! Merging

PR #261, opened 2020-06-01
Fragmentbuffer: Fix fragment parsing

jeffreystoke:
#### Description

This commit is an attempt to support multiple handshake messages in single dtls record fragment.

DTLS connection failure happened when I was trying to connect to emqx coap dtls port, according to the wireshark capture result, the emqx server was sending multiple handshake messages in single dtls record, which resulted handshake unmarshal error in this library, and as noted in the last paragraph of https://tools.ietf.org/html/rfc6347#section-4.2.3

> multiple handshake messages may be placed in the same DTLS record,
> provided that there is room and that they are part of the same flight.

So I believe we should support such feature (also bug fix?)

#### Reference issue

none

/cc @at-wat , @Sean-Der 

*I have read https://github.com/pion/webrtc/wiki/Contributing#writing-a-good-pr but didn't try to find a reviewer, hope that's ok :)


jeffreystoke:
seems the `.github/lint-disallowed-functions-in-library.sh` is flaky, I tried to run it on my local machine, the result was always correct, but CI says no

daenney:
The tests aren't failing right now on disallowed-functions though, but:

```
Please add the following contributors to the README
Jeff Ctor

The command "bash .github/assert-contributors.sh" exited with 1.
```

It's confused about your name.

jeffreystoke:
Hi @daenney , sorry, I pasted the wrong script name, it's actually `.github/assert-contributors.sh`, but the I'm not sure if it's something related to git, but if I run that script locally, the output is correct everytime, and here is the git output

```
# git log --format='%aN' | sort -u | grep Jeff
Jeffrey Stoke
```

daenney:
Mm, peculiar. If I look you up in Travis CI, you show up as Jeffrey Cstor. Which might be why: https://travis-ci.org/github/jeffreystoke

jeffreystoke:
but if you look at this [build](https://travis-ci.org/github/pion/dtls/jobs/693381203), it actually found the correct name, I will try to push it again.

daenney:
Hmmm indeed. Going to summon @Sean-Der here.

Sean-Der:
Fantastic work @jeffreystoke, merging!

Sorry about the busted CI @daenney :(

PR #267, opened 2020-06-20
Break out internal/net/udp

backkem:
Breaking out the UDP wrapper to make it re-
usable in pion/sctp.

Relates to pion/sctp#74

TODO:
- [x] Tag pion/udp after landing CI


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/pion/dtls/v2
 require (
 	github.com/pion/logging v0.2.2
 	github.com/pion/transport v0.10.0
+	github.com/pion/udp v0.0.0-20200620185302-adf622c7c621
```
Sean-Der, line 6: Mind tagging `v0.0.1`

Sean-Der: Never mind you already made a note :)

Sean-Der:
Are you able to delete `github.com/pion/dtls/v2/internal/net/udp` now?

backkem:
> I definitely agree to split udp package and make it exposed. ([pion/transport#57](https://github.com/pion/transport/issues/57))
> Do you think it's better to be a independent package out of pion/transport?

I think we've had overall positive experience with putting 'protocols' in their own repo. While pion/udp is only a thin wrapper it may still be useful on its own. In such case I prefer a separate repo since I want to encourage others to mix and match our protocols into their own stack. The DTLS implementation has been a prime example of this.
pion/transport is becoming a bit of a 'utils' package. That's fine for tooling we use across our libraries. However, if some piece, like maybe the mux, are useful to let people throw together their own stack, I'd consider exposing them using their own repo.

PR #273, opened 2020-07-15
Add AEAD_AES_128_GCM SRTP Profiles

Sean-Der:
Relates to pion/srtp#85


Sean-Der:
I couldn't find any docs on client (or server) preference for SRTP protection profile. I just added a test to make sure the current behavior never changes unexpectedly.

PR #276, opened 2020-07-23
Support multiple HelloVerifyRequest messages with different cookies

### LINKED ISSUE #275: Add support for multiple HelloVerifyRequest messages with different cookies
## Summary

[RFC 6347 Section 4.2.1](https://tools.ietf.org/html/rfc6347#section-4.2.1) (just at the end of that section) says:

> Note to implementors: This may result in clients receiving multiple HelloVerifyRequest messages with different cookies. Clients SHOULD handle this by sending a new ClientHello with a cookie in response to the new HelloVerifyRequest.

Since word `SHOULD` is used this is not mandatory but it might be good to follow the recommendation.

## Motivation

I was dealing with a custom DTLS server implementation that sometimes was sending extra HelloVerifyRequests with an updated cookie. As `pion/dtls` currently doesn't process additional HelloVerifyRequests the handshake was failing.

Changing the custom DTLS server implementation to avoid sending those extra HelloVerifyRequests.

igolaizola:
During flight3 if we receive a HelloVerifyRequest the cookie will
be updated and a new ClientHello will be sent.

Resolves #275

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -1687,3 +1687,90 @@ func TestProtocolVersionValidation(t *testing.T) {
 		}
 	})
 }
+
+func TestMultipleHelloVerifyRequest(t *testing.T) {
+	// Limit runtime in case of deadlocks
+	lim := test.TimeOut(time.Second * 20)
+	defer lim.Stop()
+
+	// Check for leaking routines
+	report := test.CheckRoutines(t)
+	defer report()
+
+	cookies := [][]byte{
+		// first clientHello contains an empty cookie
+		{},
+	}
+	var packets [][]byte
+	for i := 0; i < 2; i++ {
+		cookie := make([]byte, 20)
+		if _, err := rand.Read(cookie); err != nil {
+			t.Fatal(err)
+		}
+		cookies = append(cookies, cookie)
+
+		record := &recordLayer{
+			recordLayerHeader: recordLayerHeader{
+				sequenceNumber:  uint64(i),
+				protocolVersion: protocolVersion1_2,
+			},
+			content: &handshake{
+				handshakeHeader: handshakeHeader{
+					messageSequence: uint16(i),
+				},
+				handshakeMessage: &handshakeMessageHelloVerifyRequest{
+					version: protocolVersion1_2,
+					cookie:  cookie,
+				}},
+		}
+		packet, err := record.Marshal()
+		if err != nil {
+			t.Fatal(err)
+		}
+		packets = append(packets, packet)
+	}
+
+	ca, cb := dpipe.Pipe()
+	defer func() {
+		err := ca.Close()
+		if err != nil {
+			t.Error(err)
+		}
+	}()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+	defer cancel()
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	defer wg.Wait()
+	go func() {
+		defer wg.Done()
+		_, _ = testClient(ctx, ca, &Config{}, false)
+	}()
+
+	for i, cookie := range cookies {
+		// read client hello
+		resp := make([]byte, 1024)
+		n, err := cb.Read(resp)
+		if err != nil {
+			t.Fatal(err)
+		}
+		record := &recordLayer{}
+		if err := record.Unmarshal(resp[:n]); err != nil {
+			t.Fatal(err)
+		}
+		clientHello := record.content.(*handshake).handshakeMessage.(*handshakeMessageClientHello)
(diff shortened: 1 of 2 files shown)


conn_test.go
```diff
@@ -1687,3 +1687,87 @@ func TestProtocolVersionValidation(t *testing.T) {
...
+func TestMultipleHelloVerifyRequest(t *testing.T) {
+	// Limit runtime in case of deadlocks
+	lim := test.TimeOut(time.Second * 20)
+	defer lim.Stop()
+
+	// Check for leaking routines
+	report := test.CheckRoutines(t)
+	defer report()
+
+	cookies := make([][]byte, 1)
```
at-wat, line 1700: Better to add a comment that `cookies[0]` is `nil` for the flight1 ClientHello.

flight3handler.go
```diff
@@ -5,9 +5,24 @@ import (
 )
 
 func flight3Parse(ctx context.Context, c flightConn, state *State, cache *handshakeCache, cfg *handshakeConfig) (flightVal, *alert, error) {
-	var msgs map[handshakeType]handshakeMessage
-	var ok bool
-	var seq int
+	// Clients may receive multiple HelloVerifyRequest messages with different cookies.
+	// Clients SHOULD handle this by sending a new ClientHello with a cookie in response
+	// to the new HelloVerifyRequest. RFC 6347 Section 4.2.1
+	seq, msgs, ok := cache.fullPullMap(state.handshakeRecvSequence,
```
at-wat, line 11: What handshake message sequence number does the server use in the new `HelloVerifyRequest`?
The code currently assume that new `HelloVerifyRequest` has a sequence number continued from the previous `HelloVerifyRequest`.
In this case, `HelloVerifyRequest` will be duplicated in the `handshakeCache`.
I haven't fully checked yet, but I think duplicated `HelloVerifyRequest` in `handshakeCache` will cause `cache.pullAndMerge` result duplicated and make data signatures generated in flight4 and 5 wrong.

igolaizola: > What handshake message sequence number does the server use in the new HelloVerifyRequest?

The first `HelloVerifyRequest` goes with `MessageSequence: 0` and the second one with `MessageSequence: 1`

conn_test.go
igolaizola: updated :+1:

flight3handler.go
at-wat: So, the handshake will be like
```
         Client                             Server
         ------                             ------
         ClientHello (seq=0)  ------>

                              <------ HelloVerifyRequest (seq=0)

         ClientHello (seq=1)  ------>
         (with cookie)

                              <------ HelloVerifyRequest (seq=1) (with new cookie)

         ClientHello (seq=2)  ------>
         (with new cookie)

                              <------        ServerHello (seq=2)
                              <------        Certificate (seq=3)
                              <------    ServerHelloDone (seq=4)

         [Rest of handshake]
```

> I haven't fully checked yet, but I think duplicated HelloVerifyRequest in handshakeCache will cause cache.pullAndMerge result duplicated and make data signatures generated in flight4 and 5 wrong.

This was wrong. `handshakeCache.pull()` only outputs the last message if same type of the messages are given multiple times. It satisfies https://tools.ietf.org/html/rfc6347#section-4.2.1
> If HelloVerifyRequest is used, the initial ClientHello and HelloVerifyRequest are not included in the calculation of the handshake_messages (for the CertificateVerify message) and verify_data (for the Finished message).

and `ClientHello (seq=0)`, `HelloVerifyRequest (seq=0)` and `ClientHello (seq=1)` will not be included in the MAC calculation. This behavior makes sense to me.

igolaizola: Thanks for the analysis!

igolaizola:
> LGTM
> 
> I'm curious which DTLS server emits two HelloVerifyRequest in what case. It would be great to add such case to E2E test if possible.

I am still analyzing this scenario in the server part. The server is using `mbedTLS`.

igolaizola:
> > LGTM
> > I'm curious which DTLS server emits two HelloVerifyRequest in what case. It would be great to add such case to E2E test if possible.
> 
> I am still analyzing this scenario in the server part. The server is using `mbedTLS`.

So this is what is happening:
 - Server doesn't receive the `ClientHello` within a given timeout
 - Server sends a new `ClientHello` using a new cookie (it overrides the previous one)
 - Client doesn't send `ClientHello` with an updated cookie and the DTLS handshake never completes

I think this PR makes `pion/dtls` closer to the RFC. However, I should also adjust the timeout values on the server to avoid this scenario.

@Sean-Der any comments before merging this?

PR #280, opened 2020-08-25
Update CI configs to v0.3.4

pionbot:


Sean-Der:
@at-wat Do you know the solution to `handshake error: x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0` off the top of your head? I will dig in later if you don't know.

thanks!

at-wat:
I don't know but I guess we shouldn't use this:
https://github.com/pion/dtls/blob/c6dbd482fa3a0a7655e9dc338a0f15be905cdf35/pkg/crypto/selfsign/selfsign.go#L81

daenney:
In the Go 1.15 release notes: https://golang.org/doc/go1.15#commonname

> X.509 CommonName deprecation
>
> The deprecated, legacy behavior of treating the CommonName field on X.509 certificates as a host name when no Subject Alternative Names are present is now disabled by default. It can be temporarily re-enabled by adding the value x509ignoreCN=0 to the GODEBUG environment variable. 

Basically, we need to use the SAN instead of the CN.

daenney:
I've pushed a small commit that should fix this particular case, but realistically we need to do a bit more work in the selfsign package to tidy it up. Many of our helpers there take a "CN" as argument. We have two options:
* Drop the CN entirely from all our helpers/code
* Concatenate CN and SANs and set the result as `DNSNames`

daenney (approved):

Sean-Der:
Nice! Thanks for grabbing that @daenney upgrading other packages to 1.15 took it out of me :)

Maybe we don’t even need this being public. If WebRTC/testing is only user.

daenney:
Yeah, I considered that. They're very useful helpers though for testing, if anyone's doing E2E style testing when consuming our library so I've left them in place for now.

PR #293, opened 2020-09-17
Update CI configs to v0.4.7

pionbot:


scorpionknifes:
I'm taking this one.

cipher_suite.go
```diff
@@ -12,25 +12,23 @@ type CipherSuiteID uint16
 // Supported Cipher Suites
 const (
 	// AES-128-CCM
-	TLS_ECDHE_ECDSA_WITH_AES_128_CCM   CipherSuiteID = 0xc0ac
-	TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8 CipherSuiteID = 0xc0ae
+	TLS_ECDHE_ECDSA_WITH_AES_128_CCM   CipherSuiteID = 0xc0ac //nolint:golint,stylecheck
+	TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8 CipherSuiteID = 0xc0ae //nolint:golint,stylecheck
```
scorpionknifes, line 16: Assuming these consts are styled this way.

errors.go
```diff
@@ -13,68 +13,68 @@ import (
...
+	errHandshakeInProgress          = &TemporaryError{errors.New("handshake is in progress")}                                   //nolint:goerr113
+	errInvalidContentType           = &TemporaryError{errors.New("invalid content type")}                                       //nolint:goerr113
+	errInvalidMAC                   = &TemporaryError{errors.New("invalid mac")}                                                //nolint:goerr113
+	errInvalidPacketLength          = &TemporaryError{errors.New("packet length and declared length do not match")}             //nolint:goerr113
+	errReservedExportKeyingMaterial = &TemporaryError{errors.New("ExportKeyingMaterial can not be used with a reserved label")} //nolint:goerr113
+
+	errCertificateVerifyNoCertificate   = &FatalError{errors.New("client sent certificate verify but we have no certificate to verify")}                      //nolint:goerr113
+	errCipherSuiteNoIntersection        = &FatalError{errors.New("client+server do not support any shared cipher suites")}                                    //nolint:goerr113
```
scorpionknifes, line 30: Should we wrap each error with a static error based on TemporaryError, FatalError, etc?

e.g
```go
var errTemporaryError = errors.New("Temporary Error")

var errBufferTooSmall = &TemporaryError{fmt.Errorf("%w: buffer is too small", errTemporaryError)} 
```

pkg/crypto/ccm/ccm_test.go
```diff
@@ -18,8 +19,10 @@ func mustHexDecode(s string) []byte {
 	return r
 }
 
-var aesKey1to12 = mustHexDecode("c0c1c2c3c4c5c6c7c8c9cacbcccdcecf")
-var aesKey13to24 = mustHexDecode("d7828d13b2b0bdc325a76236df93cc6b")
+var (
+	aesKey1to12  = mustHexDecode("c0c1c2c3c4c5c6c7c8c9cacbcccdcecf") //nolint:gochecknoglobals
+	aesKey13to24 = mustHexDecode("d7828d13b2b0bdc325a76236df93cc6b") //nolint:gochecknoglobals
+)
```
scorpionknifes, line 25: If I convert this to a func would it recalculate every call? 
Keep as global to reuse variable?

scorpionknifes (commented):
Couple of unsure lints

errors.go
at-wat: It should be wrapped to implement `net.Error`.
`Temporary()` is required to handle errors in DoS-vulnerability-free way. (see https://github.com/pion/dtls/releases/tag/v2.0.0 `Fix unreachable error handling`)

at-wat: @Sean-Der I think `err113` lint should not be enabled.

https://github.com/Djarvur/go-err113#details
> error should not be created dynamically from scratch but by the wrapping the static (package-level) error.

I believe this is basically wrong from the Go's language design.
Using `fmt.Errorf("%w")` is just a simplified way of implementing Go 1.13 error interface. Having package specific error type and embed context of the error as a machine readable values must be allowed by default.

cipher_suite.go
Sean-Der: :+1: yea this is the right decision!

errors.go
Sean-Der: @at-wat What do you think of something like?

```
      // TemporaryError indicates that the DTLS connection is still available, but the request was failed temporary.
      TemporaryError    = errors.New("A error that is temporary. Please retry")                
      errBufferTooSmall = fmt.Errorf("%w: buffer is too small", TemporaryError)         
```
The users can do `errors.Is(err, dtls.TemporaryError)`

With `goerr113` I found a lot of things we did wrong. I don't think we will see immediate gain, but long term I think it will really help consistency wise.

at-wat: > The users can do `errors.Is(err, dtls.TemporaryError)`

@Sean-Der It's not `net.Error` compatible. User should check both `err.(net.Error).Temporary()` (from underlying `net.Conn`) and `errors.Is(err, dtls.TemporaryError)`. If there is another layers of `net.Conn` user should check all error type of each package, otherwise it injects DoS vulnerability.

pkg/crypto/ccm/ccm_test.go
Sean-Der: I think that is the right choice! I don't think it is so bad for tests.

For main package code unexpected things can happen (like `append` will modify) so want to avoid those sharp edges.

errors.go
Sean-Der: Nice catch, sorry missed that nuance.

So personally my vote is keeping. `goerr113`. It is very valuable for packages that get less attention `srtp`, `datachannel` and `turn` being the big ones. I think for `dtls` a lot more effort has been put in, and it is fine to ignore them on a case by case basis.

Personally I am in favor of merging this PR. I respect your decision if you disagree, we can find a better solution!

at-wat: This PR looks good to me with these nolint comments. Linter selection thing is better to be discussed at https://github.com/pion/.goassets.

Sean-Der (approved):

PR #300, opened 2020-10-22
Allow user to set own cipher suites to DTLS

jkralik:
### Description
TLS_ECDH_ANON_WITH_AES_128_CBC_SHA256 is used to own device via JustWorks ownership transfer method defined in specification https://openconnectivity.org/specs/OCF_Security_Specification.pdf 7.3.4

Allow user to set own cipher to config
Allow to use anon ciphers
Add TLS_ECDH_ANON_WITH_AES_128_CBC_SHA256


Sean-Der:
:fire: this is awesome @jkralik I am all for merging this.

I think we should move the record layer/protocol stuff into `pkg`. I also had a user ask on Slack if we could export that recently. I think if we leave it in the root it bloats the public API a bit. I think users by default just expect a simple DTLS Client/Server. They might be overwhelmed. 

I am happy to do this work if everyone agrees!

jkralik:
> 
> 
> 🔥 this is awesome @jkralik I am all for merging this.
> 
> I think we should move the record layer/protocol stuff into `pkg`. I also had a user ask on Slack if we could export that recently. I think if we leave it in the root it bloats the public API a bit. I think users by default just expect a simple DTLS Client/Server. They might be overwhelmed.
> 
> I am happy to do this work if everyone agrees!

Yes. Nice idea @Sean-Der !

cipher_suite.go
```diff
@@ -84,62 +90,80 @@ func CipherSuiteName(id CipherSuiteID) string {
...
+	return []CipherSuite{
+		NewCipherSuiteTLSEcdheEcdsaWithAes128Ccm(),
+		NewCipherSuiteTLSEcdheEcdsaWithAes128Ccm8(),
+		&CipherSuiteTLSEcdheEcdsaWithAes128GcmSha256{},
+		&CipherSuiteTLSEcdheRsaWithAes128GcmSha256{},
+		&CipherSuiteTLSEcdheEcdsaWithAes256CbcSha{},
+		&CipherSuiteTLSEcdheRsaWithAes256CbcSha{},
+		NewCipherSuiteTLSPskWithAes128Ccm(),
+		NewCipherSuiteTLSPskWithAes128Ccm8(),
+		&CipherSuiteTLSPskWithAes128GcmSha256{},
```
igolaizola, line 141: Why do we need all cipher suites to be public? This means we are adding all these structs and methods to the public api of pion/dtls.

jkralik: As @Sean-Der mentioned we want to move them to the separate package. And i want to allow customize cipher like: https://github.com/plgd-dev/sdk/blob/justWorks/local/core/otm/just-works/client.go#L45 because OCF has own cipher number for ANON.

igolaizola: > As @Sean-Der mentioned we want to move them to the separate package

:+1: Ok to that. Since it will be a public API it should be cleaned up a little bit. For example creation of CipherSuite `&CipherSuiteXXX{}` vs `NewCipherSuiteXXX`.

>And i want to allow customize cipher like: https://github.com/plgd-dev/sdk/blob/justWorks/local/core/otm/just-works/client.go#L45 because OCF has own cipher number for ANON.

Another option will be to use `CipherSuitesFactory` to append a new custom `CipherSuite` that replicates the code from `NewCipherSuiteTLSEcdhAnonWithAes128CbcSha256` but with a different cipher number. This approach means having code redundacy for a specific use case, but the pion/dtls API will be more simple.

jkralik:
> I am happy to do this work if everyone agrees!

@Sean-Der How do you see to move cipherSuites to pkg - is there any progress? If not I can try to do it tomorrow - could describe pkgs structure ? Or just create `pkg/cipherSuites` pkg and move them all ciphers?

Sean-Der:
@jkralik sorry on being slow! Had a daughter and hard to find any spare time :)

I was going to do `pkg/protocol` and maybe `pkg/ciphersuite` and then maybe have an interface for `CipherSuite` in `pion/dtls`?

I haven't put much thought into it honestly. My only goal was for `github.com/pion/dtls` to feel like `crypto/tls`. We can have much more freedom in `pkg/*` though.

ondrejtomcik:
@Sean-Der any update? I would like to use this from the master 🥇

Sean-Der:
Hey @jkralik 

Sorry this took so long. I just released pion/webrtc@v3 and getting this merged is my only priority now! There is a lot going on here. I am going to break it down into simpler commits (I need to learn them myself) and will send out PRs. I am going to do the `CipherSuiteFactory` and then `Anonymous Ciphers`

I am also going to export much less. In the future we can export more so users can define their own cipher, but doesn't actually seem to be a need (yet). All you need today is to be able to configure a cipher.

Sean-Der:
Hey @jkralik I am gonna start cherry-picking your work in another PR. You will still be the author on the commits!

thanks

PR #307, opened 2020-11-08
Support both PKI and PSK in one server

### LINKED ISSUE #297: Support both PKI and PSK (at the same time)
## Summary

I'd like to use both PKI and PSK as a DTLS server, for different clients. So one client makes a connection and uses PKI and another client makes a connection and uses PSK. As these are different 'instances', so to speak, I don't see why this wouldn't be possible. I've seen in the config validation, `validateConfig` that it will give an error on `len(config.Certificates) > 0 && config.PSK != nil`, so I know that that would be blocking supporting both for now. What I don't know and am curious to hear about, is whether the underlying code would work with supporting both or whether that change would be a difficult one.

https://github.com/pion/dtls/blob/master/config.go#L157-L158
```go
func validateConfig(config *Config) error {
	switch {
	case config == nil:
		return errNoConfigProvided
	case len(config.Certificates) > 0 && config.PSK != nil:
		return errPSKAndCertificate
	case config.PSKIdentityHint != nil && config.PSK == nil:
		return errIdentityNoPSK
	}
```

## Motivation

Why are we doing this? What use cases does it support? What is the expected outcome? Is this available in other implementations? Can this not be implemented using primitives?
Reason for this is to support different types of clients from one server. We currently have one server that can serve N clients concurrently but that now only handles PKI. We want to have devices connecting to the server using both PKI and PSK.

A clear and concise description of the alternative solutions you've considered.
One alternative would be to have a load balancer and multiple servers, some for PKI and some for PSK. The load balancer would then of course need to figure out what the client wants to do, PKI or PSK, and pass to the appropriate server. This alternative would be quite complex for us to implement and be a less 'clean' solution IMO.

CarsonHoffman:
#### Description
This PR allows a server to interact with both PKI-based and PSK-based clients. Previously, the server had to choose one (and only one) method of interaction with clients, and made its internal choices on which functionality to use based on that choice. Now, servers can support either, and chooses its functionality based on the agreed-upon cipher suite. This change removes `errPSKAndCertificate`, as it is tautologically no longer needed. 🙂

#### Reference issue
Closes #297.


CarsonHoffman:
Force-pushing to add bodies to the commit messages. Oops!

cipher_suite.go
```diff
@@ -189,18 +193,26 @@ func parseCipherSuites(userSelectedSuites []CipherSuiteID, excludePSK, excludeNo
 		cipherSuites = defaultCipherSuites()
 	}
 
+	var foundPSK, foundNonPSK bool
```
daenney, line 196: For symmetry I'd prefer `foundPKI` instead of `foundNonPSK`

CarsonHoffman: Do you think it'd be best to universally switch the "nonPSK" nomenclature to `PKI` (e.g. `includeNonPSK`, `errNoAvailableNonPSKCipherSuite`)? I just subconsciously used the original wording from the `excludeNonPSK` parameter name, but I think that changing them might be reasonable too; my only concern would be that the RFCs tend to refer to "PSK cipher suites" but don't qualify other cipher suites as "PKI cipher suites"; I'm a fairly new reader of them and could just be overthinking it, though. 🙂

daenney: I'd prefer to not have `NonX`, it ends up being awkward as it includes a negation itself. `NonX` is the same as `X` set to the other logical value, i.e flip true/false instead of inverting it through the name itself.

daenney: @Sean-Der Any thoughts on the naming?

Sean-Der: Yea I think that would be nice!

I wouldn't have thought of this before, but https://refactoring.com/catalog/removeDoubleNegative.html and https://derickbailey.com/2014/03/27/double-negatives-in-code-dont-not-avoid-double-negative-boolean-logic/ have some good points on it also.

CarsonHoffman: Thank you for the suggestion! I went ahead and switched to "certificate" throughout the PR, and while I was changing `parseCipherSuites` I switched around the order in which the certificate and PSK cipher suites are checked to match up with the order of the parameters; it was a subtle thing I missed but I think it reads more nicely now!

config_test.go
```diff
@@ -16,21 +16,45 @@ func TestValidateConfig(t *testing.T) {
...
 	}
 	config := &Config{
+		CipherSuites: []CipherSuiteID{TLS_PSK_WITH_AES_128_CCM_8, TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256},
+		PSK: func(hint []byte) ([]byte, error) {
+			return nil, nil
+		},
+		Certificates: []tls.Certificate{cert},
+	}
+	if err = validateConfig(config); err != nil {
+		t.Fatalf("TestValidateConfig: Client error exp(%v) failed(%v)", nil, err)
```
daenney, line 33: This seems a bit odd. The first `%v` will always be nil?

conn_test.go
```diff
@@ -1209,6 +1209,76 @@ func TestCipherSuiteConfiguration(t *testing.T) {
...
+				}()
+			} else {
+				t.Errorf("TestCertificateAndPSKServer: Server Error Mismatch '%s': expected(%v) actual(%v)", test.Name, nil, err)
+			}
+
+			res := <-c
+			if res.err == nil {
+				_ = server.Close()
+			} else {
+				t.Errorf("TestCertificateAndPSKServer: Client Error Mismatch '%s': expected(%v) actual(%v)", test.Name, nil, res.err)
```
daenney, line 1276: Same here, and the above 3. Interpolating nil in the format string doesn't seem right.

config_test.go
daenney: The convention in most Go project, as well as this code base is to use `expected: something, got: something else`. Make it easy for humans to read, not for computers to parse.

daenney (approved):
I've left two minor comments, but I'm happy with this regardless.

config_test.go
CarsonHoffman: Would hard-coding `nil` in the string be preferable? I had copied the format from other tests, and it feels to me that keeping both parameters as interpolated into the string would be consistent with reading other tests (i.e. you can always look at the last two parameters for expected and got when reading the tests source), but I'm okay with whatever you feel would be best for the project overall.

daenney: Yeah, lets just keep it for now as it is. I was reading through the change and somewhat surprised, until I opened the whole file and realised we do this everywhere for these tests. I'll go in at some point and fix that when the urge strikes.

daenney:
@Sean-Der any last comments or shall we go ahead with this?

CarsonHoffman:
I went ahead and rebased since the README commit was conflicting with my concurrent README change in another PR that was merged. Let me know if there's anything else I can help with on this!

daenney (approved):

daenney:
Nope sorry, I meant to merge this earlier but I completely forgot about it. My apologies.

PR #308, opened 2020-11-08
Always select matching CipherSuite

CarsonHoffman:
#### Description
Currently, the server checks to see if *any* of the client's cipher suites match *any* of its own, but always selects the first that the client offered. This can lead to a situation where the server selects a cipher suite it does not support; for example, if the client offers suites `[A, B]` and the server supports only `[B]`, the server will see the match of `B`, but will then select `A`, which it does not support.

This PR makes the server always select a matching cipher suite, and adds a test to confirm this functionality; `HEAD` currently fails the test.


daenney (approved):
Thanks for fixing this :facepalm:

Sean-Der:
Yea amazing first PR :D

Sean-Der:
Merged with ca3d9821c7adf49e429c42b51436c640f0a9d7a3 just squashed into one commit.

Amazing work @CarsonHoffman and thanks for the quick review @daenney :)

PR #309, opened 2020-11-09
DTLS fixes for TinyDTLS client

bocajim:
#### Description
I need to support clients that use a very old DTLS implementation from TinyDTLS
https://projects.eclipse.org/projects/iot.tinydtls

There are two problems:

1. The client does not advertise any extensions, and currently if there are no extensions, the handshake is dropped.  This was an easy fix to allow for 0 extensions.

2. The second problem is that the client establishes two connections using the same source IP:Port, so it will perform a full handshake and exchange some app data, then it will begin a new handshake on the same IP:Port combination.  This currently breaks the library because once a session is started it can't be re-created.  The fix for this is more dirty, it is specifically checking for ClientHello packets, and if detected, it closes the session, which will clean out the DTLS server's session and allow another ClientHello to be sent successfully.


daenney:
The fix for no extensions seems fin to me, but I really don't like the second one. I'm not sure I want to support that. Is there some reference in TLSv1.2 RFCs about this behaviour?

bocajim:
For the second one, I have no problem wrapping the bypass for ClientHello in a config parameter, but for IoT use cases in general, I have seen this behavior from many devices.  Most of the devices bind to a fixed port meaning that from client to server all clients are bound to 5684, connecting to the server on 5684 so even acts like rebooting the client device cause problems because they boot and send a new ClientHello and it gets dropped because of the replay window.

I'm open to other ways to address this issue, the main problem is that a client tries to connect (by sending a ClientHello) on an IP:Port that the server already has a session for.

The "fix" itself to close the connection is definitely not right, the appropriate handling would be to start a new handshake, but it appears that this is difficult at the point in the callstack when the code determines that it needs to re-handshake.

PR #313, opened 2020-11-18
New features for LPWAN devices

bocajim:
#### Description
1. Added support for clients that do not publish any extensions
2. Added optional configuration to bypass the replay detection window on new ClientHello packets.  (this fixes a problem with clients that use fixed IP and ports).
3. Added support for TLS_PSK_AES_128_CBC_SHA256 cipher suite.


Sean-Der:
@bocajim Merged, sorry for the delay :(  TLS_PSK_WITH_AES_128_CBC_SHA256 and the empty extension list is fixed.

I didn't take the `ClientHello` change. I am going to export all the protocol messages (in `github.com/pion/dtls/pkg/protocol`) and then you can handle it externally. I am afraid of introducing a vulnerability accidentally. It might not exist in this PR, but a future contribution could cause it.

thanks, and hope we can get you on `pion/dtls` :)

PR #331, opened 2021-01-01
Update golang.org/x/net

daenney:
Closes #330


daenney:
@at-wat @Sean-Der Would one of you mind? I had to fix the `go.sum`, seems to break every now and then.

Sean-Der:
@daenney I know it is annoying :/ 

I debugged a bit and it looks like the renovatebot is running an older version of Go, and breaks because we use APIs not available. Maybe are `go.mod` is lying and we need to bump?

Sean-Der (approved):

PR #338, opened 2021-01-19
Move Pseudorandom Functions to /pkg

Sean-Der:
With this users will be able to easily define ciphers of their own


daenney (approved):
I haven't reviewed this very thoroughly, but on the surface it looks OK to me and the tests are still happy.

PR #340, opened 2021-01-28
Add support for anonymous ciphers

### LINKED ISSUE #337: Implement TLS_ECDH_ANON_WITH_AES_128_CBC_SHA256
I am going to take the work @jkralik has already done into master. 

The existing code is on this branch https://github.com/pion/dtls/tree/cipherSuiteFactory. The in a follow up PR I am going to add the `CipherSuiteFactory` to the config that allows people to pass anything that satisfies `dtls.CipherSuite`

Sean-Der:
To use it a user must provide their own CipherSuite. Anonymous Ciphers
are vulnerable to MITM attacks so we aren't providing them in public
API.

Resolves #337


Sean-Der:
https://gist.github.com/Sean-Der/1e4192b37c8a9ef5cf18d6272d22e503

@daenney @jkralik don't bother reviewing this until #337 has been merged. This is that and the anonymous specific changes. I want a review just for this since it is security sensitive. Lots of changes in the state machine etc...

jkralik:
I just tested the branch against iotivity-lite device (mbedtls) and it works 👍

Sean-Der:
@jkralik hopefully this covers your needs/exposes everything you need

flight3handler.go
```diff
@@ -95,6 +95,8 @@ func flight3Parse(ctx context.Context, c flightConn, state *State, cache *handsh
 
 	if h, ok := msgs[handshake.TypeCertificate].(*handshake.MessageCertificate); ok {
 		state.PeerCertificates = h.Certificate
+	} else if state.cipherSuite.AuthenticationType() == CipherSuiteAuthenticationTypeCertificate {
```
Sean-Der, line 98: We need this assert now because we aren't guarded by `fullPullMap`

jkralik (approved):
Great work 🥳

Sean-Der:
@daenney mind giving it a quick look over as well, no rush :)

jkralik:
> @jkralik 99% of the credit goes to you! I was really just cherry-picking your code. Thanks for starting the PR, and sorry it took so long :( I put you down as the co-author

Thx. I'm so glad that the feature is ready to be merge to the master branch. 🚀

pkg/crypto/hash/hash.go
```diff
@@ -15,7 +15,7 @@ type Algorithm uint16
 
 // Supported hash algorithms
 const (
-	MD2     Algorithm = 0 // Blacklisted
+	None    Algorithm = 0 // Blacklisted
```
daenney, line 18: Should we not keep MD2?

pkg/crypto/hash/hash_test.go
```diff
@@ -8,12 +8,12 @@ import (
 
 func TestHashAlgorithm_StringRoundtrip(t *testing.T) {
 	for algo := range Algorithms() {
-		if algo == Ed25519 {
-			// Ed25519 does hashing and signing in one time.
-			continue
-		}
 		str := algo.String()
 		hash1 := algo.CryptoHash()
+		if algo == Ed25519 || algo == None {
```
daenney, line 13: This feels like it should move back up, above the `str` and `hash1` assignments?

Sean-Der: Fixed!

pkg/crypto/hash/hash.go
Sean-Der: I don't think this was correct before [IANA Hash Assignments](https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-18)

This showed up in bbee407bcfbf6d6a574842e09b50a5f4b67c071e and we have just carried it ever since.

Sean-Der:
@daenney fixed! Mind giving it another look over?

daenney (approved):

PR #344, opened 2021-02-04
Add support for KeyLogWriter

### LINKED ISSUE #343: Add support for KeyLogWriter
## Summary

TLS configs in Go have a `KeyLogWriter` which writes secrets for easy decryption later (e.g for use with Wireshark). This is similar to `curl`'s `SSLKEYLOGFILE=` env var. Please can you add the ability to do this in `dtls.Config`.

## Motivation

It's useful to be able to decrypt DTLS streams when debugging/testing. This feature is available in `tls.Config` - https://golang.org/pkg/crypto/tls/#Config
```go
    // KeyLogWriter optionally specifies a destination for TLS master secrets
    // in NSS key log format that can be used to allow external programs
    // such as Wireshark to decrypt TLS connections.
    // See https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format.
    // Use of KeyLogWriter compromises security and should only be
    // used for debugging.
    KeyLogWriter io.Writer // Go 1.8
```

Kegsay:
Fixes #343 - style is done in the same way as `tls.Config`.

Manually tested both as a client and as a server, then reading the resulting logs in Wireshark and confirming that data can be decrypted.

### DIFF
--- a/config.go
+++ b/config.go
@@ -6,10 +6,13 @@ import (
 	"crypto/ed25519"
 	"crypto/tls"
 	"crypto/x509"
+	"io"
 	"time"
 
 	"github.com/pion/logging"
 )
+
+const keyLogLabelTLS12 = "CLIENT_RANDOM"
 
 // Config is used to configure a DTLS client or server.
 // After a Config is passed to a DTLS function it must not be modified.
@@ -112,6 +115,14 @@ type Config struct {
 	// Packet with sequence number older than this value compared to the latest
 	// accepted packet will be discarded. (default is 64)
 	ReplayProtectionWindow int
+
+	// KeyLogWriter optionally specifies a destination for TLS master secrets
+	// in NSS key log format that can be used to allow external programs
+	// such as Wireshark to decrypt TLS connections.
+	// See https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format.
+	// Use of KeyLogWriter compromises security and should only be
+	// used for debugging.
+	KeyLogWriter io.Writer
 }
 
 func defaultConnectContextMaker() (context.Context, func()) {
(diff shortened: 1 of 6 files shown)


Sean-Der:
This looks fantastic! Would be happy to take right away. Would you mind adding a test? Then should be good to go :)

I can fix all the other little things (linters and so on)

daenney:
I'd like to see some actual tests for this. Since the KeyLogWriter is an `io.Writer`, you can put in a `bytes.Buffer` during test and check what ends up in there.

Kegsay:
It's hard to see at what level you wish to add tests for:
 - Verify something is written to the `io.Writer` which looks roughly right according to https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format - effectively a unit test for `writeKeyLog`.
 -  Verify that something is written at the right stage of the handshake process. - tests at the `handshakeFSM` level.
 - Verify that the data written to `io.Writer` can actually decrypt traffic. - E2E test.

The latter being the most useful but also the most effort to write tests for.

Sean-Der:
I am happy with 1/2! 

I think we can just remove randomness and then just assert a certain value is generated everytime. Just enough to make sure someone doesn't regress your work in the future :)

Kegsay:
Test was added a few weeks ago.

daenney:
The tests are fine, but as you can see from the PR status, our commit linter disagrees: https://github.com/pion/dtls/pull/344/checks?check_run_id=1838025192#step:3:15. It's a small thing, but we like to keep this consistent on this repo and everything else in the Pion org.

daenney (approved):

Sean-Der:
Thank you so much for contributing this @Kegsay! I squashed your commits into 4879d3479bc9bb4afdb17011d2f19708e0fa2362

PR #314, opened 2020-11-23
Implement Renegotiation Indication Extension

fffilimonov:
Implement Renegotiation Indication Extension

Add minimal renegotiation_info support

#### Description
As described in #274 it's not possible to connect to server which using Bouncy Castle DTLS library (e.g. Jitsi) as a client using Pion.
According to [rfc5746](https://tools.ietf.org/html/rfc5746) the client MUST include either an empty "renegotiation_info" extension, or the TLS_EMPTY_RENEGOTIATION_INFO_SCSV signaling cipher suite value in the ClientHello. Pion doesn't support renegotiations so it doesn't require anything except sending empty renegotiation_info. But it will increase interoperability.

#### Reference issue
Fixes #274

PR #346, opened 2021-02-16
Include RenegotiationInfo in ServerHello

Sean-Der:
Before if clients sent a ClientHello with RenegotiationInfo we would
just ignore. Now we properly respond that we don't support.

Relates to #314

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -1902,3 +1902,125 @@ func TestMultipleHelloVerifyRequest(t *testing.T) {
 	}
 	cancel()
 }
+
+// Assert that a DTLS Server only responds with RenegotiationInfo if
+// a ClientHello contained that extension
+func TestRenegotationInfo(t *testing.T) {
+	// Limit runtime in case of deadlocks
+	lim := test.TimeOut(10 * time.Second)
+	defer lim.Stop()
+
+	// Check for leaking routines
+	report := test.CheckRoutines(t)
+	defer report()
+
+	resp := make([]byte, 1024)
+
+	for _, testCase := range []struct {
+		Name                    string
+		ExpectRenegotiationInfo bool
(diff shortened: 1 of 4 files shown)


Sean-Der:
@sipsorcery mind trying this PR? You can use `go mod edit` and do a replace. If you don't have the time don't worry about it! I think this will merge super quick.

@fffilimonov @daenney mind giving a quick review?

sipsorcery:
Happy to test.

I'm having some challenges with setting the go module to the PR. Here are the steps I've taken:

 - Cloned this repo and checked out a branch with this PR,
 - In my [go.mod](https://github.com/sipsorcery/webrtc-echoes/blob/master/pion/client/go.mod) file of the application I'm testing I've added:
`replace github.com/pion/dtls => c:/dev/github/dtls`

Those steps don't seem to be getting the fix and I suspect there are some more steps needed to switch the module to my `dtls` repo.

daenney:
I don't recall exactly, but I don't think that's going to work because you don't explicitly list dtls as a dependency for your module. It's a transitive dependency through webrtc, so you probably need to clone webrtc, use the dtls replace directive there and on this module use a replace for webrtc.

sipsorcery:
Tried the below for my `go.mod` but doesn't seem to have used the dtls source from my repo.

````

Sean-Der:
I say we just get this merged instead of fighting that. This is a good chang no matter what!

Sean-Der:
@daenney awesome, mind approving when you get a chance? If you don't have bandwidth we can nag some others :)

daenney (approved):

daenney:
Done. For some reason my brain thought you had, but that can’t be since you raised the PR 😅

PR #350, opened 2021-03-12
Change findMatchingCipherSuite

PieerePi:
change findMatchingCipherSuite to pick local prefered cipher suite from client's cipher suites when pion/dtls acting as a dtls server.

The new function findMatchingCipherSuite picks the local preferred TLSEcdheEcdsaWithAes128GcmSha256 cipher suite in defaultCipherSuites to match the ECDSA 256 certificate set by pion/webrtc.

It's just a roundabout way to avoid the inconsistency between the CipherSuiteID in ServerHello message and the Certificate in Certificate message when pion/dtls is used by pion/webrtc.

https://github.com/pion/dtls/pull/350#issuecomment-798207935

previous commit log
----------
2nd:
Fix findMatchingCipherSuite, it confused with a and b. Rename them to remote and local, and change the logic.

1st:
Prefers ECDSA 256 cipher suite in DTLS handshake "Client Hello" process to improve compatibility

Pion returns TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA "Cipher Suite" in "Server Hello" message, but provides TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA type certificate in "Certificate" message. 

// Prefers TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA like Chrome does, also because
// pion/webrtc uses it by default without any configuration interface.
// https://github.com/pion/webrtc/blob/master/dtlstransport.go#L75
func findMatchingCipherSuite(a, b []CipherSuite) (CipherSuite, bool) { //nolint

There is an old similiar issue https://github.com/pion/dtls/issues/133.


daenney:
pion/dtls is used by quite a few projects outside of the WebRTC ecosystem, for which this wouldn't be appropriate at all. I don't think we should be special casing certain behaviour in here to match a particular user-agent.

Our library lets you specify a list of cipher suites through [`config.CipherSuites`](https://pkg.go.dev/github.com/pion/dtls#Config) for both client and server. Why not use that, and put the ciphers in the priority you want?

PieerePi:
@daenney Thank you for inspiring me. I dived into the code, and found that TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 is already in the first place (see defaultCipherSuites). The problem is that function findMatchingCipherSuite confused with a and b. I renamed them to remote and local, and changed the logic. It passed the inter-op test with FreeSWITCH.
@Sean-Der Please check if this change matches your idea of the function?
Thanks!

Sean-Der:
Hey @PieerePi, so excited to get you involved. Not a lot of devs get this deep :) 

I see what you are talking about. So the issue is CipherSuite priority/who picks? Right now it looks like the `ClientHello` CipherSuites takes precedence. If we have the CipherSuite locally then we get a match. 

I am totally up for switching this behavior! 
* Can you find either in the spec or another implementation (OpenSSL is probably best) on what the behavior is.
* Write some tests to assert this doesn't change in the future.

`local` and `remote` are a little confusing. Maybe `hackstack`/`needles` to match the stdlib?

daenney:
The behaviour is usually configurable on the server side. Apache/nginx etc used to have a 'prefer_server_ciphers` at which point it would iterate through the server ciphers and pick the most preferred match from there. This was mostly because there are tons of somewhat wonky ciphers supported in TLSv1-v1.2, but has disappeared since TLSv1.3 since none of the cipher suites are considered problematic (yet).

I think it might make sense to have a option in our `Config` for this, which would then define whether we start from ClientHello or ServerHello. And then I guess pion/webrtc would need to let you configure that behaivour.

PieerePi:
Hi @Sean-Der and @daenney, thanks for your reply!

First, let me describe the issue I encountered.
1. FreeSWITCH sends a ClientHello message with a list of cipher suites, of which the top two pion/dtls supported are TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA and TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA.
2. Pion/dtls picks TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA in ServerHello message (it's done by findMatchingCipherSuite, the client's cipher suites takes precedence), but provides a ecdsa-with-SHA256 type certificate in Certificate message (pion/webrtc creates an ECDSA 256 type certificate by default. https://github.com/pion/webrtc/blob/master/dtlstransport.go#L75).
3. FreeSWITCH rejects it with a DTLS alert, the underlying error is ERR_LIB_SSL/SSL_F_SSL3_GET_SERVER_CERTIFICATE/SSL_R_WRONG_CERTIFICATE_TYPE.

In pion/dtls, I can't find any connection between the CipherSuiteID in ServerHello message and Certificate in Certificate message. There is no constraints to limit how to pick cipher suite even local has a certificate.

Thanks to Sean's informative information, I studied the implementation of crypto/tls and found solution there.

pickCipherSuite in crypto/tls/handshake_server.go, https://github.com/golang/go/blob/master/src/crypto/tls/handshake_server.go#L295.
1. Function selectCipherSuite uses an ok function to filter out cipher suites that do not match the local certificate.
This solves the bug described above.
2. It implements a PreferServerCipherSuites flag to let server pick its most preferred cipher suite.

But this solution might change too much code of pion/dtls, I am not familiar with pion/dtls at present (also I am a newbie to go^_^), I think I am not competent to fix this issue completely.

daenney:
Sending a ServerHello with the RSA suite but then following up with a ServerCertificate with an ECDSA cert sounds like a bug on our side. That should never happen, regardless of client or server cipher suite priorities.

PieerePi:
@daenney Yes.

The implementation of crypto/tls adds more attributes for cipher suite than just an ID, gets more information from local certificate, and uses all these to check if client's cipher suite and local certificate match.

PieerePi:
> Sending a ServerHello with the RSA suite but then following up with a ServerCertificate with an ECDSA cert sounds like a bug on our side. That should never happen, regardless of client or server cipher suite priorities. We definitely need to fix that, if we don’t have an RSA cert we need to not pick that cipher for the ServerHello

The new function findMatchingCipherSuite in this PR picks the local preferred TLSEcdheEcdsaWithAes128GcmSha256 cipher suite to match the ECDSA 256 certificate set by pion/webrtc.

It's just a roundabout way to avoid the problem.

Sean-Der:
Upstream deprecated `PreferServerCipherSuites` so good thing we didn't add them 

```
	// It used to control whether the server would follow the client's or the
	// server's preference. Servers now select the best mutually supported
	// cipher suite based on logic that takes into account inferred client
	// hardware, server hardware, and security.
	//
	// Deprecated: PreferServerCipherSuites is ignored.
```

I am going to add a test right now that confirms ECDSA/RSA causes the proper cipher suites to be used.

Sean-Der:
We already have a test also! b8f72f3774497508fd9bc960a676dc395c572416

PR #357, opened 2021-03-26
Add missing *rsa.PrivateKey into validating config certs private keys

misak113:
#### Description

validates private keys of certs to be one of RSA, ECDSA or ED25519
everywhere else the RSA is correctly supported but the config validation

#### Reference issue
No issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Jim Wert](https://github.com/bocajim)
 * [Alvaro Viebrantz](https://github.com/alvarowolfx)
 * [Kegan Dougal](https://github.com/Kegsay)
+* [Michael Zabka](https://github.com/misak113)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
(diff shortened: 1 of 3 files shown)


misak113:
Do you have idea how to fix lint error
```
  Error: SA1019: package crypto/dsa is deprecated: DSA is a legacy algorithm, and modern alternatives such as Ed25519 (implemented by package crypto/ed25519) should be used instead. Keys with 1024-bit moduli (L1024N160 parameters) are cryptographically weak, while bigger keys are not widely supported. Note that FIPS 186-5 no longer approves DSA for signature generation.  (staticcheck)
```

DSA is used only in tests on purpose to validate wrong input certificate private key.

Sean-Der:
Hey @misak113 

thank you so much for the PR! You can add a `//nolint` to the line in question

misak113:
Ok, PR is hopefully ready for review.

sampleref:
Yes, I too could make it work only after adding _case *rsa.PrivateKey:_ as allowed case in _validateConfig_ at https://github.com/pion/dtls/blob/7afa8961e75d6df59c39cdd4b5a1cae50597d33e/config.go#L184

daenney (approved):
I'm surprised we didn't see this earlier

Sean-Der:
Merged! Also tagged `v2.0.9` thank you so much @misak113 @sampleref and @daenney :)

PR #395, opened 2021-09-11
config: adding support for GetCertificate(*tls.ClientHello)

### LINKED ISSUE #386: Support for GetCertificate and GetClientCertificate similar as in crypto/tls
## Summary

Go's `crypto/tls` has support for `GetCertificate` where it can return a certificate used for the handshake based on ClientHello. Though this feature is commonly used to issue ACME certificate on the fly, or to support certificate reloading without restarting the server.

## Motivation

It would be tremendously useful to support `GetCertificate` in dtls so the server doesn't have to have the certificates issued already on start-up, and closes the features parity between `crypto/tls` and `dtls` where it can almost be a drop in replacement, at least for the use cases outline above.

I'm not aware of any workaround for this feature, though I admit that I haven't dug into the code too deeply.

## Additional context

`nil`

zllovesuki:
#### Description

Add support for `GetCertificate func(*tls.ClientHelloInfo) (*tls.Certificate, error)`. The goal is to close the feature parity gap with stdlib's `tls` package.

Questions to reviewers: What other tests should I be adding? I'm certain that I've missed some code paths.

#### Reference issue
Fixes #386

### DIFF
--- a/certificate.go
+++ b/certificate.go
@@ -10,10 +10,24 @@ func (c *handshakeConfig) getCertificate(serverName string) (*tls.Certificate, e
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	certs := c.localCertificates
+	if len(c.localCertificates) == 0 && c.getCertificateFunc != nil {
+		cert, err := c.getCertificateFunc(&tls.ClientHelloInfo{
+			ServerName: serverName,
+		})
+		if err != nil {
+			return nil, err
+		}
+		if cert != nil {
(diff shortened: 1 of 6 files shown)


jkralik:
When I implemented GetCertificate and created PR, I saw yours. Could I add you as a contributor to my PR and close this one?

Sean-Der:
oh jeez sorry I missed this @zllovesuki 

That sounds like the right thing to do to me @jkralik! Thanks for adding as contributor to your PR :)

PR #397, opened 2021-09-29
Pin e2e Dockerfile

Sean-Der:
Using debian base instead

### DIFF
--- a/e2e/Dockerfile
+++ b/e2e/Dockerfile
@@ -1,9 +1,4 @@
-FROM golang:1.14-alpine3.11
-
-RUN apk add --no-cache \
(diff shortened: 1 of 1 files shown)


daenney (approved):
Uhg why is software so terrible.

at-wat (commented):
It's better to leave an issue to remind us to revert it or migrate to newer version of OpenSSL later.

Sean-Der:
Found the root issue here https://github.com/openssl/openssl/issues/16707

Sean-Der:
@at-wat done! I found the issue upstream and opened a PR.

PR #399, opened 2021-10-20
Choose cipher suite matching the certificate

### LINKED ISSUE #398: Server chooses cipher suite not matching its certificate
If the server only has either ECDSA or RSA certificates, it needs to choose the corresponding cipher suite, even if the client announces a preference for a cipher suite of the other type.

How to reproduce:
1. Run the server (master branch): `go run examples/listen/selfsign/main.go`. It will generate a self-signed ECDSA certificate.
2. Connect with OpenSSL s_client, without client certificates, with preference for ECDSA (works):
```openssl s_client -dtls1_2 -connect 127.0.0.1:4444 -debug -cipher aECDSA:aRSA```
3. Do the same, but with preference for RSA (fails):
```openssl s_client -dtls1_2 -connect 127.0.0.1:4444 -debug -cipher aRSA:aECDSA```

The fail makes s_client send a fatal alert and exit immediately printing:
```
140576763131264:error:1416F17F:SSL routines:tls_process_server_certificate:wrong certificate type:ssl/statem/statem_clnt.c:1958:
```
The go server then exits too because of the fatal alert:
```
net.Error: handshake error: alert: Alert Fatal: IllegalParameter
panic: handshake error: alert: Alert Fatal: IllegalParameter
```

This issue was mentioned in https://github.com/pion/dtls/pull/350#issuecomment-798403320.

lukaslihotzki:
#### Reference issue
Fixes #398

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -2,6 +2,9 @@ package dtls
 
 import (
 	"context"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/rsa"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


at-wat:
Approve and run button is not appeared. Reopening PR

at-wat:
Seems causing panic in some test cases

Sean-Der:
cherry-picked to https://github.com/pion/dtls/pull/417

Will fix and add you as reviewer @at-wat

PR #407, opened 2021-12-07
Do not include IP addresses as SNI values

### LINKED ISSUE #406: SNI field contains an IP address instead of a valid hostname
### Your environment.
  - Version: v2.0.8
  - Browser: varies

### What did you do?

We've been investigating a possible attempt to block Snowflake by fingerprinting the DTLS connections between clients and Snowflake proxies in Russia: https://ntc.party/t/ooni-reports-of-tor-blocking-in-certain-isps-since-2021-12-01/1477/18

One of the identified differences between the DTLS handshake messages created by Snowflake's use of pion/dtls and a WebRTC connection between two browsers is that pion/dtls will include the SNI extension in the client hello and it is populated with the IP address of the remote peer.

See https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/snowflake/-/issues/40014 for an overview of differences and examples.

This seems to be a bug. [Looking at the behaviour of SNI in crypto/tls, ](https://cs.opensource.google/go/go/+/master:src/crypto/tls/handshake_client.go;l=993;drc=master?q=ServerName&ss=go%2Fgo&start=21) there is the following note:
```
// hostnameInSNI converts name into an appropriate hostname for SNI.
// Literal IP addresses and absolute FQDNs are not permitted as SNI values.
// See RFC 6066, Section 3.
```

### Proposed change

I think ideally, the SNI extension would not be sent if there is no hostname. I saw in https://github.com/pion/dtls/issues/345 that this library wants to follow the behaviour of crypto/tls and it was asked whether they will send blank SNI extensions. Looking more at crypto/tls, it does appear that the server name extension is only included if the name is not empty:
https://cs.opensource.google/go/go/+/master:src/crypto/tls/handshake_messages.go;l=124

cohosh:
According to RFC 6066, IP address literals are not permitted as valid
SNI values in the server name extension.
See https://datatracker.ietf.org/doc/html/rfc6066#section-3

#### Description

I used the same method that crypto/tls uses to check to see if the `serverName` is an IP address: https://cs.opensource.google/go/go/+/master:src/crypto/tls/handshake_client.go;l=1004

I kept the port parsing logic for the case that `net.ParseIP` would also fail if the `remoteAddr` was an IP:Port pair.

I didn't see existing testing logic to check SNI values but can add tests if necessary.

#### Reference issue
Fixes #406

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -10,6 +10,7 @@ backkem <mail@backkem.me>
 bjdgyc <bjdgyc@163.com>
 Bragadeesh <bragboy@gmail.com>
 Carson Hoffman <c@rsonhoffman.com>
+Cecylia Bocovich <cohosh@torproject.org>
 Chris Hiszpanski <thinkski@users.noreply.github.com>
 Daniele Sluijters <daenney@users.noreply.github.com>
 folbrich <frank.olbricht@gmail.com>
--- a/conn.go
+++ b/conn.go
@@ -147,16 +147,10 @@ func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient
 	c.setLocalEpoch(0)
 
 	serverName := config.ServerName
-	// Use host from conn address when serverName is not provided
-	if isClient && serverName == "" && nextConn.RemoteAddr() != nil {
-		remoteAddr := nextConn.RemoteAddr().String()
-		var host string
-		host, _, err = net.SplitHostPort(remoteAddr)
-		if err != nil {
-			serverName = remoteAddr
-		} else {
-			serverName = host
-		}
+	// Do not allow the use of an IP address literal as an SNI value.
+	// See RFC 6066, Section 3.
+	if net.ParseIP(serverName) != nil {
+		serverName = ""
 	}
 
 	hsCfg := &handshakeConfig{
--- a/conn_test.go
+++ b/conn_test.go
@@ -2024,3 +2024,83 @@ func TestRenegotationInfo(t *testing.T) {
 		})
 	}
 }
+
+func TestServerNameIndicationExtension(t *testing.T) {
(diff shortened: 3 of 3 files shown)


daenney:
Thanks for you contribution! Before we can merge this we need a test for this to ensure we don't regress.

uumaro:
Go crypto/tls handles this somewhat differently. Its [`hostnameInSNI`](https://github.com/golang/go/blob/4300f105147dc0da9d1034704ad1cd24bedde5da/src/crypto/tls/handshake_client.go#L993) function maps IP addresses to `""`, but it [keeps an empty string as an empty string](https://github.com/golang/go/blob/4300f105147dc0da9d1034704ad1cd24bedde5da/src/crypto/tls/handshake_client_test.go#L1416) and never peeks at a `RemoteAddr` to set a default.

From my point of view, a good-enough solution would be to delete the whole [`serverName` inference block](https://github.com/pion/dtls/blob/415eb6f8a18b35b8fbae137c83046692e159f1e6/conn.go#L149-L160) and pass `config.ServerName` to `&handshakeConfig{}` directly. I think an empty `ServerName` should always mean to omit the SNI extension. Beyond that, it's additionally possible to filter out IP addresses, even if manually set by the caller, as in `hostnameInSNI`.

cohosh:
@daenney thanks, just pushed a test.

I'm happy to implement @uumaro 's suggestion as well.

cohosh:
Just removed the population of the serverName with the remote addr of the conn as suggested by @uumaro 

I think this is ready for review again.

Sean-Der (approved):

Sean-Der:
LGTM! Thank you so much @cohosh 

I tagged `v2.0.11` also

cohosh:
Thank you!!

PR #410, opened 2021-12-09
Only include supported_groups extension in ClientHello

### LINKED ISSUE #409: Server Hello should not contain supported_groups extension (`extension.SupportedEllipticCurves`)
### Your environment.
  - Version: v2.0.9
  - Browser: N/A

Pion includes both the supported_groups ([also called](https://www.rfc-editor.org/rfc/rfc7919.html#section-7) elliptic_curves) and ec_point_formats extensions, whether sending a Client Hello or a Server Hello. (Refs: [flight 1](https://github.com/pion/dtls/blob/v2.0.9/flight1handler.go#L66-L73), [flight 3](https://github.com/pion/dtls/blob/v2.0.9/flight3handler.go#L148-L155) for Client Hello, [flight 4](https://github.com/pion/dtls/blob/v2.0.9/flight4handler.go#L192-L199) for Server Hello). ec_point_formats is fine in a Server Hello, but supported_groups is only defined for Client Hello.

The presence of supported_groups in Server Hello [has emerged as a concrete distinguisher](https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/snowflake/-/issues/40014#note_2765074) used to block Snowflake WebRTC connections in Russia:

> Russian DPI check supported_groups extension in ServerHello payload (byte 0x5a in udp packet). It looks for DTLS packet header "magic" "16 FE FD" and then looks for "1D 00 17 00 18" at 0x5a offset.

I think that supported_groups does not belong in Server Hello. These are my reasons for thinking so:
1. [Comparison with browser DTLS](https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/snowflake/-/issues/40014#note_2764715).
2. Go crypto/tls: `clientHelloMsg` [contains](https://github.com/golang/go/blob/0a15e7851a0ea1ebe1523bb70a6cfe56488ea2ef/src/crypto/tls/handshake_messages.go#L78-L79) `supportedCurves` and `supportedPoints` fields which may be marshaled, but `serverHelloMsg` only [contains](https://github.com/golang/go/blob/0a15e7851a0ea1ebe1523bb70a6cfe56488ea2ef/src/crypto/tls/handshake_messages.go#L608) and may marshal `supportedPoints`.
3. [RFC 8422](https://datatracker.ietf.org/doc/html/rfc8422) specifies both extensions for Client Hello, but only ec_point_formats for Server Hello.
   > ### [5.1](https://datatracker.ietf.org/doc/html/rfc8422#section-5.1). Client Hello Extensions
   >
   > This section specifies two TLS extensions that can be included with the ClientHello message as described in [RFC4366]: the Supported Elliptic Curves Extension and the Supported Point Formats Extension.

   > ### [5.2](https://datatracker.ietf.org/doc/html/rfc8422#section-5.2). Server Hello Extension
   >
   > This section specifies a TLS extension that can be included with the ServerHello message as described in [RFC4366], the Supported Point Formats Extension.

(Note, though, that in TLS 1.3, Server Hello [may contain](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.7) supported_groups—though TLS 1.3 also [removes](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2) ec_point_formats. I don't think it's relevant for DTLS 1.2.)

cohosh:
#### Description

This removes the logic that adds the supported_groups extension from the ServerHello message in flight4handler.go.

There is also a new test in conn_test.go to check this functionality. To write this test, I refactored the `sendClientHello` func defined in another test to be a standalone function.

#### Reference issue
Fixes #409

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -295,6 +295,35 @@ func testServer(ctx context.Context, c net.Conn, cfg *Config, generateCertificat
 	return ServerWithContext(ctx, c, cfg)
 }
 
+func sendClientHello(cookie []byte, ca net.Conn, sequenceNumber uint64, extensions []extension.Extension) error {
+	packet, err := (&recordlayer.RecordLayer{
+		Header: recordlayer.Header{
+			Version:        protocol.Version1_2,
+			SequenceNumber: sequenceNumber,
+		},
+		Content: &handshake.Handshake{
+			Header: handshake.Header{
+				MessageSequence: uint16(sequenceNumber),
+			},
+			Message: &handshake.MessageClientHello{
+				Version:            protocol.Version1_2,
+				Cookie:             cookie,
+				CipherSuiteIDs:     cipherSuiteIDs(defaultCipherSuites()),
+				CompressionMethods: defaultCompressionMethods(),
+				Extensions:         extensions,
+			},
+		},
+	}).Marshal()
+	if err != nil {
+		return err
+	}
+
+	if _, err = ca.Write(packet); err != nil {
+		return err
+	}
+	return nil
+}
+
 func TestHandshakeWithAlert(t *testing.T) {
 	// Limit runtime in case of deadlocks
 	lim := test.TimeOut(time.Second * 20)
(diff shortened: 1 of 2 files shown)


cohosh:
Oh! I should note that we've already shipped this fix in Snowflake, and it's working to bypass the DTLS fingerprinting attacks in Russia at the moment. More info: https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/snowflake/-/issues/40014

daenney (approved):

daenney:
This is released as v2.0.12 now.

ValZapod:
Thank you!

PR #415, opened 2022-01-15
Implement rfc7301

### LINKED ISSUE #408: Support for ALPN extension
## Summary

Support ALPN extensions for DTLS sessions. This would allow callers of the DTLS library to negotiate application layer protocols.

The way this is handled in crypto/tls is through the `NextProtos []String` field in the `Config` struct: https://pkg.go.dev/crypto/tls#Config

This change will require both client and server-side support of the ALPN extension.

## Motivation

This is a major difference between pion/dtls fingerprints and the DTLS fingerprints of browser-based WebRTC proxies. For more information on our ongoing work to explore fingerprinting attacks against Snowflake, see: https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/snowflake/-/issues/40014

The [ALPN extension](https://datatracker.ietf.org/doc/html/rfc7301) is used by browser WebRTC implementations to negotiate the usage of the WebRTC protocol over DTLS sessions.

This also brings this library more in line with the features offered by crypto/tls

arlolra:
Fixes #408

/cc @cohosh @uumaro

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -2064,6 +2064,214 @@ func TestServerNameIndicationExtension(t *testing.T) {
 	}
 }
 
+func TestALPNExtension(t *testing.T) {
+	// Limit runtime in case of deadlocks
+	lim := test.TimeOut(time.Second * 20)
+	defer lim.Stop()
+
+	// Check for leaking routines
+	report := test.CheckRoutines(t)
+	defer report()
+
+	for _, test := range []struct {
+		Name                   string
+		ClientProtocolNameList []string
+		ServerProtocolNameList []string
+		ExpectedProtocol       string
+		ExpectAlertFromClient  bool
+		ExpectAlertFromServer  bool
+		Alert                  alert.Description
+	}{
+		{
+			Name:                   "Negotiate a protocol",
+			ClientProtocolNameList: []string{"http/1.1", "spd/1"},
+			ServerProtocolNameList: []string{"spd/1"},
+			ExpectedProtocol:       "spd/1",
+			ExpectAlertFromClient:  false,
+			ExpectAlertFromServer:  false,
+			Alert:                  0,
+		},
+		{
+			Name:                   "Server doesn't support any",
+			ClientProtocolNameList: []string{"http/1.1", "spd/1"},
+			ServerProtocolNameList: []string{},
+			ExpectedProtocol:       "",
+			ExpectAlertFromClient:  false,
+			ExpectAlertFromServer:  false,
+			Alert:                  0,
+		},
+		{
+			Name:                   "Negotiate with higher server precedence",
+			ClientProtocolNameList: []string{"http/1.1", "spd/1", "http/3"},
+			ServerProtocolNameList: []string{"ssh/2", "http/3", "spd/1"},
+			ExpectedProtocol:       "http/3",
+			ExpectAlertFromClient:  false,
+			ExpectAlertFromServer:  false,
+			Alert:                  0,
+		},
+		{
+			Name:                   "Empty intersection",
+			ClientProtocolNameList: []string{"http/1.1", "http/3"},
+			ServerProtocolNameList: []string{"ssh/2", "spd/1"},
+			ExpectedProtocol:       "",
+			ExpectAlertFromClient:  false,
+			ExpectAlertFromServer:  true,
+			Alert:                  alert.NoApplicationProtocol,
+		},
+		{
+			Name:                   "Multiple protocols in ServerHello",
+			ClientProtocolNameList: []string{"http/1.1"},
+			ServerProtocolNameList: []string{"http/1.1"},
+			ExpectedProtocol:       "http/1.1",
+			ExpectAlertFromClient:  true,
+			ExpectAlertFromServer:  false,
+			Alert:                  alert.InternalError,
+		},
+	} {
+		test := test
+		t.Run(test.Name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+
+			ca, cb := dpipe.Pipe()
+			go func() {
+				conf := &Config{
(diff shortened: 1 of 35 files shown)


at-wat:
If the code is copied from golang/go, https://github.com/golang/go/blob/master/LICENSE must be copied to this repository according to the license.
I'm not sure it is acceptable. @Sean-Der @daenney  What do you think?

arlolra:
I didn't mean that literally.  There's only one function and a comment copied.  If there's a licensing issue, I can rewrite them.

at-wat:
Seems from https://github.com/golang/go/blob/2580d0e08d5e9f979b943758d3c49877fb2324cb/src/crypto/tls/handshake_server.go#L278-L304

I believe it is long enough to be protected by the copyright law of most of the countries, and ways to comply with the golang/go's license would be including the golang/go's license statement or rewriting as a new code.
(am not a copyright or legal professional and not 100% sure)

pkg/protocol/extension/alpn.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// NegotiateALPN picks a shared ALPN protocol that both sides support in server
+// preference order. If ALPN is not configured or the peer doesn't support it,
+// it returns "" and no error.
+func NegotiateALPN(serverProtos, clientProtos []string) (string, error) {
+	if len(serverProtos) == 0 || len(clientProtos) == 0 {
+		return "", nil
+	}
+	var http11fallback bool
```
daenney, line 70: I don't think we need this. We won't be carrying HTTP as this is the DTLS library, and h2 goes over TCP. h3 is a whole different thing entirely for which DTLS isn't used either.

That might also resolve the whole "did we copy too many lines" from Go source code issue.

arlolra: Yes, that's the one I was referring to in https://github.com/pion/dtls/pull/415#issuecomment-1013782468

arlolra:
I rewrote the necessary lines and force pushed

pkg/protocol/extension/alpn.go
daenney: It's nice how it solves 2 problems. Less code to maintain, no copyright issue. Thanks for touching it up.

daenney:
> Seems from https://github.com/golang/go/blob/2580d0e08d5e9f979b943758d3c49877fb2324cb/src/crypto/tls/handshake_server.go#L278-L304
> 
> I believe it is long enough to be protected by the copyright law of most of the countries, and ways to comply with the golang/go's license would be including the golang/go's license statement or rewriting as a new code. (am not a copyright or legal professional and not 100% sure)

We've resolved this, by essentially dropping the whole HTTP fallback handling code. We never supported ALPN before so the conditions that code were necessary for never occurred, and HTTP 1.1/2.0 doesn't use UDP for a transport anyway.

arlolra:
I rebased, split out a few commits and tried to address the linting issues CI is complaining about

arlolra:
The metadata failure seems odd considering there is no body,

```
Fix typo in ServerName comment

-------------------------------------------------
The preceding commit message is invalid
it failed 'Separate subject from body with a blank line' of the following checks
```

daenney:
Your first commit, the one that fixes the typo, doesn't have a body. We should allow commits without a body, but it currently trips up the linter.

conn_test.go
```diff
@@ -2064,6 +2064,131 @@ func TestServerNameIndicationExtension(t *testing.T) {
...
+			ClientProtocolNameList: []string{"http/1.1", "spd/1"},
+			ServerProtocolNameList: []string{"spd/1"},
+			Expected:               "spd/1",
+		},
+		{
+			Name:                   "Server doesn't support",
+			ClientProtocolNameList: []string{"http/1.1", "spd/1"},
+			ServerProtocolNameList: []string{},
+			Expected:               "",
+		},
```
daenney, line 2093: I think we should add a third test here. Multiple client and server protocols, where their order doesn't match and not all protocols overlap.

Something like:

```
ClientProtocolNameList: []string{"http/1.1", "spd/1", "http/3"},
ServerProtocolNameList: []string{"ssh/2", "http/3", "spd/1"},
Expected: "http/3"
```

arlolra: Done

arlolra:
Added the suggested test case and commit message bodies for those commits to appease the metadata linter.  I'll look into the missing coverage in a bit.

arlolra:
I added a test for the case where the protocol lists don't intersect; that should cover a few more lines.  I guess having the client receive the ServerHello (and invalid variations of it) would cover the rest.  Do you want that?

daenney:
> I added a test for the case where the protocol lists don't intersect; that should cover a few more lines. I guess having the client receive the ServerHello (and invalid variations of it) would cover the rest. Do you want that?

If it's not too big of a hassle then yeah, that would be awesome. That said, we're basically maintaining coverage now with this patch so I'm good with merging it either way.

arlolra:
> I guess having the client receive the ServerHello (and invalid variations of it) would cover the rest.

Ok, I added a test for that.  The tests feel a bit overloaded and could use some refactoring but hopefully it's a good enough start.

arlolra:
The error `Error: File is not `gofumpt`-ed (gofumpt)` seems transient

daenney:
Mmm indeed. Don't worry about it, I'll fix that up and merge it after.

arlolra:
> Mmm indeed. Don't worry about it, I'll fix that up and merge it after.

I amended the first commit with the changes necessary for the main branch, since that commit is already about formatting, and then fixed the complaint on the third patch that I introduced

Sean-Der:
This is a really amazing PR @arlolra. Really appreciate the high quality code and you cleaning up other stuff you have noticed :)

Sean-Der (approved):

Sean-Der:
Tagged `dtls/v2 v2.1.1` and `webrtc/v3 v3.1.18`

PR #416, opened 2022-01-18
Fix goroutine leak

lenaky:
#### Description

If Conn.handshake is returned by an error,
Inner goroutine which is calling Conn.readAndBuffer could not be returned
because Conn.handleIncomingPacket waits for channels Conn.closed and Conn.decrypted
and this could block infinitely.

Wait for ctxRead.Done as well

#### Reference issue
Fixes #...

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -564,7 +564,7 @@ func (c *Conn) readAndBuffer(ctx context.Context) error {
 
 	var hasHandshake bool
 	for _, p := range pkts {
-		hs, alert, err := c.handleIncomingPacket(p, true)
+		hs, alert, err := c.handleIncomingPacket(ctx, p, true)
 		if alert != nil {
 			if alertErr := c.notify(ctx, alert.Level, alert.Description); alertErr != nil {
 				if err == nil {
@@ -603,7 +603,7 @@ func (c *Conn) handleQueuedPackets(ctx context.Context) error {
 	c.encryptedPackets = nil
 
 	for _, p := range pkts {
-		_, alert, err := c.handleIncomingPacket(p, false) // don't re-enqueue
+		_, alert, err := c.handleIncomingPacket(ctx, p, false) // don't re-enqueue
 		if alert != nil {
 			if alertErr := c.notify(ctx, alert.Level, alert.Description); alertErr != nil {
 				if err == nil {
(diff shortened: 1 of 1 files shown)


lenaky:
```
0  0x000000000043cc65 in runtime.gopark
   at /__w/_tool/go/1.16.4/x64/src/runtime/proc.go:337
1  0x000000000044dd17 in runtime.selectgo
   at /__w/_tool/go/1.16.4/x64/src/runtime/select.go:327
2  0x00000000006e554c in github.com/pion/dtls/v2.(*Conn).handleIncomingPacket
   at /github/home/go/pkg/mod/github.com/pion/dtls/v2@v2.0.13/conn.go:741
3  0x00000000006e3f98 in github.com/pion/dtls/v2.(*Conn).readAndBuffer
   at /github/home/go/pkg/mod/github.com/pion/dtls/v2@v2.0.13/conn.go:565
4  0x00000000006f92b3 in github.com/pion/dtls/v2.(*Conn).handshake.func3
   at /github/home/go/pkg/mod/github.com/pion/dtls/v2@v2.0.13/conn.go:826
5  0x0000000000473ba1 in runtime.goexit
   at /__w/_tool/go/1.16.4/x64/src/runtime/asm_amd64.s:1371
```

conn.go
```diff
@@ -198,6 +198,7 @@ func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient
 	}
 	// Do handshake
 	if err := c.handshake(ctx, hsCfg, initialFlight, initialFSMState); err != nil {
+		c.close(false)
```
daenney, line 201: Probably want to turn it into `_ = c.close(false)`. Make it explicit that we're throwing the error away.

daenney:
@at-wat This seems like the correct fix to me. If you have time to double-check it that'd be awesome. Want to ensure I'm not misunderstanding how that bit is supposed to work.

at-wat:
Closing conn seems fixing the problem, but I think blocking-`select`s in that routine should be broken by `ctxRead.Done()` to make it safer for other potential cases.

lenaky:
@at-wat I agree. to do this, `handleIncomingPacket` must have ctxRead right?

at-wat:
https://github.com/pion/dtls/blob/fe3a675d7c926e91adc09540b4526236367f9e48/conn.go#L844-L847
https://github.com/pion/dtls/blob/fe3a675d7c926e91adc09540b4526236367f9e48/conn.go#L857-L860
as well?

conn.go
```diff
@@ -603,7 +603,7 @@ func (c *Conn) handleQueuedPackets(ctx context.Context) error {
 	c.encryptedPackets = nil
 
 	for _, p := range pkts {
-		_, alert, err := c.handleIncomingPacket(p, false) // don't re-enqueue
+		_, alert, err := c.handleIncomingPacket(context.Background(), p, false) // don't re-enqueue
```
at-wat, line 606:
```suggestion
		_, alert, err := c.handleIncomingPacket(ctx, p, false) // don't re-enqueue
```

lenaky:
Should I rewrite commit message? I guess need to be shortened.

at-wat:
Moving line break position to make each line <70 bytes should fix the lint error.

jkralik:
Hi. Can we merge this PR?

at-wat (approved):
LGTM

PR #417, opened 2022-01-26
Choose cipher suite matching the certificate

Sean-Der:
cherry-pick of https://github.com/pion/dtls/pull/399


lukaslihotzki:
Maybe the problem is that the cipher suites are cleared completely if the certificate is not found. In the original branch (https://github.com/Famedly/dtls/commit/f37ebb3609e6b7cd8742e9dddaa61445a37e84b3), I've amended the commit to keep all cipher suites if no certificate is found.

lukaslihotzki:
`go test` passes on my machine with my amended commit, but not with my original commit.

lukaslihotzki:
So the added lines need to be factored out into a separate function. You or me?

Sean-Der:
Hey @lukaslihotzki! Thanks for the super fast response. Sorry it took us so long to look at this :( 

I added you to the Pion organization, please push to the branch w/e you want!

Sean-Der:
If you can add a test this is a instant merge from me. I just don't want it to regress in the future (unless it is covered by tests already)

e2e/e2e_v113_test.go
```diff
@@ -30,7 +30,8 @@ func testPionE2ESimpleED25519(t *testing.T, server, client func(*comm)) {
 		dtls.TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8,
 		dtls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
 		dtls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
-		dtls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+		// Is TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 allowed with ed25519 keys?
```
daenney, line 33: Eh... yeah. I don't think it should be. The server key needs to be an RSA key in that case, which is no true here.

Sean-Der: @lukaslihotzki Looking really good! Thanks for fixing up so quick.

Sean-Der:
@daenney @at-wat @lukaslihotzki I just forced push to delete that RSA cipher from the ED25519 test. 
This PR LGTM to me code wise.

I would love to see a E2E test (or even just a unit test) to make sure that a refactor doesn't break this functionality in the future. @lukaslihotzki Do you have the time to do that? If not I can write one tonight.

lukaslihotzki:
@Sean-Der I have modeled the original problem with OpenSSL Client (#398) as E2E test.

lukaslihotzki:
@Sean-Der, why have you removed the E2E test addition?

Sean-Der:
Hey @lukaslihotzki 

Sorry for the churn. These are the changes I made/why I made them.

* Added error handling to `filterCipherSuitesForCertificate` call, before we ignored the error.

* Removed the E2E test and replaced with an explicit test for the behavior. I was worried that if someone regressed this feature it would be hard to deduced from the E2E test what exactly went wrong. It also would be easy to remove the check from the E2E test. Since it isn't targeted at testing a specific behavior it might be easy to miss in a review.

Fantastic work on the PR! Really appreciate your involvement.

PR #419, opened 2022-01-27
Only use algorithms specified in CertificateRequest msg

### LINKED ISSUE #418: Signature algorithm used in CertificateVerify selects algorithm not listed in CertificateRequest message
This is a violation of the RFC -  https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.8

      "The hash and signature algorithms used in the signature MUST be
      one of those present in the supported_signature_algorithms field
      of the CertificateRequest message"

### Your environment.
  - Version: verified problem in current code, last commit: 7d92e05b506557005042a79a3c62e17864afaa61
 
Server responds with a certificate request not containing SHA-256 listed as a signature hash algorithm
Client sends certificate verify response using SHA-256 as the hash algorithm

### What did you expect?
Client should have used one of the algorithms listed in the certificate request message

### What happened?
Connection failed - failed to verify certificate

I researched this and the root cause appears to be that in flight5handler.go it is using the local hash algorithms for selection, not those contained in the certificate request message.

https://github.com/pion/dtls/blob/fe3a675d7c926e91adc09540b4526236367f9e48/flight5handler.go#L178

The fix for this seems straightforward - we need to remember the hash algorithms listed in the certificate request message ( possibly in State ? )  and then use these in selection of the algorithm used to generate the CertificateVerify message.

mschexnaydre:
#### Description
We should only be using one of the signature algorithms specified in the CertificateRequest message when generating the CertificateVerify message.  Prior to this fix SHA-256 was always being used.

This change stores the HASH algorithm from the CertificateRequest message in the State object so that we can reference these later when generating the CertificateVerify message.

Removed hard-coded usage of SHA-256 in generateCertificateVerify, now uses the Digest method of the passed in algorithm.

#### Reference issue
Fixes #418

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -27,8 +27,10 @@ Jozef Kralik <jojo.lwin@gmail.com>
 Julien Salleyron <julien.salleyron@gmail.com>
 Kegan Dougal <kegan@matrix.org>
 Lander Noterman <lander.noterman@basalte.be>
+Len <len@hpcnt.com>
 Michael Zabka <zabka.michael@gmail.com>
 Michiel De Backker <mail@backkem.me>
+mschexnaydre <michael.schexnaydre@viasat.com>
 Robert Eperjesi <eperjesi@uber.com>
 Ryan Gordon <ryan.gordon@getcruise.com>
 Sean DuBois <seaduboi@amazon.com>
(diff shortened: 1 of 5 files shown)


mschexnaydre:
Fixing commit messages.  Will open new PR.

PR #421, opened 2022-01-28
Read ServerName from Client, not server Config

### LINKED ISSUE #319: Mistake of SNI extention handling
Maybe the code below is getting the server name from wrong place when the dtls server handling SNI extension.
I think the server should get the server name from the state of the connection, which variable name is `state` in this function scope. Am I right?

https://github.com/pion/dtls/blob/d01fc8597ebe97c464de6f45664a2340dd431bed/flight4handler.go#L204-L208

Sean-Der:
Resolves #319

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -2384,3 +2384,83 @@ func TestCipherSuiteMatchesCertificateType(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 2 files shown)


Sean-Der:
Will add test tomorrow, just wanted to get this easy one done :)

Sean-Der:
@daenney @at-wat ready for review!

PR #424, opened 2022-02-02
Fix DTLS client role in long delay connections

boks1971:
Fixes https://github.com/pion/webrtc/issues/2089

When a retranmission from the remote side arrives after
the handshake is complete, the `finish` routine
puts it back into retransmit loop.

With Chrome, this fails after 15 seconds.
Firefox does not error out though.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -9,6 +9,7 @@ Arlo Breault <arlolra@gmail.com>
 Atsushi Watanabe <atsushi.w@ieee.org>
 backkem <mail@backkem.me>
 bjdgyc <bjdgyc@163.com>
+boks1971 <raja.gobi@tutanota.com>
 Bragadeesh <bragboy@gmail.com>
 Carson Hoffman <c@rsonhoffman.com>
 Cecylia Bocovich <cohosh@torproject.org>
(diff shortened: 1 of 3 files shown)


at-wat (commented):
According to RFC6347, last flight should be retransmitted, even after entering the FINISHED state.
https://datatracker.ietf.org/doc/html/rfc6347#page-22
I think transition to the SENDING state was wrong, but retransmission without loop is still needed.

at-wat:
@boks1971 Your explanation seems right!

It would be better to add test cases that fail on master branch and pass on this branch.
(Maybe based on `TestHandshaker`?)

boks1971:
> @boks1971 Your explanation seems right!
> 
> It would be better to add test cases that fail on master branch and pass on this branch. (Maybe based on `TestHandshaker`?)

Will take a look @at-wat . Just started with this repo yesterday. So, it might take me a while to get a handle of how tests are written :-)

boks1971:
@at-wat I have added a couple of tests in `TestHandshaker` like you suggested to test this path. Ensured that it was failing without the fix and it passes after the fix. Please review when you get a chance.

boks1971:
Thank you for looking at this @at-wat . Can you approve this if you are happy with the change and tests?

at-wat (approved):
LGTM!

It might be nice if we can avoid increasing the test duration, but the test stability would have priority.

boks1971:
Thank you @at-wat . I increased the timeout as a precaution. It was not strictly needed. Just wanted to prioritise test stability like you mentioned just in case CI is having some trouble.

PR #428, opened 2022-02-10
Add cipher suite TLS_PSK_WITH_AES_256_CCM_8

### LINKED ISSUE #427: Add cipher suite TLS_PSK_WITH_AES_256_CCM_8
## Summary

Supporting TLS_PSK_WITH_AES_256_CCM_8 cipher suite out of the box. 

## Motivation

It could be injected as a customer cipher suite, however most code used for this is already present with
TLS_PSK_WITH_AES_128_CCM_8 cipher suite. I think the effort needed to built in this cipher suite to pion/dtls
or even the whole stack of AES_256_CCM cipher suites is rather small compared to the benefits and the extra
length a user has to go to implement his CustomCipherSuites (as he has no access to the internal and therefore needs to 
basically copy libraries code).

The alternative would be to use CustomCipherSuites, which I already tried successfully. However I think the extra milage
spent here is worth to integrate it into the module, since code from Aes128Ccm can be reused for Aes256Ccm.

ManuelBk:
#### Description
Refactoring to a more general AesCcm base class for old Aes128Ccm base class and new Aes256Ccm base class to allow easy creation of TLS_PSK_WITH_AES_256_CCM_8. Finally added this new ciphersuite.

#### Reference issue
Fixes #427

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ We would love contributions that fall under the 'Planned Features' and any bug f
 ##### PSK
 * TLS_PSK_WITH_AES_128_CCM ([RFC 6655][rfc6655])
 * TLS_PSK_WITH_AES_128_CCM_8 ([RFC 6655][rfc6655])
+* TLS_PSK_WITH_AES_256_CCM_8 ([RFC 6655][rfc6655])
 * TLS_PSK_WITH_AES_128_GCM_SHA256 ([RFC 5487][rfc5487])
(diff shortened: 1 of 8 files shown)


Sean-Der:
That works for me! We support `TLS_PSK_WITH_AES_128_CCM_8` so `TLS_PSK_WITH_AES_256_CCM_8` is fine with me :)

Can you put the new `TLS_PSK_WITH_AES_256_CCM_8` next to `TLS_PSK_WITH_AES_128_CCM_8` in the lists?

PR #430, opened 2022-02-18
Only filter CipherSuites to match certs on server

### LINKED ISSUE #429: Handshake failed. Broken by #398 "Server chooses cipher suite not matching its certificate"
In the scenario of that client cert is signed by RSA but server side only configures cipher with ECDSA, the DTLS handshake fails now. Which was working fine on or before v2.1.0. 

By looking at the recent commits and this one is the cause of the issue. "Server chooses cipher suite not matching its certificate #398". The fix for #398 put filter on client side also which cuts client supported cipher by half. This is not correct. The fix should only apply to server side not client side as my opinion. Please revisit #398 with a proper fix. 

Currently we have to stick on v2.1.0 in our pre production and product line. Please help to fix the issue as soon as possibly.

Sean-Der:
Resolves #429


Sean-Der:
@lukaslihotzki @wumiaoeric @jfdive Does this look ok/work for you?

@wumiaoeric is broken by b8f72f3774497508fd9bc960a676dc395c572416, this should unblock.

at-wat:
I think it's better to add a reference comment to RFC or any other DTLS libraries if possible

Sean-Der:
@at-wat Will do! I don't fully understand this behavior, will build OpenSSL locally and document it in Pion.

Sean-Der:
@at-wat Ok I see the original mistake! The RFC explicitly says `server's end-entity certificate.` not `end-entity certificate`, beyond that the logic is correct I believe.

PR #436, opened 2022-03-04
Add cipher suite TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256

jkralik:
#### Description

Extends for `TLS_ECDHE_PSK_*` ciphers + implementation TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256

### DIFF
--- a/cipher_suite.go
+++ b/cipher_suite.go
@@ -38,6 +38,8 @@ const (
 	TLS_PSK_WITH_AES_256_CCM_8      CipherSuiteID = ciphersuite.TLS_PSK_WITH_AES_256_CCM_8      //nolint:golint,stylecheck
 	TLS_PSK_WITH_AES_128_GCM_SHA256 CipherSuiteID = ciphersuite.TLS_PSK_WITH_AES_128_GCM_SHA256 //nolint:golint,stylecheck
 	TLS_PSK_WITH_AES_128_CBC_SHA256 CipherSuiteID = ciphersuite.TLS_PSK_WITH_AES_128_CBC_SHA256 //nolint:golint,stylecheck
+
+	TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256 CipherSuiteID = ciphersuite.TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256 //nolint:golint,stylecheck
 )
 
 // CipherSuiteAuthenticationType controls what authentication method is using during the handshake for a CipherSuite
@@ -48,6 +50,16 @@ const (
 	CipherSuiteAuthenticationTypeCertificate  CipherSuiteAuthenticationType = ciphersuite.AuthenticationTypeCertificate
 	CipherSuiteAuthenticationTypePreSharedKey CipherSuiteAuthenticationType = ciphersuite.AuthenticationTypePreSharedKey
 	CipherSuiteAuthenticationTypeAnonymous    CipherSuiteAuthenticationType = ciphersuite.AuthenticationTypeAnonymous
+)
+
+// CipherSuiteKeyExchangeAlgorithm controls what exchange algorithm is using during the handshake for a CipherSuite
+type CipherSuiteKeyExchangeAlgorithm = ciphersuite.KeyExchangeAlgorithm
+
+// CipherSuiteKeyExchangeAlgorithm Bitmask
+const (
+	CipherSuiteKeyExchangeAlgorithmNone  CipherSuiteKeyExchangeAlgorithm = ciphersuite.KeyExchangeAlgorithmNone
+	CipherSuiteKeyExchangeAlgorithmPsk   CipherSuiteKeyExchangeAlgorithm = ciphersuite.KeyExchangeAlgorithmPsk
+	CipherSuiteKeyExchangeAlgorithmEcdhe CipherSuiteKeyExchangeAlgorithm = ciphersuite.KeyExchangeAlgorithmEcdhe
 )
 
 var _ = allCipherSuites() // Necessary until this function isn't only used by Go 1.14
@@ -68,6 +80,13 @@ type CipherSuite interface {
 
 	// AuthenticationType controls what authentication method is using during the handshake
 	AuthenticationType() CipherSuiteAuthenticationType
+
+	// KeyExchangeAlgorithm controls what exchange algorithm is using during the handshake
+	KeyExchangeAlgorithm() CipherSuiteKeyExchangeAlgorithm
+
+	// ECC (Elliptic Curve Cryptography) determines whether ECC extesions will be send during handshake.
+	// https://datatracker.ietf.org/doc/html/rfc4492#page-10
+	ECC() bool
 
 	// Called when keying material has been generated, should initialize the internal cipher
 	Init(masterSecret, clientRandom, serverRandom []byte, isClient bool) error
@@ -120,6 +139,8 @@ func cipherSuiteForID(id CipherSuiteID, customCiphers func() []CipherSuite) Ciph
 		return &ciphersuite.TLSEcdheEcdsaWithAes256GcmSha384{}
 	case TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384:
 		return &ciphersuite.TLSEcdheRsaWithAes256GcmSha384{}
+	case TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256:
+		return ciphersuite.NewTLSEcdhePskWithAes128CbcSha256()
 	}
 
 	if customCiphers != nil {
--- a/conn.go
+++ b/conn.go
@@ -702,13 +702,12 @@ func (c *Conn) handleIncomingPacket(ctx context.Context, buf []byte, enqueue boo
 	} else if isHandshake {
 		markPacketAsValid()
 		for out, epoch := c.fragmentBuffer.pop(); out != nil; out, epoch = c.fragmentBuffer.pop() {
-			rawHandshake := &handshake.Handshake{}
-			if err := rawHandshake.Unmarshal(out); err != nil {
+			header := &handshake.Header{}
+			if err := header.Unmarshal(out); err != nil {
 				c.log.Debugf("%s: handshake parse failed: %s", srvCliStr(c.state.isClient), err)
 				continue
 			}
-
-			_ = c.handshakeCache.push(out, epoch, rawHandshake.Header.MessageSequence, rawHandshake.Header.Type, !c.state.isClient)
+			_ = c.handshakeCache.push(out, epoch, header.MessageSequence, header.Type, !c.state.isClient)
 		}
 
 		return true, nil, nil
--- a/flight1handler.go
+++ b/flight1handler.go
@@ -14,7 +14,7 @@ import (
 func flight1Parse(ctx context.Context, c flightConn, state *State, cache *handshakeCache, cfg *handshakeConfig) (flightVal, *alert.Alert, error) {
 	// HelloVerifyRequest can be skipped by the server,
 	// so allow ServerHello during flight1 also
-	seq, msgs, ok := cache.fullPullMap(state.handshakeRecvSequence,
+	seq, msgs, ok := cache.fullPullMap(state.handshakeRecvSequence, state.cipherSuite,
 		handshakeCachePullRule{handshake.TypeHelloVerifyRequest, cfg.initialEpoch, false, true},
 		handshakeCachePullRule{handshake.TypeServerHello, cfg.initialEpoch, false, true},
 	)
@@ -62,7 +62,16 @@ func flight1Generate(c flightConn, state *State, cache *handshakeCache, cfg *han
 			RenegotiatedConnection: 0,
 		},
 	}
-	if cfg.localPSKCallback == nil {
+
+	var setEllipticCurveCryptographyClientHelloExtensions bool
+	for _, c := range cfg.localCipherSuites {
+		if c.ECC() {
+			setEllipticCurveCryptographyClientHelloExtensions = true
+			break
+		}
+	}
+
+	if setEllipticCurveCryptographyClientHelloExtensions {
 		extensions = append(extensions, []extension.Extension{
 			&extension.SupportedEllipticCurves{
 				EllipticCurves: []elliptic.Curve{elliptic.X25519, elliptic.P256, elliptic.P384},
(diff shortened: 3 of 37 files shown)


jkralik:
@Sean-Der Is there anything else that needs to be done for this PR? Thx

Sean-Der:
No! Let me review now and get it merged, sorry about that @jkralik

flight1handler.go
```diff
@@ -62,7 +62,16 @@ func flight1Generate(c flightConn, state *State, cache *handshakeCache, cfg *han
 			RenegotiatedConnection: 0,
 		},
 	}
-	if cfg.localPSKCallback == nil {
```
Sean-Der, line 65: Is it possible to use PSK suites without `localPSKCallback` being set?

flight3handler.go
```diff
@@ -193,7 +193,17 @@ func handleServerKeyExchange(_ flightConn, state *State, cfg *handshakeConfig, h
 			return &alert.Alert{Level: alert.Fatal, Description: alert.InternalError}, err
 		}
 		state.IdentityHint = h.IdentityHint
-		state.preMasterSecret = prf.PSKPreMasterSecret(psk)
+		if state.cipherSuite == nil || !state.cipherSuite.UsesEllipticCurves() {
```
Sean-Der, line 196: I think `state.cipherSuite == nil` should be fatal. Maybe return a `alert.Fatal`?

It is great to check, but undefined behavior if we fall into the else I am worried about.

pkg/protocol/handshake/message_client_key_exchange.go
```diff
@@ -24,8 +24,21 @@ func (m MessageClientKeyExchange) Type() Type {
 // Marshal encodes the Handshake
 func (m *MessageClientKeyExchange) Marshal() ([]byte, error) {
 	switch {
-	case (m.IdentityHint != nil && m.PublicKey != nil) || (m.IdentityHint == nil && m.PublicKey == nil):
+	case (m.IdentityHint == nil && m.PublicKey == nil):
 		return nil, errInvalidClientKeyExchange
+	case (m.IdentityHint != nil && m.PublicKey != nil):
+		out := make([]byte, 2+len(m.IdentityHint)+1+len(m.PublicKey))
```
Sean-Der, line 30: Can you add some space to these operations? Maybe put allocations first, and then do copies? Somewhat harder to follow with no empty lines.

These are subjective, so feel free to push back.
* Use offsets instead of two slices. Users that don't know Go as well can be confused about reference vs value.
* Don't allocate `lenIdentityHint` and `lenPublicKey`. As a reader just have to go back and forth to confirm that everything is right.

flight4handler.go
```diff
@@ -319,36 +325,28 @@ func flight4Generate(c flightConn, state *State, cache *handshakeCache, cfg *han
 				},
 			})
 		}
-	case cfg.localPSKIdentityHint != nil:
+	case cfg.localPSKIdentityHint != nil || (state.cipherSuite != nil && state.cipherSuite.UsesEllipticCurves()):
```
Sean-Der, line 328: `state.cipherSuite` is asserted earlier (and fatal) so we shouldn't need this check anymore.

pkg/protocol/handshake/message_server_key_exchange.go
```diff
@@ -65,7 +68,15 @@ func (m *MessageServerKeyExchange) Unmarshal(data []byte) error {
 	if _, ok := elliptic.CurveTypes()[elliptic.CurveType(data[0])]; ok {
 		m.EllipticCurveType = elliptic.CurveType(data[0])
 	} else {
-		return errInvalidEllipticCurveType
+		// checking for ECDHE_PSK_*
```
Sean-Der, line 71: Parsing with heuristics seems dangerous to me, does the RFC recommend a different way?

I worry that parse fail + move pointer try again could lead to a bug.

pkg/crypto/prf/prf.go
```diff
@@ -74,6 +74,24 @@ func PSKPreMasterSecret(psk []byte) []byte {
 	return out
 }
 
+// EcdhePSKPreMasterSecret implements TLS 1.2 Premaster Secret generation given a psk, a keypair and a curve
+func EcdhePSKPreMasterSecret(psk, publicKey, privateKey []byte, curve elliptic.Curve) ([]byte, error) {
+	preMasterSecret, err := PreMasterSecret(publicKey, privateKey, curve)
+	if err != nil {
+		return nil, err
+	}
+	pskEnc := make([]byte, 2+len(preMasterSecret)+2+len(psk))
```
Sean-Der, line 83: Can you add some space to these operations? Maybe put allocations first, and then do copies? Somewhat harder to follow with no empty lines.

These are subjective, so feel free to push back.
* Use offsets instead of two slices. Users that don't know Go as well can be confused about reference vs value.

cipher_suite.go
```diff
@@ -69,6 +71,9 @@ type CipherSuite interface {
 	// AuthenticationType controls what authentication method is using during the handshake
 	AuthenticationType() CipherSuiteAuthenticationType
 
+	// UsesEllipticCurves determines whether elliptic curves will be used in the client hello message.
+	UsesEllipticCurves() bool
```
Sean-Der, line 75: Should this be `SendEllipticCurvesInClientHello`? Or is this really communicating `Use EcdhePSKPreMasterSecret when generating preMasterSecret`

Would it make sense to add `GeneratePreMasterSecret` to the CipherSuite interface?

flight1handler.go
jkralik: This is not about PSK suites. This setup extension is used for elliptic curves. A good example of this is the cipher [AECDH-AES128-SHA](https://ciphersuite.info/cs/TLS_ECDH_anon_WITH_AES_128_CBC_SHA), which I have used as a custom cipher to implement the [OCF specification](https://openconnectivity.org/specs/OCF_Security_Specification_v2.2.5.pdf).

cipher_suite.go
jkralik: Refactored, introduced `KeyExchangeAlgorithm()` which is more complaint with [RFC4279](https://datatracker.ietf.org/doc/html/rfc4279)

pkg/crypto/prf/prf.go
jkralik: fixed

pkg/protocol/handshake/message_server_key_exchange.go
jkralik: introduced `KeyExchangeAlgorithm`  as is in [RFC4279](https://datatracker.ietf.org/doc/html/rfc4279)

flight4handler.go
jkralik: fixed

pkg/protocol/handshake/message_client_key_exchange.go
jkralik: fixed

flight3handler.go
jkralik: fixed

Sean-Der:
@jkralik The PR looks fantastic to me! Having the constant for KeyExchange is amazing, makes all the existing code better as well. I love checking the algo and not just if PSK callback is nil.

Sean-Der:
This an approve from me. Really great work :)

@at-wat @daenney if you have some bandwidth mind looking as well? If you don't have time happy to merge this in a day or two.

conn.go
```diff
@@ -702,12 +703,20 @@ func (c *Conn) handleIncomingPacket(ctx context.Context, buf []byte, enqueue boo
...
+			rawHandshake := &handshake.Handshake{
+				KeyExchangeAlgorithm: keyExchangeAlgorithm,
+			}
 			if err := rawHandshake.Unmarshal(out); err != nil {
 				c.log.Debugf("%s: handshake parse failed: %s", srvCliStr(c.state.isClient), err)
 				continue
 			}
-
+			if m, ok := rawHandshake.Message.(*handshake.MessageServerHello); ok && m.CipherSuiteID != nil {
+				c.state.cipherSuite = cipherSuiteForID(ciphersuite.ID(*m.CipherSuiteID), c.fsm.cfg.customCipherSuites)
```
Sean-Der, line 718: in `flight3handler` we have a lot more logic and use `findMatchingCipherSuite`. 

I think we need to refactor it so decoding of messages happens when pulling from the cache?

@jkralik what do you think? I am happy to look into it also, this PR is so close it is a frustrating one :)

pkg/protocol/handshake/handshake.go
```diff
@@ -61,6 +62,27 @@ type Message interface {
 	Type() Type
 }
 
+// CipherSuiteAuthenticationType controls what authentication method is using during the handshake for a CipherSuite
```
Sean-Der, line 65: @jkralik Do we need to export these? 

It isn't great, but `CipherSuiteKeyExchangeAlgorithm` already exists in `github.com/pion/dtls` so seems confusing to export twice.

conn.go
Sean-Der: Oh nice I see a way out of this! We shouldn't unmarshal until CipherSuite has been set, let me see if I can get this working.

Will ping you @jkralik when done and get your take. If you are happy it is an approve from me.

Sean-Der:
@jkralik Mind looking at the changes I made and ok'ing?

* I removed the heuristics from `MessageClientKeyExchange` and it matches `ServerKeyExchange` now.
* MessageClientKeyExchange during the Marshal only has one branch. We don't use separate code IdentityHint, PublicKey and IdentityHint+PublicKey
* In Conn we just unmarshal the header. That way we don't need to know the `keyExchangeAlgorithm`

Sean-Der (approved):

at-wat:
https://github.com/pion/dtls/pull/436#issuecomment-1071204324
I'll take a look today

pkg/protocol/handshake/message_server_key_exchange.go
```diff
@@ -27,19 +31,22 @@ func (m MessageServerKeyExchange) Type() Type {
...
+		return out, nil
+	}
+	out = append(out, byte(m.EllipticCurveType), 0x00, 0x00)
+	binary.BigEndian.PutUint16(out[len(out)-2:], uint16(m.NamedCurve))
 
 	out = append(out, byte(len(m.PublicKey)))
 	out = append(out, m.PublicKey...)
 
-	if m.HashAlgorithm == hash.None && m.SignatureAlgorithm == signature.Anonymous && len(m.Signature) == 0 {
+	if m.HashAlgorithm == hash.None || m.SignatureAlgorithm == signature.Anonymous || len(m.Signature) == 0 {
```
at-wat, line 49: Are these OR cases all valid?

pkg/crypto/prf/prf.go
```diff
@@ -74,6 +74,32 @@ func PSKPreMasterSecret(psk []byte) []byte {
 	return out
 }
 
+// EcdhePSKPreMasterSecret implements TLS 1.2 Premaster Secret generation given a psk, a keypair and a curve
```
at-wat, line 77:
```suggestion
// EcdhePSKPreMasterSecret implements TLS 1.2 Premaster Secret generation given a psk, a keypair and a curve
//
// https://datatracker.ietf.org/doc/html/rfc5489#section-2
```

flight1handler.go
```diff
@@ -62,7 +62,16 @@ func flight1Generate(c flightConn, state *State, cache *handshakeCache, cfg *han
...
+
+	var setEllipticCurves bool
+	for _, c := range cfg.localCipherSuites {
+		if c.KeyExchangeAlgorithm().Has(CipherSuiteKeyExchangeAlgorithmEcdhe) {
```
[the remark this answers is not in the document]
jkralik: added

pkg/crypto/prf/prf.go
jkralik: fixed

internal/ciphersuite/types/exchangeType.go
```diff
@@ -0,0 +1,19 @@
...
+// KeyExchangeAlgorithm controls what exchange algorithm was chosen.
+type KeyExchangeAlgorithm int
+
+// KeyExchangeAlgorithm Enums
+const (
+	KeyExchangeAlgorithmNone KeyExchangeAlgorithm = 0
+	KeyExchangeAlgorithmPsk  KeyExchangeAlgorithm = iota << 1
+	KeyExchangeAlgorithmEcdhe
+
+	KeyExchangeAlgorithmEcdhePsk KeyExchangeAlgorithm = KeyExchangeAlgorithmPsk | KeyExchangeAlgorithmEcdhe
```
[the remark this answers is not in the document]
jkralik: `KeyExchangeAlgorithmEcdhePsk` has been removed

cipher_suite.go
```diff
@@ -50,6 +52,16 @@ const (
 	CipherSuiteAuthenticationTypeAnonymous    CipherSuiteAuthenticationType = ciphersuite.AuthenticationTypeAnonymous
 )
 
+// CipherSuiteKeyExchangeAlgorithm controls what exchange algorithm is using during the handshake for a CipherSuite
+type CipherSuiteKeyExchangeAlgorithm = ciphersuite.KeyExchangeAlgorithm
+
+// CipherSuiteKeyExchangeAlgorithm Enums
```
at-wat, line 58: Bitmask or bitset instead of enum maybe?

at-wat (approved):
LGTM

Sean-Der:
Feel free to merge w/e you want @jkralik!

Really appreciate your patience with this PR. This was a tricky one, but super happy with how it ended up.

pkg/protocol/handshake/message_server_key_exchange.go
Sean-Der: @jkralik

jkralik: Yes. Each field must be filled in otherwise the signature is not valid. (Used in the ANON cipher - all mentioned members are set as is in the condition)

at-wat: Isn't it better to return error on invalid cases?

jkralik: So I added checks for invalid cases.

jkralik:
> Feel free to merge w/e you want @jkralik!
> 
> Really appreciate your patience with this PR. This was a tricky one, but super happy with how it ended up.

Great. Thx for the code review.

PR #446, opened 2022-04-14
Errorf can't wrap the error

daenney:
We don't want to wrap anything here, only output it.

### DIFF
--- a/pkg/crypto/fingerprint/fingerprint_test.go
+++ b/pkg/crypto/fingerprint/fingerprint_test.go
@@ -7,8 +7,6 @@ import (
 	"testing"
 )
 
-var errInvalidHashID = errors.New("invalid hash ID")
-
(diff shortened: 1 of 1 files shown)


daenney:
Without this:

```
# github.com/pion/dtls/v2/pkg/crypto/fingerprint
pkg/crypto/fingerprint/fingerprint_test.go:50:3: (*testing.common).Errorf does not support error-wrapping directive %w
FAIL	github.com/pion/dtls/v2/pkg/crypto/fingerprint [build failed]
```

daenney:
I'm not even sure why we're outputting `errInvalidHashID` at all here?

Sean-Der (approved):

Sean-Der:
Yea not sure either! Honestly I think it would be fine to just `t.Fatal()` the conditional is pretty description

PR #447, opened 2022-04-14
Add support for session resumption when using client certificates

eh-steve:
#### Description
Previously, if a client submitted a certificate, any session resumption would be ignored in order to mitigate indefinite extension attacks as per https://curl.se/docs/CVE-2016-5419.html, but this prevented improving performance under mutual authentication use cases.

Now if `RequireAndVerifyClientCert`/`VerifyClientCertIfGiven` are used, the certificate expiry time is recorded in the session struct, allowing the user provided session store to decide what to do with "expired" sessions.

Revocations are still not handled via this mechanism, and the old behaviour can be preserved using the new config `PeerCertDisablesSessionResumption`


daenney:
This is going to need some tests.

eh-steve:
I can’t really tell from the logs whether the CI is failing due to my changes…

daenney:
Yeah, it's just the commit message. All the other checks passed.

Sean-Der:
What's your take on this @theodorsm @adrianosela?

It feels super valuable to not waste as much resources/save battery life. I don't want to risk security though. I can rebase/fix up if that helps.

PR #448, opened 2022-04-14
Implement retransmit backoff according to 4.2.4.1

eh-steve:
#### Description

As per [RFC4347](https://www.rfc-editor.org/rfc/rfc4347.txt)  4.2.4.1 (excluding reset after long period of idleness)

>  Though timer values are the choice of the implementation, mishandling
>    of the timer can lead to serious congestion problems; for example, if
>    many instances of a DTLS time out early and retransmit too quickly on
>    a congested link.  Implementations SHOULD use an initial timer value
> of 1 second (the minimum defined in RFC 2988 [RFC2988]) and double
>    the value at each retransmission, up to no less than the RFC 2988
>    maximum of 60 seconds.  Note that we recommend a 1-second timer
>    rather than the 3-second RFC 2988 default in order to improve latency
>    for time-sensitive applications.  Because DTLS only uses
>    retransmission for handshake and not dataflow, the effect on
>    congestion should be minimal.
> 
>    Implementations SHOULD retain the current timer value until a
>    transmission without loss occurs, at which time the value may be
>    reset to the initial value.  After a long period of idleness, no less
>    than 10 times the current timer value, implementations may reset the
>    timer to the initial value.  One situation where this might occur is
>    when a rehandshake is used after substantial data transfer.


eh-steve:
I can’t really tell from the logs whether the CI is failing due to my changes…

Sean-Der (approved):

Sean-Der:
Thank you so much @eh-steve 

Sorry it took such a long time to get to these. I am slowly making my way through the backlog

PR #449, opened 2022-04-18
Remove usage of x/xerrors

jech:
The pieces of x/xerrors that we use have been integrated into
stdlib in go-1.13, and we don't support any earlier versions.

Tested under go-1.18 and go-1.13.


daenney:
This seems to have a conflict on go.mod that needs to be resolved first.

daenney:
```
# github.com/pion/dtls/v2/pkg/crypto/signaturehash [github.com/pion/dtls/v2/pkg/crypto/signaturehash.test]
pkg/crypto/signaturehash/signaturehash_go113_test.go:38:31: undefined: errors
pkg/crypto/signaturehash/signaturehash_test.go:93:31: undefined: errors
```

jech:
Grr, I ran `go test` rather than `go test ./...`.  Sorry for that.

daenney (approved):

daenney:
Thanks for this!

PR #451, opened 2022-04-24
Remove spurious key/cert loading code

daenney:
We can use tls.LoadX509KeyPair from stdlib instead of our custom LoadKeyAndCertificate. This means we can also get rid of LoadKey, since we only need LoadCert for the client.


daenney:
Also means your example code now works if you supply an ED25519 key, since LoadX509KeyPair supports that, but our custom LoadKeyAndCertificate didn't.

Sean-Der (approved):

PR #463, opened 2022-05-11
Add handshake callback/monitor

alvarowolfx:
#### Description

This issue is a draft and I wanted to discuss a need that we had here in our project. We use the DTLS library in the IoT context, for IoT devices to authenticate to our system. One issue that we started seeing is that the auth/handshake can fail in some cases, and we want to keep track of when that happens. We have 3 major scenarios:
* Devices on a constrained network and might take too long to finish the handshake ( basically times out ) 
* The device informed the wrong set of credentials ( for now just PSK ID and PSK, but later we want to use certificates )
   *  Right now we can identify when the PSK ID is wrong, but when the PSK is not, even though I see crypto errors internally in the library, I could not capture that and It still falls under the previous "timeout" category. 

With this PR, I added a way to keep track of the Handshake as it progress, tracking the current Flight, Handshake State, and the connection itself ( so we can trace back to which client/device is using)

```go
OnHandshakeState: func(conn *dtls.Conn, f dtls.FlightVal, s dtls.HandshakeState, err error) {
			fmt.Printf("Handshake state %d: %s\n", f, s)
			if err != nil {
				pskID := conn.ConnectionState().IdentityHint
				fmt.Printf("Handshake error for device `%s`: %s\n", pskID, err)
			}
		},
```		

In our use case here, we save those error states as audit logs in our system so users can know when they have authentication errors on their system. 

I ended up having to make the types and constants related `FlightVal` and `HandshakeState` public to be shared on this function, so this generated many changes to the codebase. 

Again, I'm open to suggestions on better ways to track handshake errors and wanted to open this draft preview to see if this makes sense and can be merged or if we can do this differently. 

#### Reference issue
No related issue

### DIFF
--- a/config.go
+++ b/config.go
@@ -57,6 +57,9 @@ type Config struct {
 	PSK             PSKCallback
 	PSKIdentityHint []byte
 
+	// Callback to monitor handshake state
+	OnHandshakeState HandshakeCallback
+
 	// InsecureSkipVerify controls whether a client verifies the
 	// server's certificate chain and host name.
 	// If InsecureSkipVerify is true, TLS accepts any certificate
@@ -148,6 +151,9 @@ const defaultMTU = 1200 // bytes
 // PSKCallback is called once we have the remote's PSKIdentityHint.
 // If the remote provided none it will be nil
 type PSKCallback func([]byte) ([]byte, error)
+
+// HandshakeCallback is called multiple times as handshake progresses.
+type HandshakeCallback func(conn *Conn, f FlightVal, s HandshakeState, err error)
 
 // ClientAuthType declares the policy the server will follow for
 // TLS Client Authentication.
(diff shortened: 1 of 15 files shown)


daenney:
Thanks for starting a discussion around this plus the code. I think this is a useful thing to have in many scenarios and we should have a way to make the internals more accessible for debugging purposes like this.

As you noted you had to make a number of things public now which does increase our API surface quite a bit, for things that otherwise really are internal to the library. That's something I'm going to have to mull over for a bit, because once we do this we can't easily go back without needing to release a v3.

I'm hoping @at-wat and @Sean-Der will have a look too over the next few weeks, maybe come up with feedback/suggestions of their own.

Sean-Der:
Hi @alvarowolfx 

Sorry I didn't respond to this sooner. I have had this same question/problem and this is how I ended up solving it/my conclusion.

* With pion/webrtc I care about particular events. Usually as a result of a bug/issue happening and we start tracking it.
* I am able to capture the data by providing my own logger. Most things are logged that we care about.
* I don't want to expose more APIs because it makes the library harder to use for the majority of users.

I think the same sentiment applies to this library. If a highly technical user (as yourself) needs something I would say `pkg/protocol` and using a custom logger gets you everything. However providing more APIs tends to make the internals brittle/confusing.

If you disagree I am happy to re-visit this though! Thanks for taking the time to write this up, sorry I have been away from the project so long.

PR #464, opened 2022-05-15
Require CertificateVerify when Client Cert sent

Sean-Der:
Before a Client could send a certificate and the server would accept
without processing a CertificateVerify.


daenney (approved):

at-wat (approved):
LGTM

PR #470, opened 2022-05-22
Fix ED25519 client certificate

### LINKED ISSUE #469: Ed25519 keys are not fully supported.
I first noticed this issue trying to self sign a certificate with an Ed25519 key.  Ed25519 keys do not work for self-signing client certificates.

On the client:
```
handshake error: alert: Alert Fatal: BadCertificate
```

On the server:
```
handshake error: dtls fatal: expected and actual key signature do not match
```

But that is just the proximate problem I am having.  Several tests fail if you apply this diff, which switches to ed25519 keys in the `selfsigned` package.

```
diff --git a/pkg/crypto/selfsign/selfsign.go b/pkg/crypto/selfsign/selfsign.go
index 581e0a2..e3b57ac 100644
--- a/pkg/crypto/selfsign/selfsign.go
+++ b/pkg/crypto/selfsign/selfsign.go
@@ -5,7 +5,6 @@ import (
        "crypto"
        "crypto/ecdsa"
        "crypto/ed25519"
-       "crypto/elliptic"
        "crypto/rand"
        "crypto/rsa"
        "crypto/tls"
@@ -20,17 +19,16 @@ var errInvalidPrivateKey = errors.New("selfsign: invalid private key type")

 // GenerateSelfSigned creates a self-signed certificate
 func GenerateSelfSigned() (tls.Certificate, error) {
-       priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+       _, priv, err := ed25519.GenerateKey(rand.Reader)
        if err != nil {
                return tls.Certificate{}, err
        }
-
        return SelfSign(priv)
 }

 // GenerateSelfSignedWithDNS creates a self-signed certificate
 func GenerateSelfSignedWithDNS(cn string, sans ...string) (tls.Certificate, error) {
-       priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+       _, priv, err := ed25519.GenerateKey(rand.Reader)
        if err != nil {
                return tls.Certificate{}, err
        }
```

Failing tests:
```
--- FAIL: TestHandshakeWithAlert (0.41s)
    --- FAIL: TestHandshakeWithAlert/SignatureSchemesNoIntersection (0.00s)
--- FAIL: TestClientCertificate (0.42s)
    --- FAIL: TestClientCertificate/parallel (0.01s)
        --- FAIL: TestClientCertificate/parallel/NoClientCert_cert (0.00s)
        --- FAIL: TestClientCertificate/parallel/RequireAnyClientCert (0.00s)
        --- FAIL: TestClientCertificate/parallel/VerifyClientCertIfGiven_cert (0.00s)
        --- FAIL: TestClientCertificate/parallel/RequireAndVerifyClientCert (0.00s)
        --- FAIL: TestClientCertificate/parallel/RequestClientCert_cert (0.00s)
--- FAIL: TestServerCertificate (0.43s)
    --- FAIL: TestServerCertificate/parallel (0.02s)
        --- FAIL: TestServerCertificate/parallel/good_ca_skip_verify_custom_verify_peer (0.00s)
        --- FAIL: TestServerCertificate/parallel/good_ca_verify_custom_verify_peer (0.00s)
--- FAIL: TestErrorsTemporary (0.00s)
```

daenney:
#### Description

We were incorrectly hashing the handshakeBodies and passing that into Sign. However, for ed25519 you're not supposed to do that, it's part of the signature scheme (aside from the fact that it also uses SHA-512, not SHA-256).

This should now result in ED25519 client certificates working correctly.

#### Reference issue
Fixes #469

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -108,6 +108,13 @@ func verifyKeySignature(message, remoteKeySignature []byte, hashAlgorithm hash.A
 // the private key in the certificate.
 // https://tools.ietf.org/html/rfc5246#section-7.3
 func generateCertificateVerify(handshakeBodies []byte, privateKey crypto.PrivateKey, hashAlgorithm hash.Algorithm) ([]byte, error) {
+	if p, ok := privateKey.(ed25519.PrivateKey); ok {
+		// https://pkg.go.dev/crypto/ed25519#PrivateKey.Sign
+		// Sign signs the given message with priv. Ed25519 performs two passes over
+		// messages to be signed and therefore cannot handle pre-hashed messages.
+		return p.Sign(rand.Reader, handshakeBodies, crypto.Hash(0))
+	}
+
 	h := sha256.New()
 	if _, err := h.Write(handshakeBodies); err != nil {
 		return nil, err
@@ -115,9 +122,6 @@ func generateCertificateVerify(handshakeBodies []byte, privateKey crypto.Private
 	hashed := h.Sum(nil)
 
 	switch p := privateKey.(type) {
-	case ed25519.PrivateKey:
-		// https://crypto.stackexchange.com/a/55483
-		return p.Sign(rand.Reader, hashed, crypto.Hash(0))
 	case *ecdsa.PrivateKey:
 		return p.Sign(rand.Reader, hashed, hashAlgorithm.CryptoHash())
 	case *rsa.PrivateKey:
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"crypto/ed25519"
 	"crypto/rand"
+	"crypto/rsa"
 	"crypto/tls"
 	"errors"
 	"fmt"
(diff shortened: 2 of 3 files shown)


Sean-Der (approved):
Bug fix on a Sunday can’t beat that service:)

crypto.go
```diff
@@ -117,7 +117,13 @@ func generateCertificateVerify(handshakeBodies []byte, privateKey crypto.Private
...
 	case ed25519.PrivateKey:
 		// https://crypto.stackexchange.com/a/55483
-		return p.Sign(rand.Reader, hashed, crypto.Hash(0))
+		// [..] You do not have to hash the inputs to Ed25519 because hashing is already
+		// part of Ed25519 itself. If you do hash inputs in advance, you become vulnerable
+		// to collisions in the hash function you use. [..]
+		//
+		// ed25519.PrivateKey.Sign(): signs the given message with priv. Ed25519 performs
+		// two passes over messages to be signed and therefore cannot handle pre-hashed messages.
+		return p.Sign(rand.Reader, handshakeBodies, crypto.Hash(0))
```
at-wat, line 126: It looks better to skip https://github.com/pion/dtls/pull/470/files#diff-69c5c48c1da9bf0e59ac32c7d0c0089fee30329073ffcc86cf7200f224f44ecfR111-R115 which isn't used in this path

crypto.go
```diff
@@ -117,7 +117,13 @@ func generateCertificateVerify(handshakeBodies []byte, privateKey crypto.Private
 	switch p := privateKey.(type) {
 	case ed25519.PrivateKey:
 		// https://crypto.stackexchange.com/a/55483
-		return p.Sign(rand.Reader, hashed, crypto.Hash(0))
+		// [..] You do not have to hash the inputs to Ed25519 because hashing is already
+		// part of Ed25519 itself. If you do hash inputs in advance, you become vulnerable
+		// to collisions in the hash function you use. [..]
+		//
+		// ed25519.PrivateKey.Sign(): signs the given message with priv. Ed25519 performs
+		// two passes over messages to be signed and therefore cannot handle pre-hashed messages.
```
at-wat, line 125: Maybe better to refer https://pkg.go.dev/crypto/ed25519#PrivateKey.Sign
> Sign signs the given message with priv. Ed25519 performs two passes over messages to be signed and therefore **cannot handle pre-hashed messages**.

daenney: I'm not sure I understand, that's already in the comment? Or do you mean remove the Stackexchange link entirely?

crypto.go
daenney: GitHub is doing something wrong for me I think because that link ends up in the middle of `e2e_test.go` without highlighting anything. What do you want us to skip?

at-wat: https://github.com/pion/dtls/blob/8f9e598d8c5a129bc234f26abe7fdcf21b884626/crypto.go#L111-L115
`hashed` is not used in this `case`.

crypto.go
at-wat: I think it's better to refer only official API documents and RFCs rather than someone's stackexchange answer. (not very strong opinion)
At least we should add the source of reference of L124-125.

crypto.go
daenney: Ah, right yes. Lets fix that!

at-wat (approved):
LGTM:+1:

PR #474, opened 2022-05-24
Shuffle Elliptic Curves in ClientHello to circumvent Russian censorship

ValdikSS:
Since the December 2021, Russia is blocking Tor, including Snowflake
circumvention method which internally uses pion dtls for WebRTC.

Russian state DPI system fingerprints pion handshake features, namely
Elliptic Curves in ClientHello/ServerHello. The issue has been fixed
for ServerHello in 2f8ef4e48879f1129b3432c9ac04d2f4aad8049e
(by removing supported_groups extension from ServerHello), but
apparently ClientHello is also getting filtered.

Shuffle the order of Elliptic Curves in order to break the fingerprint
for ClientHello.

The fingerprint is performed by comparing the `00 1D 00 17 00 18` bytes at 0x6e offset of the packet, as well as at 0x59 offset.

More information:
https://gitlab.torproject.org/tpo/anti-censorship/censorship-analysis/-/issues/40030#note_2804998

### DIFF
--- a/flight1handler.go
+++ b/flight1handler.go
@@ -74,7 +74,7 @@ func flight1Generate(c flightConn, state *State, cache *handshakeCache, cfg *han
 	if setEllipticCurveCryptographyClientHelloExtensions {
 		extensions = append(extensions, []extension.Extension{
 			&extension.SupportedEllipticCurves{
-				EllipticCurves: []elliptic.Curve{elliptic.X25519, elliptic.P256, elliptic.P384},
+				EllipticCurves: []elliptic.Curve{elliptic.P256, elliptic.P384, elliptic.X25519},
 			},
 			&extension.SupportedPointFormats{
 				PointFormats: []elliptic.CurvePointFormat{elliptic.CurvePointFormatUncompressed},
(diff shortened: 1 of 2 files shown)


daenney:
> What about elliptic.X25519, elliptic.P384, elliptic.P256?

That solution has similar performance implications. Right now the ciphers are ordered from least to most computationally expensive, and I think that's a sensible default that should remain. If we want malleability here, it's going to need a thing on `dtls.Config` for it.

> Adding supported_groups extension before or after current position in the extension list would also work.

I suppose that's feasible without breaking any existing clients or changing the performance characteristics. Though I'd rather have the curves and their order configurable, as that's a generically useful thing beyond just the Russian filtering evasion use case. Otherwise it feels like we're playing implementation whack-a-mole by moving lines of code around and that doesn't seem very sustainable/stable long term. We'd also need to explicitly test for the order now to guarantee it doesn't get refactored in some way and regresses which I don't find very appealing.

I don't think we should be hardcoding things one way or another in this library for the sole purpose of enabling the Russian filter evasion or be constrained by that in your implementation. But I am happy to accept changes that make the library configurable enough to enable your use cases.

Sean-Der:
@ValdikSS Thanks for the PR, glad to see you are able to get around these restrictions :)

Would you be up for putting this in [Config](https://github.com/pion/dtls/blob/master/config.go#L20) so you can configure this yourself? 

Maybe we should introduce a system that lets user modify packets as they leave? That way if things get blocked in the future we can get around it again (without having to make changes to pion/dtls). Long term solution though!

daenney:
I've pushed #475 which makes the curves configurable. Feel free to try it out and let me know if this works for you!

PR #475, opened 2022-05-24
Make the Elliptic curves and order configurable

daenney:
This allows the client/server to chose which elliptic curves they wish to advertise, and in which order.

Fixes #474

### DIFF
--- a/config.go
+++ b/config.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"time"
 
+	"github.com/pion/dtls/v2/pkg/crypto/elliptic"
 	"github.com/pion/logging"
 )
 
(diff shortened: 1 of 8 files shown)


daenney:
Working on redoing the tests, since I realised that changing config in validateConfig wasn't the right place to do it.

config.go
```diff
@@ -144,6 +151,7 @@ func (c *Config) connectContextMaker() (context.Context, func()) {
 }
 
 const defaultMTU = 1200 // bytes
+var defaultCurves = []elliptic.Curve{elliptic.X25519, elliptic.P256, elliptic.P384}
```
daenney, line 154: Our linter doesn't like this one, but I'm not sure what else to do. I can't make it a `const` since Go doesn't accept `const` slices.

config.go
```diff
@@ -130,6 +131,12 @@ type Config struct {
 
 	// List of application protocols the peer supports, for ALPN
 	SupportedProtocols []string
+
+	// List of Elliptic Curves to use
+	//
+	// If an ECC ciperhsuite is configured and EllipticCurves is empty
```
kevmo314, line 137:
```suggestion
	// If an ECC ciphersuite is configured and EllipticCurves is empty
```

kevmo314 (approved):
Just a typo, otherwise this change looks good to me

PR #485, opened 2022-08-02
Implement GetCertificate and GetClientCertificate

### LINKED ISSUE #386: Support for GetCertificate and GetClientCertificate similar as in crypto/tls

jkralik:
#### Description
For rotating certificate's by K8s cert manager this feature is needed. The implementation is close to `crypto/tls` package.

#### Reference issue
Fixes #386

### DIFF
--- a/certificate.go
+++ b/certificate.go
@@ -1,35 +1,102 @@
 package dtls
 
 import (
+	"bytes"
 	"crypto/tls"
 	"crypto/x509"
+	"fmt"
 	"strings"
 )
 
-func (c *handshakeConfig) getCertificate(serverName string) (*tls.Certificate, error) {
+// ClientHelloInfo contains information from a ClientHello message in order to
+// guide application logic in the GetCertificate.
+type ClientHelloInfo struct {
+	// ServerName indicates the name of the server requested by the client
+	// in order to support virtual hosting. ServerName is only set if the
+	// client is using SNI (see RFC 4366, Section 3.1).
+	ServerName string
+
+	// CipherSuites lists the CipherSuites supported by the client (e.g.
+	// TLS_AES_128_GCM_SHA256, TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256).
+	CipherSuites []CipherSuiteID
+}
+
+// CertificateRequestInfo contains information from a server's
+// CertificateRequest message, which is used to demand a certificate and proof
+// of control from a client.
+type CertificateRequestInfo struct {
+	// AcceptableCAs contains zero or more, DER-encoded, X.501
+	// Distinguished Names. These are the names of root or intermediate CAs
+	// that the server wishes the returned certificate to be signed by. An
+	// empty slice indicates that the server has no preference.
+	AcceptableCAs [][]byte
+}
+
+// SupportsCertificate returns nil if the provided certificate is supported by
+// the server that sent the CertificateRequest. Otherwise, it returns an error
+// describing the reason for the incompatibility.
+// NOTE: original src: https://github.com/golang/go/blob/29b9a328d268d53833d2cc063d1d8b4bf6852675/src/crypto/tls/common.go#L1273
+func (cri *CertificateRequestInfo) SupportsCertificate(c *tls.Certificate) error {
+	if len(cri.AcceptableCAs) == 0 {
+		return nil
+	}
+
+	for j, cert := range c.Certificate {
+		x509Cert := c.Leaf
+		// Parse the certificate if this isn't the leaf node, or if
+		// chain.Leaf was nil.
+		if j != 0 || x509Cert == nil {
+			var err error
+			if x509Cert, err = x509.ParseCertificate(cert); err != nil {
+				return fmt.Errorf("failed to parse certificate #%d in the chain: %w", j, err)
+			}
+		}
+
+		for _, ca := range cri.AcceptableCAs {
+			if bytes.Equal(x509Cert.RawIssuer, ca) {
+				return nil
+			}
+		}
+	}
+	return errNotAcceptableCertificateChain
+}
+
+func (c *handshakeConfig) setNameToCertificateLocked() {
+	nameToCertificate := make(map[string]*tls.Certificate)
+	for i := range c.localCertificates {
+		cert := &c.localCertificates[i]
+		x509Cert := cert.Leaf
+		if x509Cert == nil {
+			var parseErr error
+			x509Cert, parseErr = x509.ParseCertificate(cert.Certificate[0])
+			if parseErr != nil {
+				continue
+			}
+		}
+		if len(x509Cert.Subject.CommonName) > 0 {
+			nameToCertificate[strings.ToLower(x509Cert.Subject.CommonName)] = cert
+		}
+		for _, san := range x509Cert.DNSNames {
+			nameToCertificate[strings.ToLower(san)] = cert
+		}
+	}
+	c.nameToCertificate = nameToCertificate
+}
+
+func (c *handshakeConfig) getCertificate(clientHelloInfo *ClientHelloInfo) (*tls.Certificate, error) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	if c.nameToCertificate == nil {
-		nameToCertificate := make(map[string]*tls.Certificate)
-		for i := range c.localCertificates {
-			cert := &c.localCertificates[i]
-			x509Cert := cert.Leaf
-			if x509Cert == nil {
-				var parseErr error
-				x509Cert, parseErr = x509.ParseCertificate(cert.Certificate[0])
-				if parseErr != nil {
-					continue
-				}
-			}
-			if len(x509Cert.Subject.CommonName) > 0 {
-				nameToCertificate[strings.ToLower(x509Cert.Subject.CommonName)] = cert
-			}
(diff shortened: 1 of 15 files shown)


flight4handler.go
```diff
@@ -318,15 +322,25 @@ func flight4Generate(c flightConn, state *State, cache *handshakeCache, cfg *han
...
 
 		if cfg.clientAuth > NoClientCert {
+			// An empty list of certificateAuthorities signals to
+			// the client that it may send any certificate in response
+			// to our request. When we know the CAs we trust, then
+			// we can send them down, so that the client can choose
+			// an appropriate certificate to give to us.
+			var certificateAuthorities [][]byte
+			if cfg.clientCAs != nil {
+				certificateAuthorities = cfg.clientCAs.Subjects()
```
at-wat, line 332: `SertPool.Subjects()` seems be marked as deprecated since Go 1.18.
https://pkg.go.dev/crypto/x509#CertPool.Subjects
https://tip.golang.org/doc/go1.18#crypto/x509
I haven't dug into the reason and alternatives.

pkg/protocol/handshake/message_certificate_request_test.go
```diff
@@ -28,8 +29,8 @@ func TestHandshakeMessageCertificateRequest(t *testing.T) {
 			{Hash: hash.SHA512, Signature: signature.RSA},
 			{Hash: hash.SHA1, Signature: signature.RSA},
 		},
+		CertificateAuthoritiesNames: [][]byte{[]byte("test")},
```
at-wat, line 32: Please separate test case to check behavior for empty CA names as well.

flight5handler.go
```diff
@@ -63,33 +63,40 @@ func flight5Parse(ctx context.Context, c flightConn, state *State, cache *handsh
...
-		pkts = append(pkts,
-			&packet{
-				record: &recordlayer.RecordLayer{
-					Header: recordlayer.Header{
-						Version: protocol.Version1_2,
-					},
-					Content: &handshake.Handshake{
-						Message: &handshake.MessageCertificate{
-							Certificate: certBytes,
+		if certificate != nil && certificate.Certificate != nil {
```
at-wat, line 84: In what case this condition become false, and is it ok not to return error?

at-wat (commented):
Some code seems be copied from Go's source code, so we should add license information and copyright holder of the original code (BSD, The Go Authors).
(maybe from before?)

certificate.go
```diff
@@ -1,35 +1,101 @@
...
+	}
+
+	for j, cert := range c.Certificate {
+		x509Cert := c.Leaf
+		// Parse the certificate if this isn't the leaf node, or if
+		// chain.Leaf was nil.
+		if j != 0 || x509Cert == nil {
+			var err error
+			if x509Cert, err = x509.ParseCertificate(cert); err != nil {
+				return fmt.Errorf("failed to parse certificate #%d in the chain: %w", j, err)
```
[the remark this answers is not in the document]
jkralik: This is intentional as is in `crypto/tls`:
```
        // If Certificate.Certificate is empty then no certificate will be sent to
	// the server. If this is unacceptable to the server then it may abort
	// the handshake.
```

jkralik:
> Some code seems be copied from Go's source code, so we should add license information and copyright holder of the original code (BSD, The Go Authors). (maybe from before?)

I will put NOTE with the link to original code.

flight4handler.go
jkralik: Yes. I know, but `CertPool` doesn't provide any alternatives. When I look to source code of go 1.18 they still use it:
https://cs.opensource.google/go/go/+/master:src/crypto/tls/handshake_server_tls13.go;l=593;drc=fe4de36198794c447fbd9d7cc2d7199a506c76a5

e2e/e2e_test.go
```diff
@@ -391,7 +392,15 @@ func testPionE2ESimpleED25519ClientCert(t *testing.T, server, client func(*comm)
...
 
+	clientCAs := x509.NewCertPool()
+	caCert, err := x509.ParseCertificate(ccert.Certificate[0])
+	if err != nil {
+		t.Fatal(err)
+	}
+	clientCAs.AddCert(caCert)
+
 	scfg := &dtls.Config{
+		ClientCAs:    clientCAs,
```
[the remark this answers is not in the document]
jkralik: Ok

pkg/protocol/handshake/message_certificate_request.go
```diff
@@ -96,5 +110,32 @@ func (m *MessageCertificateRequest) Unmarshal(data []byte) error {
 		m.SignatureHashAlgorithms = append(m.SignatureHashAlgorithms, signaturehash.Algorithm{Signature: s, Hash: h})
 	}
 
+	offset += signatureHashAlgorithmsLength
+	if len(data) < offset+2 {
+		return nil
+	}
```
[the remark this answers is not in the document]
jkralik: fixed

pkg/protocol/handshake/message_certificate_request_test.go
jkralik: fixed

flight5handler.go
jkralik: When the server CA names do not match the issuers on the client's certificate, this happens. This is the case when the client does not return the certificate.

config_test.go
```diff
@@ -116,4 +116,26 @@ func TestValidateConfig(t *testing.T) {
...
+	// Valid config with get client certificate
+	config = &Config{
+		CipherSuites: []CipherSuiteID{TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256},
+		GetClientCertificate: func(cri *CertificateRequestInfo) (*tls.Certificate, error) {
+			return &tls.Certificate{Certificate: cert.Certificate, PrivateKey: rsaPrivateKey}, nil
+		},
+	}
+	if err = validateConfig(config); err != nil {
+		t.Fatalf("TestValidateConfig: Client error exp(%v) failed(%v)", nil, err)
+	}
```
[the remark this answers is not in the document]
jkralik: fixed

pkg/crypto/selfsign/selfsign.go
```diff
@@ -40,7 +40,7 @@ func GenerateSelfSignedWithDNS(cn string, sans ...string) (tls.Certificate, erro
 
 // SelfSign creates a self-signed certificate from a elliptic curve key
 func SelfSign(key crypto.PrivateKey) (tls.Certificate, error) {
-	return WithDNS(key, hex.EncodeToString(make([]byte, 16)))
+	return WithDNS(key, "self-signed cert" /*hex.EncodeToString(make([]byte, 16)*/)
```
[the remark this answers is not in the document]
jkralik: removed

flight4handler.go
at-wat: Looks fine to keep using as long as crypto/tls is using too.

jkralik:
@at-wat @daenney @Sean-Der Hi, sorry for interrupting, but is there something that needs to be done to review this PR? Thx

flight5handler.go
at-wat: https://www.rfc-editor.org/rfc/rfc5246.html#section-7.4.6
> This message is only sent if the server requests a certificate.  If no suitable certificate is available, the client MUST send a certificate message containing no certificates.

In my understanding, we need to send an empty `handshake.MessageCertificate{}` when this condition is not satisfied.

jkralik: Thx. Fixed

daenney:
> @at-wat @daenney @Sean-Der Hi, sorry for interrupting, but is there something that needs to be done to review this PR? Thx

I think now that the last of @at-wat's review comments have been resolved there shouldn't be anything holding it up. Once they approve I'd say go ahead and merge+tag.

flight5handler.go
```diff
@@ -63,33 +63,42 @@ func flight5Parse(ctx context.Context, c flightConn, state *State, cache *handsh
...
-		pkts = append(pkts,
-			&packet{
-				record: &recordlayer.RecordLayer{
-					Header: recordlayer.Header{
-						Version: protocol.Version1_2,
-					},
-					Content: &handshake.Handshake{
-						Message: &handshake.MessageCertificate{
-							Certificate: certBytes,
+		if certificate != nil {
```
at-wat, line 84: `certificate != nil` seems only happens when `Config.GetClientCertificate()` returns nil and
https://github.com/pion/dtls/blob/e547c81e37695577e1d4fa3176406dc2f3c2492b/config.go#L164-L166
So, should it be error if `certificate == nil`?

jkralik: Sure. Fixed !

at-wat (approved):
LGTM!

PR #487, opened 2022-08-02
Implement VerifyConnection as is in tls.Config

### LINKED ISSUE #486: Add Config.VerifyConnection callback
## Summary

Allows to close connection during handshake process. https://github.com/golang/go/issues/36736

## Motivation

Have similar implementation as is in tls.Config.

jkralik:
#### Description

VerifyConnection, if not nil, is called after normal certificate
verification/PSK and after VerifyPeerCertificate by either a TLS client
or server. If it returns a non-nil error, the handshake is aborted
and that error results.

If normal verification fails then the handshake will abort before
considering this callback. This callback will run for all connections
regardless of InsecureSkipVerify or ClientAuth settings.

#### Reference issue
Fixes #486

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -15,6 +15,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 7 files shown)


kevmo314 (approved):

Sean-Der (approved):
LGTM! Thanks for adding @jkralik 

I don't have a lot of context, but this is just duplicating behavior that was added to the stdlib so I am all for that :)

conn_test.go
```diff
@@ -528,11 +550,22 @@ func TestPSK(t *testing.T) {
...
-				PSKIdentityHint: test.ServerIdentity,
-				CipherSuites:    test.CipherSuites,
+				PSKIdentityHint:  test.ServerIdentity,
+				CipherSuites:     test.CipherSuites,
+				VerifyConnection: test.ServerVerifyConnection,
 			}
 
 			server, err := testServer(ctx, cb, config, false)
+			if test.WantFail {
+				if err == nil {
```
at-wat, line 560: It would be better to check which error was returned.

jkralik: fixed

conn_test.go
```diff
@@ -466,15 +484,36 @@ func TestPSK(t *testing.T) {
...
-		Name           string
-		ServerIdentity []byte
-		CipherSuites   []CipherSuiteID
+		Name                   string
+		ServerIdentity         []byte
+		CipherSuites           []CipherSuiteID
+		ClientVerifyConnection func(*State) error
+		ServerVerifyConnection func(*State) error
+		WantFail               bool
 	}{
```
at-wat, line 493: Maybe adding
```
ExpectedServerErr string
ExpectedClientErr string
```
to each `WantFail=true` test case and checking the expected string is container would be easier to read?

jkralik: Sure. Fixed

at-wat (approved):
:+1:

PR #490, opened 2022-08-11
Close connection when handshake timeout occurs

### LINKED ISSUE #483: not close udp conn when dtls handshake failed？
### Your environment.
  - Version: *2.1.5*
  - Browser: *include version*

### What did you do?

when client connect to dtls service, if psk callback return error, handshake also return error but not return conn. I see the following code in conn.go:887, psk error not in fatal and close, so handshake function is close, but udp conn is running, how to close udp conn？
```
if e != nil {
	if e.IsFatalOrCloseNotify() {
		_ = c.close(false) //nolint:contextcheck
	}
}
```

jkralik:
Fixes #483

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -896,6 +896,10 @@ func (c *Conn) handshake(ctx context.Context, cfg *handshakeConfig, initialFligh
 						_ = c.close(false) //nolint:contextcheck
 					}
 				}
+				if !c.isConnectionClosed() && errors.Is(err, context.Canceled) {
+					c.log.Trace("handshake timeouts - closing underline connection")
+					_ = c.close(false) //nolint:contextcheck
+				}
 				return
 			}
 		}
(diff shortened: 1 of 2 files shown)


jkralik:
@at-wat @Sean-Der Hi, sorry for interrupting, but is there something that needs to be done to review this PR? Thx

conn.go
```diff
@@ -941,13 +947,18 @@ func (c *Conn) close(byUser bool) error {
...
+	isClosed := c.isConnectionClosed()
 	c.closed.Close()
 	c.closeLock.Unlock()
 
 	if closedByUser {
 		return ErrConnClosed
 	}
 
+	if isClosed {
+		return nil
```
at-wat, line 959: `crypto/tls.Conn` (based on `net.TCPConn`) seems to return `use of closed network connection` error if closed twice.

at-wat: Never mind.
In the current implementation, first `Close()` after timeout returns `nil` by this condition and second `Close()` returns `ErrConnClosed` by the above condition. It matches `crypto/tls.Conn`'s behavior.

at-wat (approved):
LGTM

PR #507, opened 2022-12-31
Update hash name check to be case insensitive

kevmo314:
Some SDPs emit uppercase hash names. This change makes the comparison case insensitive.


daenney:
Thanks for this. It needs a rebase now that #508 is merged in order to make the tests pass.

at-wat (approved):
The change looks reasonable.

In RFC4572, the syntax uses only lower cases:
https://www.rfc-editor.org/rfc/rfc4572#page-8
but the example uses upper case:
https://www.rfc-editor.org/rfc/rfc4572#section-3.4
I'm not 100% sure but upper case seems be allowed.

PR #516, opened 2023-01-22
Fix fuzzing of recordLayer

stv0g:
See for the following failed action:

https://github.com/pion/stun/actions/runs/3979448280/jobs/6821876144

### DIFF
--- a/pkg/protocol/recordlayer/fuzz.go
+++ b/pkg/protocol/recordlayer/fuzz.go
@@ -1,18 +1,20 @@
 //go:build gofuzz
 // +build gofuzz
 
-package dtls
+package recordlayer
 
-import "fmt"
+import (
+	"fmt"
+)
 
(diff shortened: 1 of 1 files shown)


ashellunts (approved):
👍

daenney:
I don't really get what this supposed to fix? It seems to move the code into a new package, but nothing about the code itself changes.

stv0g:
> I don't really get what this supposed to fix? 

The old code was broken by some earlier refactoring and likely never tested due to the build tag.
Here are the exact errors:

```
Error: fuzz.go:8:33: undefined: recordlayer
Error: fuzz.go:15:8: undefined: recordLayer
Error: fuzz.go:26:9: undefined: recordLayer
```

> The code has a conditional build tag, pion/stun should never be running into this or setting it.

Yes, I am also surprised that the CI workflow of pion/stun was running into this issue.

PR #519, opened 2023-01-23
Update github.com/pion/transport to v2

stv0g:
Result in no changes to the API as we use the package mostly for testing


at-wat (approved):
LGTM

stv0g:
Thanks for the review :)

PR #534, opened 2023-03-11
Cleanup README

stv0g:


README.md
```diff
@@ -132,8 +132,19 @@ go run examples/dial/psk/main.go
   openssl s_client -dtls1_2 -connect 127.0.0.1:4444 -psk abc123 -cipher PSK-AES128-CCM8
 ```
 
+### Roadmap
+The library is used as a part of our WebRTC implementation. Please refer to that [roadmap](https://github.com/pion/webrtc/issues/9) to track our major milestones.
```
daenney, line 136: I'd like to remove this. The DTLS library is used well beyond WebRTC and we don't plan future work on DTLS in the webrtc project.

README.md
```diff
@@ -132,8 +132,19 @@ go run examples/dial/psk/main.go
...
+
+### Community
+Pion has an active community on the [Slack](https://pion.ly/slack).
+
+Follow the [Pion Twitter](https://twitter.com/_pion) for project updates and important WebRTC news.
```
daenney, line 141: Given the current state of Twitter, I would suggest we don't push towards that right now and leave that out of the README until that situation has calmed down.

daenney (changes requested):

README.md
stv0g: I fully agree👍 How about a Mastodon account for news at Fosstodon.org?

daenney: That would be nice. Since @Sean-Der normally owns all the accounts for this stuff it's probably best if they set it up though.

README.md
stv0g: Alright. I removed the section.

However, regarding your second point: I still see DTLS as an important component for WebRTC and there are definitely others how contribute to this repo with having better interoperability of Pion's WebRTC/TURN implementation in mind.

README.md
stv0g: Thanks for pinging Sean. I will wait until he replies and then adjust this notice in all Pion repos..

davidzhao: The Pion account has 1700 followers and it's arguably where Pion has the largest reach. I would suggest not removing our active social accounts because of personal feelings about Twitter.

I'm in favor of also adding Mastodon if there's a volunteer to maintain it.

stv0g: Thats a good point. It also reflects Sean's opinion on this as discussed via Slack

daenney (approved):

stv0g:
Thanks @daenney :)

PR #541, opened 2023-04-18
Update CI configs to v0.10.7

pionbot:


.golangci.yml
```diff
@@ -10,7 +13,15 @@ linters-settings:
       modules:
         - github.com/pkg/errors:
             recommendations:
-            - errors
+              - errors
+  forbidigo:
+    forbid:
+      - Fatal(f|ln)?$
```
daenney, line 19: This needs to be removed for DTLS. The protocol itself defines a `Fatal` alert level that must be sent to the other party in case certain errors occur.

daenney (changes requested):
This needs to be removed for DTLS. The protocol itself defines a `Fatal` alert level that must be sent to the other party in case certain errors occur.

.golangci.yml
stv0g: I see.. Let me check what I can do. Would you be fine with using `//nolint` directives?

Otherwise, I need to build a workaround which would allow us to customize the `golangci-lint.yaml` per-repo..

daenney: If the `//nolint` ones work that's totally fine

stv0g: Argh, there are 116 occurences of `alert.Fatal`. Adding exceptions for all of them is probably not the best way here :(

stv0g: I changed the regex which catches `Fatal` usage. It should work now.

stv0g (approved):

.golangci.yml
daenney: Nice!

daenney (approved):
:shipit:

examples/certificates/client.pem
daenney, line 1: I'm somewhat surprised the tests haven't exploded on having extra data like a license header in the pem files...

daenney: I guess the functions loading these files ignore everything before the `----- BEGIN` etc. Which is good :smile:.

stv0g: Yeah :) I've actually had a look at the PEM parser to verify this..

PR #543, opened 2023-04-30
Avoid a copy in compactRawPackets

jech:
The method compactRawPackets accumulates packets into MTU-sized
chunks.  Unfortunately, it systematically copies the list of
packets.

Fortunately, it's usually called with a list of just one packet,
in which case there's nothing to do.  Rather than rewriting the
code to avoid allocation altogether, we just optimise away the
trivial case.

This reduces the amount of garbage generated during bulk transfers
by some 10%.


daenney (approved):
Hah, nice :rocket:

PR #553, opened 2023-06-26
Remove extraneous error checks in handshaker

hasheddan:
#### Description

Removes extraneous checks for flight error after alert error occurs. If we are sending alert then we already know that a flight error occurred.


daenney (approved):
That seems pretty reasonable. I'm not really sure why the additional `if errFlight != nil` got in there, we're already in that branch.

PR #554, opened 2023-06-29
Wait for OpenSSL server shutdown in e2e test

### LINKED ISSUE #526: OpenSSL E2E Tests are broken in Go 1.20
PR is #525 

Failing tests: https://github.com/pion/dtls/actions/runs/4092274657/jobs/7056887569

Alle the OpenSSL tests are failing in the Goroutine check:

```
  util.go:41: Unexpected routines on test startup: 
        goroutine 1098 [chan send, 2 minutes]:
        os/exec.(*Cmd).watchCtx(0xc0002e4160, 0xc00006a1e0)
        	/usr/local/go/src/os/exec/exec.go:782 +0x413
        created by os/exec.(*Cmd).Start
        	/usr/local/go/src/os/exec/exec.go:743 +0x9f3
        
        goroutine 1067 [chan send, 2 minutes]:
        os/exec.(*Cmd).watchCtx(0xc0002e4000, 0xc000083140)
        	/usr/local/go/src/os/exec/exec.go:782 +0x413
        created by os/exec.(*Cmd).Start
        	/usr/local/go/src/os/exec/exec.go:743 +0x9f3
        
        goroutine 1114 [chan send, 2 minutes]:
        os/exec.(*Cmd).watchCtx(0xc00010c000, 0xc00024bc20)
        	/usr/local/go/src/os/exec/exec.go:782 +0x413
        created by os/exec.(*Cmd).Start
        	/usr/local/go/src/os/exec/exec.go:743 +0x9f3
```

I don't really know why this is happening.. Maybe @jkralik or @daenney have an idea?

hasheddan:
#### Description

Updates e2e tests to wait until server shutdown before completing to
ensure that we do not trigger errors due to leaked watchCtx goroutine,
which was added to os/exec.Command in Go 1.20. We do so by moving from
implicit shutdown via context cancellation to explicit shutdown.

https://tip.golang.org/doc/go1.20

#### Reference issue

Fixes #526 
Closes #525

### DIFF
--- a/e2e/Dockerfile
+++ b/e2e/Dockerfile
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
 # SPDX-License-Identifier: MIT
 
-FROM docker.io/library/golang:1.18-bullseye
+FROM docker.io/library/golang:1.20-bullseye
 
 COPY . /go/src/github.com/pion/dtls
 WORKDIR /go/src/github.com/pion/dtls/e2e
--- a/e2e/e2e_openssl_test.go
+++ b/e2e/e2e_openssl_test.go
@@ -78,7 +78,7 @@ func serverOpenSSL(c *comm) {
 
 		// launch command
 		// #nosec G204
-		cmd := exec.CommandContext(c.ctx, "openssl", args...)
+		cmd := exec.Command("openssl", args...)
 		var inner net.Conn
 		inner, c.serverConn = net.Pipe()
 		cmd.Stdin = inner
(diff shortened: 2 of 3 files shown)


daenney (approved):
I haven't had time to dig into the upstream change yet, but this does work so lets :shipit:.

hasheddan (commented):
Thanks @daenney -- the primary issue being addressed here is that we cancelled context but didn't wait for the process to terminate, meaning that our `watchCtx` would hang [here](https://github.com/golang/go/blob/18e17e2cb12837ea2c8582ecdb0cc780f49a1aac/src/os/exec/exec.go#L782). I think there is a larger refactor that could be done to how we are passing context around here, but this offers a minimal diff to unblock moving forward 👍🏻

PR #556, opened 2023-07-06
Enable supported EC point formats and supported algorithm extensions

hasheddan:
Enables parsing the elliptic curve supported points format extension.

https://www.rfc-editor.org/rfc/rfc8422.html#section-5.1.2

Enables parsing the supported signature algorithms extension.

https://datatracker.ietf.org/doc/html/rfc5246#autoid-39

> Note: I noticed these while tracing and capturing packets in wireshark. They were present in the relevant `ClientHello` / `ServerHello`, but were not be extracted by their counterpart.

### DIFF
--- a/pkg/protocol/extension/supported_point_formats.go
+++ b/pkg/protocol/extension/supported_point_formats.go
@@ -44,12 +44,14 @@ func (s *SupportedPointFormats) Marshal() ([]byte, error) {
 func (s *SupportedPointFormats) Unmarshal(data []byte) error {
 	if len(data) <= supportedPointFormatsSize {
 		return errBufferTooSmall
-	} else if TypeValue(binary.BigEndian.Uint16(data)) != s.TypeValue() {
+	}
+
+	if TypeValue(binary.BigEndian.Uint16(data)) != s.TypeValue() {
 		return errInvalidExtensionType
 	}
 
-	pointFormatCount := int(binary.BigEndian.Uint16(data[4:]))
-	if supportedGroupsHeaderSize+(pointFormatCount) > len(data) {
+	pointFormatCount := int(data[4])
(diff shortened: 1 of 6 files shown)


hasheddan (commented):
Looks like our tests fail due to an error parsing the supported points format:

```
dtls internal: data length and declared length do not match
```

We are sending the supported points so it seems like we just have been avoiding this error due to not parsing them.

image

pkg/protocol/extension/supported_point_formats.go
```diff
@@ -44,12 +44,14 @@ func (s *SupportedPointFormats) Marshal() ([]byte, error) {
...
+	}
+
+	if TypeValue(binary.BigEndian.Uint16(data)) != s.TypeValue() {
 		return errInvalidExtensionType
 	}
 
-	pointFormatCount := int(binary.BigEndian.Uint16(data[4:]))
-	if supportedGroupsHeaderSize+(pointFormatCount) > len(data) {
+	pointFormatCount := int(data[4])
+	if supportedPointFormatsSize+pointFormatCount > len(data) {
```
hasheddan, line 54: This now adheres to the same way that we encode the supported EC point formats.

kevmo314 (approved):

daenney:
Could we add a test or two for this before we merge?

hasheddan:
@daenney sure thing -- this is being tested in the general case, which is why enabling these extensions started causing test failures. However, I think the unit tests could be improved... will do so ASAP.

hasheddan (commented):
@daenney added checks for unmarshal in the extensions that were missing them if you want to take another look :)

daenney (approved):
Nice, thanks for adding those.

PR #557, opened 2023-07-11
Fix return of nil alertErrors

hasheddan:
#### Description

Fixes two cases where nil alert errors were being returned rather than
the underlying error. Calling methods on these nil alert errors can lead
to panics.


Sean-Der:
Great catch, thank you @hasheddan :)

Sean-Der (approved):

PR #558, opened 2023-07-12
Implement DTLS 1.2 Connection IDs

### LINKED ISSUE #256: Plans for DTLS 1.2 Connection ID
Hi everyone, first of all congrats for the job you all are doing here. You got to a point where the product is really strong and reliable. 👏 

## Summary

* Support DTLS 1.2 Connection ID proposal: https://github.com/tlswg/dtls-conn-id
* RFC tracker: https://tools.ietf.org/html/draft-ietf-tls-dtls-connection-id

## Motivation

As we use DTLS 1.2 for our IoT fleet this would help using it on NATed environments.

It seems that DTLS 1.3 includes the CID also as @daenney pointed in #188.

### LINKED ISSUE #367: net.PacketConn
I'm a little surprised to find that a DTLS implementation, which is specifically datagram-oriented, seems to expect a stream-oriented `net.Conn` rather than a datagram-oriented `net.PacketConn` as an input to `Client()` and `Server()`.

Curious if there is a reason for this? It means that anything `net.PacketConn`-based has to be wrapped.

hasheddan:
#### Description

Implements DTLS 1.2 Connection ID support as defined in https://datatracker.ietf.org/doc/html/rfc9146.

Remaining work:
- [x] Clean up TODO's
- [x] Update `cbc` and `ccm` suites to support Connection IDs
- [x] Expand e2e tests with dedicated Connection ID tests
- [x] Expand unit tests to cover all new functionality
- [ ] Add examples for enabling Connection ID support (dependent on whether UDP connection package is implemented in this PR -- see below)
- [ ] Expand e2e tests with changing remote IP addresses (dependent on whether UDP connection package is implemented in this PR -- see below)
- [x] Clean up commit history (to be completed prior to merge)

##### Notes for early reviewers:

Currently, if you want an endpoint to be able to update its remote address, you need to implement a custom `Listener` that uses connection ID's rather than remote address to determine what `PacketServer` to route packets to. This functionality is dependent on `pion/transport/udp`, and I demonstrated what an example could look like in https://github.com/pion/transport/pull/252. However, as mentioned on that PR, the intention is to move the functionality to live here in `pion/dtls`, and I am in progress of doing so. We could feasibly make that a separate PR given that this one is "feature complete" in that it can successfully negotiate and use connection IDs as both a client and server (as demonstrated by the new e2e tests). Either way, the UDP connection package needs to be added in this PR or a subsequent one.

Another area of interest is automated testing against other libraries. The current OpenSSL tests cannot be expanded for this use-case as it does not support connection IDs. I have performed manual validation by expanding the DTLS client example in mbedTLS as demonstrated here: https://github.com/Mbed-TLS/mbedtls/compare/development...hasheddan:mbedtls:feat/cid-example. I would like to expand the e2e testing suite in this repository to include mbedTLS with connection IDs enabled, though we will need to build our own client, which could also be deferred to a subsequent PR.

An area that I _do not_ anticipate expanding scope of this PR to cover quite yet is saving connection state. For my particular use-case is is vital that we able to persist connection state across server restarts. Note that this is different from saving sessions as connection IDs must actually be renegotiated upon session resumption, whereas ideally with connection IDs enabled a server could continue receiving from an endpoint after restart by recovering all state. I _would_ like to ensure this functionality is enabled prior to cutting a `v3` release as I believe it could result in breaking changes.

Lastly, I am in progress of increasing unit test coverage across the board, though it is not anticipated that this will result in any functional changes to what is currently implemented as it is already being validated through existing unit tests (testing with connection IDs not enabled), as well as e2e test.

##### Testing Locally

If reviewers would like to experiment with enabling connection IDs locally, you can use the following commands to only run the `CID` e2e tests.

```
docker build -t pion-dtls-e2e -f e2e/Dockerfile .
```
```
docker run -i --rm --net=host pion-dtls-e2e go test -run=.*CID.* -v
```

Using the host's network namespace allows you to capture on the loopback interface in wireshark. Using the `dtls` filter will help eliminate other noise. Your output should look something like the following:

image

#### Reference issue

Fixes #256 
Fixes #367

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -4,6 +4,7 @@
 package dtls
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -21,8 +22,8 @@ import (
 	"github.com/pion/dtls/v2/pkg/protocol/handshake"
 	"github.com/pion/dtls/v2/pkg/protocol/recordlayer"
 	"github.com/pion/logging"
-	"github.com/pion/transport/v2/connctx"
 	"github.com/pion/transport/v2/deadline"
+	"github.com/pion/transport/v2/netctx"
 	"github.com/pion/transport/v2/replaydetector"
 )
 
@@ -45,21 +46,27 @@ func invalidKeyingLabels() map[string]bool {
 	}
 }
 
+type addrPkt struct {
+	rAddr net.Addr
+	data  []byte
+}
+
 // Conn represents a DTLS connection
 type Conn struct {
-	lock           sync.RWMutex     // Internal lock (must not be public)
-	nextConn       connctx.ConnCtx  // Embedded Conn, typically a udpconn we read/write from
-	fragmentBuffer *fragmentBuffer  // out-of-order and missing fragment handling
-	handshakeCache *handshakeCache  // caching of handshake messages for verifyData generation
-	decrypted      chan interface{} // Decrypted Application Data or error, pull by calling `Read`
-
-	state State // Internal state
+	lock           sync.RWMutex      // Internal lock (must not be public)
+	nextConn       netctx.PacketConn // Embedded Conn, typically a udpconn we read/write from
+	fragmentBuffer *fragmentBuffer   // out-of-order and missing fragment handling
+	handshakeCache *handshakeCache   // caching of handshake messages for verifyData generation
+	decrypted      chan interface{}  // Decrypted Application Data or error, pull by calling `Read`
+	rAddr          net.Addr
+	state          State // Internal state
 
 	maximumTransmissionUnit int
+	paddingLengthGenerator  func(uint) uint
 
 	handshakeCompletedSuccessfully atomic.Value
 
-	encryptedPackets [][]byte
+	encryptedPackets []addrPkt
 
 	connectionClosedByUser bool
 	closeLock              sync.Mutex
@@ -81,9 +88,8 @@ type Conn struct {
 	replayProtectionWindow uint
 }
 
-func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient bool, initialState *State) (*Conn, error) {
-	err := validateConfig(config)
-	if err != nil {
+func createConn(ctx context.Context, nextConn net.PacketConn, rAddr net.Addr, config *Config, isClient bool, initialState *State) (*Conn, error) {
+	if err := validateConfig(config); err != nil {
 		return nil, err
 	}
 
@@ -123,11 +129,18 @@ func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient
 		replayProtectionWindow = defaultReplayProtectionWindow
 	}
 
+	paddingLengthGenerator := config.PaddingLengthGenerator
+	if paddingLengthGenerator == nil {
+		paddingLengthGenerator = func(uint) uint { return 0 }
+	}
+
 	c := &Conn{
-		nextConn:                connctx.New(nextConn),
+		rAddr:                   rAddr,
+		nextConn:                netctx.NewPacketConn(nextConn),
 		fragmentBuffer:          newFragmentBuffer(),
 		handshakeCache:          newHandshakeCache(),
 		maximumTransmissionUnit: mtu,
+		paddingLengthGenerator:  paddingLengthGenerator,
 
 		decrypted: make(chan interface{}, 1),
 		log:       logger,
@@ -188,6 +201,7 @@ func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient
 		localGetCertificate:         config.GetCertificate,
 		localGetClientCertificate:   config.GetClientCertificate,
 		insecureSkipHelloVerify:     config.InsecureSkipVerifyHello,
+		connectionIDGenerator:       config.ConnectionIDGenerator,
 	}
 
 	// rfc5246#section-7.4.3
@@ -234,44 +248,49 @@ func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient
 // Dial connects to the given network address and establishes a DTLS connection on top.
 // Connection handshake will timeout using ConnectContextMaker in the Config.
 // If you want to specify the timeout duration, use DialWithContext() instead.
-func Dial(network string, raddr *net.UDPAddr, config *Config) (*Conn, error) {
+func Dial(network string, rAddr *net.UDPAddr, config *Config) (*Conn, error) {
 	ctx, cancel := config.connectContextMaker()
 	defer cancel()
 
-	return DialWithContext(ctx, network, raddr, config)
+	return DialWithContext(ctx, network, rAddr, config)
 }
 
 // Client establishes a DTLS connection over an existing connection.
 // Connection handshake will timeout using ConnectContextMaker in the Config.
 // If you want to specify the timeout duration, use ClientWithContext() instead.
-func Client(conn net.Conn, config *Config) (*Conn, error) {
+func Client(conn net.PacketConn, rAddr net.Addr, config *Config) (*Conn, error) {
 	ctx, cancel := config.connectContextMaker()
 	defer cancel()
 
-	return ClientWithContext(ctx, conn, config)
+	return ClientWithContext(ctx, conn, rAddr, config)
 }
 
 // Server listens for incoming DTLS connections.
 // Connection handshake will timeout using ConnectContextMaker in the Config.
 // If you want to specify the timeout duration, use ServerWithContext() instead.
-func Server(conn net.Conn, config *Config) (*Conn, error) {
+func Server(conn net.PacketConn, rAddr net.Addr, config *Config) (*Conn, error) {
 	ctx, cancel := config.connectContextMaker()
 	defer cancel()
 
-	return ServerWithContext(ctx, conn, config)
+	return ServerWithContext(ctx, conn, rAddr, config)
 }
 
-// DialWithContext connects to the given network address and establishes a DTLS connection on top.
-func DialWithContext(ctx context.Context, network string, raddr *net.UDPAddr, config *Config) (*Conn, error) {
-	pConn, err := net.DialUDP(network, nil, raddr)
+// DialWithContext connects to the given network address and establishes a DTLS
+// connection on top.
+func DialWithContext(ctx context.Context, network string, rAddr *net.UDPAddr, config *Config) (*Conn, error) {
+	// net.ListenUDP is used rather than net.DialUDP as the latter prevents the
+	// use of net.PacketConn.WriteTo.
+	// https://github.com/golang/go/blob/ce5e37ec21442c6eb13a43e68ca20129102ebac0/src/net/udpsock_posix.go#L115
+	pConn, err := net.ListenUDP(network, nil)
 	if err != nil {
 		return nil, err
 	}
-	return ClientWithContext(ctx, pConn, config)
+
+	return ClientWithContext(ctx, pConn, rAddr, config)
 }
 
 // ClientWithContext establishes a DTLS connection over an existing connection.
-func ClientWithContext(ctx context.Context, conn net.Conn, config *Config) (*Conn, error) {
+func ClientWithContext(ctx context.Context, conn net.PacketConn, rAddr net.Addr, config *Config) (*Conn, error) {
 	switch {
 	case config == nil:
 		return nil, errNoConfigProvided
@@ -279,16 +298,16 @@ func ClientWithContext(ctx context.Context, conn net.Conn, config *Config) (*Con
 		return nil, errPSKAndIdentityMustBeSetForClient
 	}
 
-	return createConn(ctx, conn, config, true, nil)
+	return createConn(ctx, conn, rAddr, config, true, nil)
 }
 
 // ServerWithContext listens for incoming DTLS connections.
-func ServerWithContext(ctx context.Context, conn net.Conn, config *Config) (*Conn, error) {
+func ServerWithContext(ctx context.Context, conn net.PacketConn, rAddr net.Addr, config *Config) (*Conn, error) {
 	if config == nil {
 		return nil, errNoConfigProvided
 	}
 
-	return createConn(ctx, conn, config, false, nil)
+	return createConn(ctx, conn, rAddr, config, false, nil)
 }
 
 // Read reads data from the connection.
@@ -352,6 +371,7 @@ func (c *Conn) Write(p []byte) (int, error) {
 					Data: p,
 				},
 			},
+			shouldWrapCID: len(c.state.remoteConnectionID) > 0,
 			shouldEncrypt: true,
 		},
 	})
@@ -400,7 +420,8 @@ func (c *Conn) writePackets(ctx context.Context, pkts []*packet) error {
 			c.log.Tracef("[handshake:%v] -> %s (epoch: %d, seq: %d)",
 				srvCliStr(c.state.isClient), h.Header.Type.String(),
 				p.record.Header.Epoch, h.Header.MessageSequence)
-			c.handshakeCache.push(handshakeRaw[recordlayer.HeaderSize:], p.record.Header.Epoch, h.Header.MessageSequence, h.Header.Type, c.state.isClient)
+
+			c.handshakeCache.push(handshakeRaw[recordlayer.FixedHeaderSize:], p.record.Header.Epoch, h.Header.MessageSequence, h.Header.Type, c.state.isClient)
 
 			rawHandshakePackets, err := c.processHandshakePacket(p, h)
 			if err != nil {
@@ -421,7 +442,7 @@ func (c *Conn) writePackets(ctx context.Context, pkts []*packet) error {
 	compactedRawPackets := c.compactRawPackets(rawPackets)
 
 	for _, compactedRawPackets := range compactedRawPackets {
-		if _, err := c.nextConn.WriteContext(ctx, compactedRawPackets); err != nil {
+		if _, err := c.nextConn.WriteToContext(ctx, compactedRawPackets, c.rAddr); err != nil {
 			return netError(err)
 		}
 	}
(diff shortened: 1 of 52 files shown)


pkg/protocol/recordlayer/fuzz_test.go
```diff
@@ -5,12 +5,14 @@ package recordlayer
 
 import (
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
```
Sean-Der, line 9: Possible to do this with testify/not take on another dependency?

Not a big deal, but users do appreciate keeping deps small :)

pkg/protocol/recordlayer/recordlayer.go
```diff
@@ -85,11 +82,40 @@ func UnpackDatagram(buf []byte) ([][]byte, error) {
...
+
+// ContentAwareUnpackDatagram is the same as UnpackDatagram but considers the
+// presence of a connection identifier if the record is of content type
+// tls12_cid.
+func ContentAwareUnpackDatagram(buf []byte, cidLength int) ([][]byte, error) {
+	out := [][]byte{}
+
+	for offset := 0; len(buf) != offset; {
+		headerSize := DefaultHeaderSize
+		lenIdx := 11
```
Sean-Der, line 109: Worth making `11` a constant and dropping a visualization of the header with it?

pkg/protocol/recordlayer/header.go
```diff
@@ -17,11 +17,14 @@ type Header struct {
...
 	SequenceNumber uint64 // uint48 in spec
+
+	// Optional Fields
+	ConnectionID []byte
 }
 
 // RecordLayer enums
 const (
-	HeaderSize        = 13
+	DefaultHeaderSize = 13
```
Sean-Der, line 27: Instead of `Default` it would be nice to hint at CID somehow.

`HeaderSizePreCID` might not be the right name.

hasheddan: @Sean-Der sounds good -- I was actually thinking of just keeping this as `HeaderSize` when I finish out this PR as this is a public `const` so any change is technically a breaking change. What do you think?

conn.go
```diff
@@ -270,6 +284,44 @@ func DialWithContext(ctx context.Context, network string, raddr *net.UDPAddr, co
 	return ClientWithContext(ctx, pConn, config)
 }
 
+type connPacket struct {
```
Sean-Der, line 287: mind moving into its own file?

As a reader it is to just ignore a bunch of stubs. As I move through a large file I find it overwhelming otherwise.

internal/ciphersuite/aes_ccm.go
```diff
@@ -103,11 +103,11 @@ func (c *AesCcm) Encrypt(pkt *recordlayer.RecordLayer, raw []byte) ([]byte, erro
 }
 
 // Decrypt decrypts a single TLS RecordLayer
-func (c *AesCcm) Decrypt(raw []byte) ([]byte, error) {
+func (c *AesCcm) Decrypt(h recordlayer.Header, raw []byte) ([]byte, error) {
```
hasheddan, line 106: If we don't want to break this API, we could define separate `DecryptWithHeader` methods.

pkg/protocol/recordlayer/header.go
Sean-Der: @hasheddan either works for me!

ConnectionID warrants a `V3`. It will be easier to get excitement/have people see the feature that way. So feel free to break the API in all the ways that make sense :)

I hope we can merge to master and tag `v3.0.0-alpha.1` very soon then!

internal/ciphersuite/aes_ccm.go
Sean-Der: I think it is fine breaking them!

Anyone using these APIs will understand/be able to adapt quickly I bet

daenney: Nobody can use these APIs outside of the DTLS library as it's in the `internal/` package. The only usages are within this library, so you can break it to your hearts' content.

hasheddan: @daenney good point -- I intended to comment on `pkg/crypto/ciphersuite`, which has similar changes.

pkg/protocol/recordlayer/header.go
hasheddan: Went with `FixedHeaderSize` as it feels like it could age a little better since it isn't directly coupled to the implementation of connection IDs (i.e. in theory there could be another extension in the future that modified header structure).

conn.go
```diff
@@ -261,6 +268,13 @@ func Server(conn net.Conn, config *Config) (*Conn, error) {
 	return ServerWithContext(ctx, conn, config)
 }
 
+// PacketServer listens for incoming DTLS connections.
+// Unlike Server, PacketServer allows for connections to change remote address.
+// The provided rAddr will be used as the initial remote address for sending.
+func PacketServer(ctx context.Context, conn net.PacketConn, rAddr net.Addr, config *Config) (*Conn, error) {
```
hasheddan, line 274: As an alternative to adding `PacketServer`, we could update the existing `Server` and `Client` to require passing a `net.PacketConn` rather than `net.Conn`. The reasoning for introducing `PacketServer` was to minimize breaking changes, but if we are cutting a `v3` one could make the case that requiring the user to wrap a `net.Conn` before passing it to the `Server` or `Client` would make it more obvious that remote address updates would not be supported. Said another way, if we are going to have breaking changes, it may be worth doing a larger refactor.

However, I also want to balance getting this functionality out to folks sooner rather than later, and I anticipate that we may have more breaking changes when we add support for DTLS 1.3.

daenney: Changing to `net.PacketConn` is something we've wanted to do for a long time, as it's confusing in many places for people to have a `net.Conn` in the context of UDP. It came up in #367 and we tagged it as a 3.0.0 milestone. So I'd be all in favour for changing that, as that's something we want to do one way or another.

hasheddan: @daenney thanks for this feedback! I'll push up an example of how the change could look to get your thoughts 👍🏻

conn.go
```diff
@@ -697,18 +780,66 @@ func (c *Conn) handleIncomingPacket(ctx context.Context, buf []byte, enqueue boo
...
+		// If connection ID does not match discard the packet.
+		if !bytes.Equal(c.state.localConnectionID, h.ConnectionID) {
+			c.log.Debug("unexpected connection ID")
+			return false, nil, nil
+		}
+
+		// Any valid connection ID record is a candidate for updating the remote
+		// address.
+		// https://datatracker.ietf.org/doc/html/rfc9146#peer-address-update
+		if rAddr != c.RemoteAddr() {
```
edaniels, line 838: Do we need to account for condition 2 of 3 for address update?:
```
The received datagram is "newer" (in terms of both epoch and sequence number) than the newest datagram received. Reordered datagrams that are sent prior to a change in a peer address might otherwise cause a valid address change to be reverted
```

edaniels (changes requested):
Done with a first pass. DTLS guts is very new to me but this looks great so far after my reading of RFC9146!

pkg/protocol/extension/connection_id.go
```diff
@@ -0,0 +1,58 @@
...
+	if TypeValue(extension) != c.TypeValue() {
+		return errInvalidExtensionType
+	}
+
+	var extData cryptobyte.String
+	val.ReadUint16LengthPrefixed(&extData)
+
+	var cid cryptobyte.String
+	if !extData.ReadUint8LengthPrefixed(&cid) {
+		return errInvalidSNIFormat
```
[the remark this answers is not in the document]
hasheddan: Ah yes -- thank you for catching!

internal/util/util.go
```diff
@@ -40,3 +42,10 @@ func Max(a, b int) int {
 	}
 	return b
 }
+
+// AddUint48 appends a big-endian, 48-bit value to the byte string.
+// Remove if / when https://github.com/golang/crypto/pull/265 is merged
+// upstream.
+func AddUint48(b *cryptobyte.Builder, v uint64) {
```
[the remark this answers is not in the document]
hasheddan: Agreed -- unit tests here and a number of other areas are coming 👍🏻

boaks:
Pretty cool! I like the docker stuff! Is it possible to start a server only with the docker image? And to try some handshakes with clients using other implementations?

daenney:
> Pretty cool! I like the docker stuff! Is it possible to start a server only with the docker image? And to try some handshakes with clients using other implementations?

Right now we only use Docker for the E2E CI tests, and they start a specific subset of tests using a build tag. However, you can look at `examples/listen` for how to start a server using this library to handle DTLS. Since it's Go the resulting binary can be copied into a tiny base image.

So, very crudely, it could be along the lines of:

```dockerfile
# Start by building the application.
FROM golang:1.20 as build

WORKDIR /go/src/app
COPY . .

RUN go mod download
RUN CGO_ENABLED=0 go build examples/listen/verify -o /go/bin/verify

# Now copy it into our base image.
FROM gcr.io/distroless/static-debian11
COPY --from=build /go/bin/verify /
CMD ["/verify"]
```

The only thing you'd really have to do is adapt the accept loop.

boaks:
> So, very crudely, it could be along the lines of:

That results on "my machine" in:

```
package examples/listen/verify is not in GOROOT (/usr/local/go/src/examples/listen/verify)
malformed import path "-o": leading dash
stat /go/bin/verify: directory not found
```

I'm no GO user, so I'll just wait for others results.

hasheddan:
@boaks apologies for the delay here! I have put together a test server and image that can be found at https://github.com/hasheddan/dtls-interop/tree/pion/dtls/cid/pion/cid_server. I will continue building this out with more functionality as I test against DTLS implementations. There are instructions for building and running the container image, and I have also pushed a preliminary version to DockerHub at [hasheddan/dtls-interop-pion-cid-server](https://hub.docker.com/repository/docker/hasheddan/dtls-interop-pion-cid-server/general).

The test server is setup to work out of the box with the example californium DTLS client. I tested by running the following and then executing the californium DTLS client.
```
docker run --rm --net host hasheddan/dtls-interop-pion-cid-server
```

The resulting packet capture can be found [here](https://github.com/hasheddan/dtls-interop/tree/pion/dtls/cid/tests/1_2-cid). Please let me know if there is anything else I can do to help assist in your testing!

boaks:
Very nice!
I've tested it with an mbedtls client successfully.

boaks:
> The resulting packet capture can be found [here](https://github.com/hasheddan/dtls-interop/tree/pion/dtls/cid/tests/1_2-cid). 

Just to mention, as in my WikiPage for Captures: if the capture is compressed, it can be uploaded to a git comment just by drag&drop.

> Please let me know if there is anything else I can do to help assist in your testing!

Thanks. For now, I only test the basic interoperability. That looks pretty good.

conn.go
```diff
@@ -499,22 +551,49 @@ func (c *Conn) processHandshakePacket(p *packet, h *handshake.Handshake) ([][]by
...
-			ContentLen:     uint16(len(handshakeFragment)),
-			Epoch:          p.record.Header.Epoch,
-			SequenceNumber: seq,
-		}
+		var rawPacket []byte
+		if p.shouldWrapCID {
+			inner := &recordlayer.InnerPlaintext{
+				Content:  handshakeFragment,
+				RealType: protocol.ContentTypeHandshake,
+				Zeros:    8,
```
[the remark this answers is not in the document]
hasheddan: Agreed, though I really should have a TODO here. Using a constant padding size doesn't really provide much value here. There are a variety of both deterministic and non-deterministic padding strategies that can be used, so I was considering allowing for a padding generator to be passed in config similar to the connection ID generator. If none is passed, the default would just be zero-length padding.

For some examples of how other (D)TLS libraries handle this:
- mbedTLS has an [`ssl_compute_padding_length`](https://github.com/Mbed-TLS/mbedtls/blob/17526a0d168276aa3ba5833cbb1f8fcd69d688ba/library/ssl_msg.c#L313) function that takes the content length and a granularity argument. The granularity is configured via the [`MBEDTLS_SSL_CID_TLS1_3_PADDING_GRANULARITY`](https://github.com/Mbed-TLS/mbedtls/blob/17526a0d168276aa3ba5833cbb1f8fcd69d688ba/include/mbedtls/mbedtls_config.h#L3825) option.
- scandium (californium) allows passing a `pad` when [generating a `Record`](https://github.com/eclipse-californium/californium/blob/83dd5fe528237b715dbc31a38a9cd8726ccfc248/scandium-core/src/main/java/org/eclipse/californium/scandium/dtls/Record.java#L215). However, it looks like it is [only ever set to `0` currently](https://github.com/eclipse-californium/californium/blob/83dd5fe528237b715dbc31a38a9cd8726ccfc248/scandium-core/src/main/java/org/eclipse/californium/scandium/DTLSConnector.java#L280).

Given these as a small sample, I think the generator strategy which would default to zero-length padding seems like a reasonable solution for initial implementation, while also allowing for flexibility.

/cc @boaks @Sean-Der if either of you want to weigh-in :)

flight1handler.go
```diff
@@ -118,6 +118,14 @@ func flight1Generate(c flightConn, state *State, _ *handshakeCache, cfg *handsha
 		}
 	}
 
+	// If we have a connection ID generator, use it. The CID may be zero length,
+	// in which case we are just requesting that the server send us a CID to
+	// use.
+	if cfg.connectionIDGenerator != nil {
```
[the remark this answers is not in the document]
edaniels: I like the generator idea!

conn.go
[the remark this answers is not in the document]
edaniels: zero length seems fine since yeah, a constant doesn't really add much value when it's known

conn.go
hasheddan: Checking a few things here, but will follow up in a moment :)

flight1handler.go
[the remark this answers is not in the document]
hasheddan: Added `OnlySendCIDGenerator()` and corresponding tests 👍🏻

conn.go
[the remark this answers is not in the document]
hasheddan: Added a `PaddingLengthGenerator` config that defaults to `0` 👍🏻

conn.go
hasheddan: Following up: from my understanding, we don't currently do any reordering based on sequence number, only on epoch. The latest sequence number is [updated in the replay detector](https://github.com/pion/transport/blob/41845b068b67baa290c06efdadc59b5c5300c8ba/replaydetector/replaydetector.go#L50) when we call `markPacketAsValid()`. The [anti-replay strategy defined in the DTLS 1.2 RFC](https://datatracker.ietf.org/doc/html/rfc6347#autoid-21) is used.

I think a reasonable solution without too much churn would be to return from the callback (i.e. `markPacketAsValid()`) a `bool` that indicates whether marking the packet as valid resulted in a latest sequence number update. We could then use that value to determine whether we need to update remote address or not.

cc @Sean-Der @daenney

edaniels: I think that's a perfectly acceptable solution. Honestly if there's a unit test backing it, that's more important to me right now!

hasheddan: @edaniels thanks! PR opened here with implementation https://github.com/pion/transport/pull/256

edaniels: nice. the change from there looks good here

conn.go
```diff
@@ -234,61 +248,66 @@ func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient
...
 
-	return DialWithContext(ctx, network, raddr, config)
+	return DialWithContext(ctx, network, rAddr, config)
 }
 
 // Client establishes a DTLS connection over an existing connection.
 // Connection handshake will timeout using ConnectContextMaker in the Config.
 // If you want to specify the timeout duration, use ClientWithContext() instead.
-func Client(conn net.Conn, config *Config) (*Conn, error) {
+func Client(conn net.PacketConn, rAddr net.Addr, config *Config) (*Conn, error) {
```
hasheddan, line 261: Note: all client / server constructors are updated to accept `net.PacketConn`, rather than accepting `net.Conn` as we are moving towards using `net.PacketConn`. However, fully switching over to `net.PacketConn` involves refactoring / porting a fair amount of `pion/transport` (i.e. `udp.Listener`, `packetio.Buffer`, etc.) to live in their new forms in `pion/dtls`. The current state of this PR allows for any consumers to use all of the client / server constructors directly to exercise existing and new functionality (i.e. connection IDs). The follow-up work is mostly around updating the `dtls.Listener` (see https://github.com/pion/transport/pull/252 for the general direction of what those changes will look like). However, given that those changes will be fairly large and will include their own rounds of review, which is fairly unrelated to the connection ID work in this PR, I would propose that we work to get this PR reviewed and merged, then I can follow with updates to support the new listener behavior. In the interim, folks will be able to exercise the connection ID functionality _and_ the previous listener behavior will continue to work as expected.

@Sean-Der @edaniels @daenney do y'all have any thoughts in favor or against this proposal? I have the work in progress to port the aforementioned packages, but I also am cognizant that introducing them will likely 3x the size of this PR, which could make reviewing even more cumbersome.

edaniels: I think sticking with this new Client/Server + FromConn is preferred. It also lets us get other packages updated with the breaking API change and allows it to bake for a bit.

pkg/protocol/extension/connection_id_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+func TestExtensionConnectionID(t *testing.T) {
+	rawExtensionConnectionID := []byte{1, 6, 8, 3, 88, 12, 2, 47}
+	parsedExtensionConnectionID := &ConnectionID{
+		CID: rawExtensionConnectionID,
+	}
+
+	raw, err := parsedExtensionConnectionID.Marshal()
+	if err != nil {
+		t.Error(err)
```
edaniels, line 19: are these errors better as Fatal so as to stop the current test immediately?

pkg/crypto/ciphersuite/ciphersuite_test.go
```diff
@@ -0,0 +1,55 @@
...
+			reason: "Should successfully generate additional data with valid header",
+			header: &recordlayer.Header{
+				ContentType:    protocol.ContentTypeConnectionID,
+				ConnectionID:   []byte{1, 2, 3, 4, 5, 6, 7, 8},
+				Version:        protocol.Version1_2,
+				Epoch:          2,
+				SequenceNumber: 277,
+			},
+			payloadLen: 1784,
+			expected:   []byte{255, 255, 255, 255, 255, 255, 255, 255, 25, 8, 25, 254, 253, 0, 2, 0, 0, 0, 0, 1, 21, 1, 2, 3, 4, 5, 6, 7, 8, 6, 248},
```
edaniels, line 32: would it be reasonable to prove that these are the same by not just checking the bytes but also unmarshaling them and checking the structure? I only say this because it could help prove that the bytes aren't just copied over from test output that could be wrong while an unmarshal proves with little doubt that the bytes are good.

edaniels (approved):
This LGTM. I have two test suggestions but they do not block merging the code in. Great work!

pkg/crypto/ciphersuite/ciphersuite_test.go
edaniels: Gotcha. Good point

PR #571, opened 2023-08-22
Convert nil CIDs to empty byte slice

hasheddan:
#### Description

Adds a check on the generated CID in flight 1 to convert to an empty
byte slice if the CID is nil. This allows us to distinguish not
supporting CIDs to only suppporting sending CIDs in the second
ClientHello in flight 3.

While the existing functionality actually results in a successful handshake with the server (if that server is `pion/dtls`), it has the undesirable effect of sending a different set of extensions on the first client hello and the second. This is due to the fact that we decide to send the CID on the second if we sent it on the first, but because we were using a `nil` CID when only indicating support, that resulted in us dropping the CID on the second client hello.

https://github.com/pion/dtls/blob/83b125485ec0277011ba97d26c2c5578ea96d8a7/flight3handler.go#L283-L285

Example from previous functionality:
Screenshot from 2023-08-22 17-06-17

After the changes in this patchset:
Screenshot from 2023-08-22 17-05-24

### DIFF
--- a/flight1handler.go
+++ b/flight1handler.go
@@ -123,6 +123,13 @@ func flight1Generate(c flightConn, state *State, _ *handshakeCache, cfg *handsha
 	// use.
 	if cfg.connectionIDGenerator != nil {
 		state.localConnectionID = cfg.connectionIDGenerator()
+		// The presence of a generator indicates support for connection IDs. We
+		// use the presence of a non-nil local CID in flight 3 to determine
+		// whether we send a CID in the second ClientHello, so we convert any
+		// nil CID returned by a generator to []byte{}.
+		if state.localConnectionID == nil {
+			state.localConnectionID = []byte{}
+		}
 		extensions = append(extensions, &extension.ConnectionID{CID: state.localConnectionID})
 	}
 


connection_id.go
```diff
@@ -23,6 +23,6 @@ func RandomCIDGenerator(size int) func() []byte {
 // necessary.
 func OnlySendCIDGenerator() func() []byte {
```
edaniels, line 24: Instead of this, I think it's better to be ultra explicit and use a second return value of a boolean to indicate, there is a CID. Do you think it makes sense to do something like this now or just make the fix and consider it later?

hasheddan: Hmm, I don't love the idea of expanding the signature just for a single scenario when we are allowing folks to supply their own generators. We say that the presence of a connection ID generator enables CID support, which would then only be true if the generator returned `true`. A potential alternative would be to evaluate whether the generated CID is `nil` in flight 1, then set the local connection ID to `[]byte{}` instead to ensure we properly set it in the second client hello. That would make returning `nil` or `[]byte{}` from a generator have the same effect, which was the original intention.
https://github.com/pion/dtls/blob/83b125485ec0277011ba97d26c2c5578ea96d8a7/flight1handler.go#L124-127

Thoughts?

edaniels: I like that idea!

edaniels: Otherwise there's more onus on other impls of this to correctly distinguish from nil/[]byte{} which is uncommon from what I've seen with slice usage

hasheddan: Awesome! Update incoming 👍🏻

edaniels (approved):
LGTM

PR #573, opened 2023-08-28
Include connection IDs in serialized state

hasheddan:
#### Description

Updates state serialization and deserialization to include connection
IDs if they were negotiated.

Updates the flight0handler to clear connection IDs in the event the
client is attempting to resume an existing session. If connection IDs
are to be used in the resumed session, they should be negotiated afresh.

https://datatracker.ietf.org/doc/html/rfc9146#name-the-connection_id-extension


hasheddan (commented):
Example of using this functionality demonstrated in https://github.com/plgd-dev/go-coap/pull/473

edaniels (approved):

PR #577, opened 2023-09-01
Correctly identify client and server with PSK ID

hasheddan:
#### Description

Updates the PSK ID hints on the client and server examples to match the type of endpoint they represent.


daenney (approved):
Hah, looks like we've had those the wrong way around for ages. Thanks for cleaning that up.

PR #582, opened 2023-09-05
Use atomic.Value to maintain Go 1.13 compatibility 

hasheddan:
#### Description

Updates to use atomic.Value in place of atomic.Bool such that older Go
versions can continue to be supported.

Similar fix to https://github.com/pion/transport/commit/701ff64e8ae945808b3e82c783e37d514a731c44

### DIFF
--- a/internal/net/udp/packet_conn.go
+++ b/internal/net/udp/packet_conn.go
@@ -101,7 +101,7 @@ func (l *listener) Close() error {
 				}
 				// If we haven't already removed the remote address, remove it
 				// from the connection map.
-				if !c.rmraddr.Load() {
+				if c.rmraddr.Load() == nil {
 					delete(l.conns, c.raddr.String())
 					c.rmraddr.Store(true)
 				}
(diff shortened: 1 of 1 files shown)


daenney:
Honestly, I'd rather we don't and update the minimum version in `go.mod` instead.

Go 1.13 is from September 2019. It's now 4 years later. Go is very forwards compatible so the burden of upgrading a version once a while, especially once every 4 years, is basically 0. Go 1.19, which is what we'd need for `atomic.Bool`, is already a year old and with the release of 1.21 effectively no longer supported by the Go team. I think it's fine to move our baseline to it. That would also enable the use of generics if we want it at some point.

hasheddan (commented):
> Go 1.13 is from September 2019. It's now 4 years later. Go is very forwards compatible so the burden of upgrading a version once a while, especially once every 4 years, is basically 0. Go 1.19, which is what we'd need for atomic.Bool, is already a year old and with the release of 1.21 already no longer supported by the Go team. I think it's fine to move our baseline to it. That would also enable the use of generics if we want it at some point.

@daenney I tend to agree, but I believe there is a desire for other `pion` libraries to continue to support older Go versions, and other consumers are having to workaround this to maintain their older version support as well (e.g. https://github.com/plgd-dev/go-coap/pull/481). Perhaps it makes more sense to maintain the compatibility for now, then start a discussion about what our minimum Go version requirement policy will be going forward?

daenney (approved):

at-wat:
- related discussion: https://github.com/pion/webrtc/issues/2292

PR #585, opened 2023-09-22
Export CipherSuite in connection State

### LINKED ISSUE #584: Please export some connection state variables
## Motivation
I need to obtain the cipher suite negotiated for the connection in an application, Is it possible to export some [variables](https://github.com/pion/dtls/blob/744e27a38f208e465de597d9da1644b66224ffc9/state.go#L18), at least including CipherSuite, like https://pkg.go.dev/crypto/tls#ConnectionState

itviewer:
#### Description
This commit exports the CipherSuite member variable in dtls connection state, which our application currently requires
#### Reference issue
Fixes #584

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -17,7 +17,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
       - name: test
         run: |
           docker build -t pion-dtls-e2e -f e2e/Dockerfile .


daenney:
This feels a little dangerous to me, as it now becomes possible to accidentally assign something to the cipher suite state by making it exported.

State currently already has a public `NegotiatedProtocol`. I would suggest we mirror that, and add a `NegotiatedCipherSuite` that gets written once we've completed the handshake. That avoids the internal state ever becoming modifiable through the public API.

itviewer:
Should the API consider aligning with https://pkg.go.dev/crypto/tls#ConnectionState as closely as possible?

itviewer:
> Once the ID is public, the name can always be retrieved by running the ID through `CipherSuiteName()` which is already public, if anyone has a need for that. Since we have a specific type for this, the CipherSuiteID, I would suggest exposing that and not the underlying `uint16` that's actually `ciphersuite.ID`. That makes it a little different than crypto/tls, but it makes it easier to work with in the context of this library and if you really need the `uint16` it's trivial to get.

@daenney Thanks for your advice, Is this okay? https://github.com/pion/dtls/pull/586

PR #586, opened 2023-09-22
Export CipherSuiteID in connection State

### LINKED ISSUE #584: Please export some connection state variables

itviewer:
#### Description
This commit exports the `CipherSuiteID`  in connection state for further use, e.g. to get the name of the CipherSuite via `CipherSuiteName()`
#### Reference issue
Fixes #584

### DIFF
--- a/state.go
+++ b/state.go
@@ -21,6 +21,7 @@ type State struct {
 	localRandom, remoteRandom handshake.Random
 	masterSecret              []byte
 	cipherSuite               CipherSuite // nil if a cipherSuite hasn't been chosen
+	CipherSuiteID             CipherSuiteID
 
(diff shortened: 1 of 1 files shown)


daenney:
This looks fine by me. Any thoughts @Sean-Der? I'll leave it up for a few days so other folks have a chance to see and comment it. If nothing happens I'll merge it in a few days.

hasheddan (approved):
+1 from me 👍🏻

daenney:
@itviewer Thanks for the change!

PR #588, opened 2023-09-30
flight3: respect curves configuration

### LINKED ISSUE #587: Flight3 does not respect Config.EllipticCurves
### Your environment.
  - Version: v2.2.7
  - Browser: none

### What did you do?

I specified custom EllipticCurves list in dtls.Config

### What did you expect?

Modified list of curves in the *supported_groups* extension.

### What happened?

Flight1 packet is sent as expected, but Flight3 packet contains all curves regardless configuration. Problem does not manifests itself if server Config.InsecureSkipVerifyHello is set to true.

### Additional information

I think responsible part of code is https://github.com/pion/dtls/blob/744e27a38f208e465de597d9da1644b66224ffc9/flight3handler.go#L251-L260

I'm also willing to try to put together a fix for this if we can discuss it and get some clarity about what to do with it.

Snawoot:
#### Description

As far as I understand Flight3 is pretty much like Flight1, but with Cookie. So I just reuse logic from Flight1 to ensure correct behaviour.

#### Reference issue
Fixes #587

### DIFF
--- a/flight3handler.go
+++ b/flight3handler.go
@@ -249,10 +249,11 @@ func flight3Generate(_ flightConn, state *State, _ *handshakeCache, cfg *handsha
 			RenegotiatedConnection: 0,
 		},
 	}
+
 	if state.namedCurve != 0 {
 		extensions = append(extensions, []extension.Extension{
 			&extension.SupportedEllipticCurves{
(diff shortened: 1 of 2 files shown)


at-wat:
Seems related to https://github.com/pion/dtls/pull/475 (@daenney FYI)

@Snawoot Could you add a test as well as flight1?

Sean-Der:
I added a test, after CI passes will merge!

PR #594, opened 2023-10-17
Fix for issue 418 with unit test

mschexnaydre:
For RFC compliance only algorithms in the certificate request shall be used.

Description
We should only be using one of the signature algorithms specified in the CertificateRequest message when generating the CertificateVerify message. Prior to this fix SHA-256 was always being used.

This change stores the HASH algorithm from the CertificateRequest message in the State object so that we can reference these later when generating the CertificateVerify message.

Removed hard-coded usage of SHA-256 in generateCertificateVerify, now uses the Digest method of the passed in algorithm.

Added unit test in conn_test that specifies SignatureSchemes with a non SHA256 only option.  This causes the unit test to fail prior to this fix.

Reference issue
Fixes https://github.com/pion/dtls/issues/418

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -38,6 +38,7 @@ Lukas Lihotzki <lukas@lihotzki.de>
 ManuelBk <26275612+ManuelBk@users.noreply.github.com>
 Michael Zabka <zabka.michael@gmail.com>
 Michiel De Backker <mail@backkem.me>
+mschexnaydre <mschex@viasat.com>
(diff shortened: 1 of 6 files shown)


mschexnaydre:
@Sean-Der @hasheddan If you guys could review and merge when you get a chance - this is a nasty bug that prevents connections to devices that do not support SHA-256.  I have added the unit test and would love to see this put back into the main branch.  Thanks!

Sean-Der (approved):

PR #595, opened 2023-11-15
Use atomic to avoid stale SRTP protection profile

boks1971:
`state` is acccessed without lock in the FSM.
In some cases, that leads to stale values.
For example, `srtpProtectionProfile` is set in flight handlers (differnt flight handlers in client and server). But, when it is accessed via the API
`SelectedSRTPProtectionProfile`,
it gets a stale value as it appears that the two goroutines are out-of-sync on that piece of shared memory.

This is a larger concern for use of `state`.
Ideally, either
- `state` should have a lock internally and all fields are accessed through methods.
- carefully split fields of `state` to ensure process access/sync.

Doing the smaller change here to address one field that has seen stale value.


cnderrauber (approved):
Greate finding!!

PR #596, opened 2023-11-22
Proposal: Include Client Address in PSK Validation for Brute Force Detection

tonisole:
#### Description

I am proposing a modification to the current PSK validation mechanism to include the client's address as part of the validation process. This change is crucial for implementing a Brute Force Detection mechanism in our system.

Currently, the PSK validation does not provide any information about the client attempting to connect. This lack of information makes it impossible to detect if a specific IP address is repeatedly trying to discover the Hint/PSK authorization, a common sign of a brute force attack.

```go
var attempts = make(map[string]int) // Map of attempts for each IP address for a Brute Force Protection

PSK: func(hint []byte, addr net.Addr) ([]byte, error) {
	fmt.Printf("Server's hint: %s \n", hint)
	// *************** Brute Force Attack protection ***************
	// Check if the IP address is in the map, and the IP address has exceeded the limit
	if attempts[addr.(*net.UDPAddr).IP.String()] > 5 {
		return nil, fmt.Errorf("too many attempts from this IP address")
	}
	// Here I increment the number of attempts for this IP address
	attempts[addr.(*net.UDPAddr).IP.String()]++
```

I am using this library in conjunction with the go-coap library, and having the ability to implement this detection mechanism would significantly enhance the security of our system. By including the client's address in the PSK validation, we can monitor and limit the number of attempts from a single IP address, effectively preventing potential brute force attacks.

This draft is intended to open a discussion about the necessity and feasibility of this proposed change. I believe that with this modification, we can make our system more secure and robust against potential threats. I look forward to hearing your thoughts and suggestions on this matter.

#### Reference issue

No related issue


at-wat (commented):
Looks really nice to have this feature, however, it requires major version bump as it's a breaking change.
How about keeping `PSK: func([]byte) ([]byte, error)` and adding `PSKWithAddr: func([]byte, net.Addr) ([]byte, error)`?

Sean-Der:
@at-wat @tonisole We do need to tag a `/v3`!

@hasheddan can I get a short write up on your changes? I say we do this change proposed here, and yours? I really would like to get pion/dtls stabilized soon :)

hasheddan (commented):
I'm on board with these changes! @Sean-Der sending over write-up now :)

Sean-Der:
Hi @tonisole 

Sorry I didn't make more movement on this! I would love to see this get into pion/dtls

What do you think of making it non-PSK specific? I think this would be useful for certificates also?

PR #604, opened 2024-01-02
Drop invalid record silently during handshake

cnderrauber:
Fix issue: invalid record in handshake staging cause readloop exited then handshake failed.


boks1971 (approved):
LGTM! Great job chasing this one @cnderrauber !!!

PR #613, opened 2024-03-18
address #279: allow for parallel processing of handshakes

### LINKED ISSUE #279: Listener doesn't process parallel handshakes
### Your environment.
https://github.com/go-ocf/go-coap/issues/109

### What did you do?
Just look at last comment. But it seems that other handshakes waits each other.

### What did you expect?
dlts.Listener is able to process parallel handshakes.

jordan-bonecutter:
#### Description
To allow for parallel handshake processing, I propose to introduce the `AcceptHandshake` method on `listener`. `AcceptHandshake` returns a `Handshaker` interface, which is just a conn who has not yet performed their handshake and a config for doing so. Handshakes may be processed in parallel by calling the `Handshake` method on the `Handshaker` in a background goroutine.

#### Reference issue
Fixes #279

### DIFF
--- a/examples/listen/handshake_psk/main.go
+++ b/examples/listen/handshake_psk/main.go
@@ -1,0 +1,85 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+// Package main implements a DTLS server using a pre-shared key.
+package main
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/pion/dtls/v2"
+	"github.com/pion/dtls/v2/examples/util"
+)
+
+func main() {
+	// Prepare the IP to connect to
+	addr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 4444}
+
+	// Create parent context to cleanup handshaking connections on exit.
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	//
+	// Everything below is the pion-DTLS API! Thanks for using it ❤️.
+	//
+
+	// Prepare the configuration of the DTLS connection
+	config := &dtls.Config{
+		PSK: func(hint []byte) ([]byte, error) {
+			fmt.Printf("Client's hint: %s \n", hint)
+			return []byte{0xAB, 0xC1, 0x23}, nil
+		},
+		PSKIdentityHint:      []byte("Pion DTLS Server"),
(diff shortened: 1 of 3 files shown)


daenney:
Thanks for the proposed change. I'm not sure about the Handshake interface, but either way we'd need to at least also have tests, and ideally an example that shows/documents how to use it.

jordan-bonecutter:
> Thanks for the proposed change. I'm not sure about the Handshake interface, but either way we'd need to at least also have tests, and ideally an example that shows/documents how to use it.

Definitely, didn't want to go too far down the rabbit hole if people didn't like the change. I'll spend some time making a test today. I can make an example as well.

The basic idea (if unclear) is that before one would do:


```go
conn, err := ln.Accept()
if err != nil {
  return err
}
go handleConn(conn)
```

Now, one may do:

```go
hs, err := ln.AcceptHandshake()
if err != nil {
  return err
}
go handleHandshaker(hs)
```

The only annoying thing is that the `AcceptHandshake` method won't be exported, so it'll require a type assertion unless a different type with this method exported is returned.

jordan-bonecutter:
@daenney I've added a test + example. Let me know what needs changing, thanks!

daenney:
@Sean-Der Care to take a look? This has come up before and it might just be nice to fix it. It's a backwards compatible change since folks can upgrade to the proposed Handshake method of doing it. I know we've discussed something similar to this at some point for a v3 too.

jordan-bonecutter:
@daenney bump on this. Anything I can do to speed up the process?

daenney:
No not really, things take time as we're all unpaid volunteers. A few other maintainers need to take a look, as adding new APIs has long-term-ish consequences.

Sean-Der:
Fixed by https://github.com/pion/dtls/commit/e4064683d277f946e5d1a4b359ca4e552c23131c

pion/dtls now matches the stdlib. The handshake is performed on the Read/Write.

We no longer block the Accept with the handshake!

Sean-Der:
I am sorry I didn't respond to this sooner @jordan-bonecutter I have just renewed the effort on pion/dtls in the last weeks :/ 

This is a shitty open source experience, I apologize for that. Totally my fault

jordan-bonecutter:
> I am sorry I didn't respond to this sooner @jordan-bonecutter I have just renewed the effort on pion/dtls in the last weeks :/ 
> 
> 
> 
> This is a shitty open source experience, I apologize for that. Totally my fault

No worries. Thanks for creating some awesome libraries, looking forward to contributing to some of them eventually!

PR #622, opened 2024-04-02
Update go.mod to 1.19

edaniels:


daenney:
I don't think we can do this. Other things like pion/webrtc import this module, and they're still on 1.19. The dtls library would have to be one of the last ones to bump the Go version dependency.

daenney:
That said I am in favour of starting to bump the minimum version across Pion. Nobody should be depending on Go 1.13 now. Would be nice to get to at least 1.18 so we can start making some use of generics. 

Given 1.22 is current, I wouldn't mind bumping to 1.20 baseline overall either, as that also brings `errors.Join` in to the party which can be handy and a bunch more types satisfying Comparable which is nice when using generics.

edaniels:
Hey @daenney, we're upgrading to 1.19 across the board per https://github.com/pion/webrtc/issues/2292

Sean-Der (approved):

Sean-Der:
Thanks for pushing back/asking questions @daenney :) lucky to have you around!

A few weeks ago we had a get together on Slack. The consensus was 1.19 was a good first step. I don't want to alienate users who get Go via pkg manager (Debian mostly)

I really would like to use more modern features/tooling though!

daenney:
Ah, perfect! Nice to see the upgrade.

PR #627, opened 2024-04-17
Fix typing for alertErrors

sukunrt:


sukunrt:
I think there's some problem with CI

Sean-Der:
@sukunrt can you open this against the master branch? 

The linter configs are global, and I haven't been updating the `v2` branch with new lints as they come up.

sorry about that!

conn.go
```diff
@@ -643,13 +642,12 @@ func (c *Conn) handleQueuedPackets(ctx context.Context) error {
...
 		var e *alertError
-		if errors.As(err, &e) {
-			if e.IsFatalOrCloseNotify() {
-				return e
-			}
-		} else if err != nil {
+		if errors.As(err, &e) && e.IsFatalOrCloseNotify() {
 			return e
 		}
+		if err != nil {
```
daenney, line 648: I'm not entirely sure this makes sense? We're first checking if it's a particular error with a particular condition and then `return e`, but that's still of type `error`. Right after, we're doing `if err != nil { return err }`. It seems at that point we don't need the check for `if errors.As` at all, since we always return `err` if it's not nil.

MarcoPolo:
> @sukunrt can you open this against the master branch?

I believe the intent here is to patch the v2 release to allow a patch release with this change. This allows users to get this fix with no changes on their end besides updating dtls by a patch version.

My understanding is that master doesn't have this issue, but cutting a release from master is a bit more involved and requires changes in pion/webrtc as well in order for users to get the fix.

conn.go
sukunrt: I think you're right. I copied this from master. 

I think all we need is `if err != nil {return err}`

daenney: Thanks for fixing that. I think you might've forgotten to push the right branch? I don't see the change and CI hasn't picked up on it.

sukunrt: done.

sukunrt:
@daenney, I've updated the PR now. I made one change after this comment. https://github.com/pion/dtls/pull/627#discussion_r1569251804

conn.go
sukunrt: Adding the `else if err != nil` in place of `if err != nil` to keep the current behavior as is.

MarcoPolo (approved):

conn.go
```diff
@@ -648,7 +643,7 @@ func (c *Conn) handleQueuedPackets(ctx context.Context) error {
 				return e
 			}
 		} else if err != nil {
-			return e
+			return err
```
daenney, line 646: I think this is the same situation as on L610, the additional checking for `alertError` doesn't do anything. But if this exists on master still then lets fix that there instead.

daenney (approved):

conn.go
sukunrt: I think so. I didn't want to change any behavior so went for the minimal change. 

On line 610 the caller checks whether alertError is fatal so we dont need the check here and just return err if it isn't nil. 

The caller for line 651 doesn't check if the alertError is fatal so we check it here and suppress alertErrors that are not fatal.

PR #628, opened 2024-04-17
limit size of encrypted packet queue

sukunrt:


conn.go
```diff
@@ -149,6 +137,38 @@ func createConn(ctx context.Context, nextConn net.Conn, config *Config, isClient
...
+
+	mtu := config.MTU
+	if mtu <= 0 {
+		mtu = defaultMTU
+	}
+
+	replayProtectionWindow := config.ReplayProtectionWindow
+	if replayProtectionWindow <= 0 {
+		replayProtectionWindow = defaultReplayProtectionWindow
+	}
```
MarcoPolo, line 171: Are these used in this func? Or is this a remnant of a copy/paste?

MarcoPolo (changes requested):
This looks good, just a small ask first.

conn.go
sukunrt: done. These were a remnant of copy/paste.

MarcoPolo (approved):

PR #631, opened 2024-04-19
Add handshake hooking capabilities

theodorsm:
#### Description

This PR is part of a larger project adding anti-fingerprint features to make this library censorship-resistant, as it is used by Snowflake. The main features of this PR are the following:

- <s>Added a workflow for automatically generating fresh DTLS handshakes (fingerprints) of new browser versions (Firefox and Chrome) by using a minimal webrtc example application and Selenium. The handshakes are captured and stored as pcap artifacts. The pcaps are further parsed and a fingerprint is added to `pkg/mimicry/fingerprints.go`.</s>
- <s>Mimick client hello messages by replaying fingerprints. This is enabled by a config flag and does not make any other API changes for users. Specific fingerprints can be chosen by a config flag or if it's empty, the most recent fingerprint will be chosen.</s>
- Hooking of the client hello, server hello and certificate request messages.

*Note*:
- Enabling hooking might be unstable as a client/server might announce a cipher it actually does not support or because the internal state is not changed.
- <s>The fingerprints are collected from a webrtc use case, the `use_srtp` extension is added to the handshake.</s>
 
#### Reference issue
https://gitlab.torproject.org/tpo/anti-censorship/pluggable-transports/snowflake/-/issues/40014
 
 #### Todos
 
 - [x] Documentation
 - [x] Examples
 - [x] <s> Decide how often the fingerprint generation workflow should be run. </s>
 - [x] <s> Decide if pcaps should be committed to this repo or not. </s>


daenney:
I'm personally a fairly strong no on this change.

I don't mind taking on changes to make this library configurable enough that it can be used successfully to mimic a particular handshake. But taking on the burden of providing censorship resistance, keeping the fingerprints up to date and potentially dealing with any issues from it doesn't strike me as the right thing for the Pion community to spend its innovation tokens/budget on.

I'm curious to here how other maintainers feel about this.

theodorsm:
Thanks for your input @daenney. I have talked with @Sean-Der about this, and he seems positive about the idea, but I do see your point in spending the innovation budget on this. 

I am doing this as part of my master thesis on "Reducing distinguishability of DTLS for usage in Snowflake". The goal is to create something similar to [uTLS](https://github.com/refraction-networking/utls) for DTLS. An option is to fork the repo and add the features, trying to keep it up-to-date with Pion upstream as the anti-censorship team working on Snowflake wants the library they are using to be regularly updated of course. 

Keeping the fingerprints up to date is taken care of by the added workflow and should not be something you as the maintainers should have to spend time on. The workflow as it is now auto-commits the fingerprints only after passing linting and E2E tests. Another possibility would be to just add the config/replay fingerprints feature and create a bot that auto-creates PRs for fresh fingerprints instead if you are concerned about the fingerprint generation workflow.

Mimicking will always be a bit unstable, but should not affect regular users of this library. The tradeoff has to be made clear and would be worth it for those who need fingerprint/censorship-resistance.

theodorsm:
I have now moved the fingerprint generation workflow to its own [repo](https://github.com/theodorsm/dtls-webrtc-fingerprint). This PR now only consists of the mimicking and configuration features. A BYOF (bring your own fingerprints) approach will relieve pion maintainers from the burden of keeping fingerprints up to date and working. An example of how to use generated fingerprints can be found in `pkg/mimicry/README.md`.

at-wat:
I second https://github.com/pion/dtls/pull/631#issuecomment-2066973238

Supporting a general hook mechanism to allow adding such functionality would be fine, but I think the implementation of the hook should be out of this package unless it's standardized by RFC or it's a de-facto standard.

theodorsm:
Thanks for the feedback @at-wat, @daenney. I have now reduced this PR to a simple and general hook of the client hello message and moved the implementation to a separate package. This also leaves room for implementing other anti-fingerprint features than mimicking without pushing upstream to pion. 

I will write some docs on _why_ this hooking feature is needed and how it could be used soon.

theodorsm:
This PR should be done now and will hopefully pass all the tests. My module [covertDTLS](https://github.com/theodorsm/covertDTLS) provides fingerprints and the implementation of mimicking (I am also planning some other features). An example of usage can be found in the other repo's README.

config.go
```diff
@@ -196,6 +197,15 @@ type Config struct {
...
 	PaddingLengthGenerator func(uint) uint
+
+	// ClientHelloMessageHook, if not nil, is called when a Client Hello message is sent
+	// from a client. The returned handshake message replaces the original message. This
+	// hook can be used for testing invalid messages, mimicking other implementations or
+	// randomizing fields, which is valuable for applications that need
+	// censorship-resistance by making fingerprinting more difficult.
+	//
+	// Random, SessionID, Cookie
+	ClientHelloMessageHook func(handshake.Random, []byte, []byte) handshake.Message
```
at-wat, line 208: Is it generally enough without other ClientHello parameters like CipherSuiteIDs, CompressionMethods and Extensions?

at-wat: I think it's better at interface clarity to add argument names

theodorsm:
I have also added hooks for the server hello and certificate request as the messages contain extensions or fields with lists which will be relevant to re-order for fingerprint-resistance.

at-wat (commented):
The hook API looks good to me!
Could you add a test for CertificateRequestMessageHook as well?

Also, please squash the intermediate commits before merge

at-wat (approved):
LGTM

daenney:
Nothing really stands out to me, I think this makes sense :+1:

at-wat:
@theodorsm Thank you for your contribution! Merging now

PR #640, opened 2024-06-04
Proposal: Add a OnConnectionAttempt callback for Brute Force Detection

tonisole:
#### Description
I am propose a modification to the current *Config* structure to add a callback mechanism that includes the client's address as part of the validation process. This change is crucial for implementing a Brute Force Detection mechanism in our system.

Currently, the DTLS server validation process does not provide any information about the client attempting to connect. This lack of information makes it impossible to detect if a specific IP address is repeatedly trying to guess the correct authorization, a common sign of a brute force attack.

```go
var attempts = make(map[string]int) // Map of attempts for each IP address for a Brute Force Protection

OnConnectionAttempt: func(addr net.Addr) error {
	// *************** Brute Force Attack protection ***************
	// Check if the IP address is in the map, and if the IP address has exceeded the limit
	if attempts[addr.(*net.UDPAddr).IP.String()] > 10 {
		return fmt.Errorf("too many attempts from this IP address")
	}
	// Here I increment the number of attempts for this IP address
	attempts[addr.(*net.UDPAddr).IP.String()]++
	return nil
}
```

### OnConnectionAttempt func(net.Addr) error
Whenever a connection attempt is made, the server or application can call this callback function.
The callback function can then implement logic to handle the connection attempt, such as logging the attempt, checking against a list of blocked IPs, or counting the attempts to prevent brute force attacks.
If the callback function returns an error, the connection attempt will be aborted.

#### Reference issue
No related issue

### DIFF
--- a/config.go
+++ b/config.go
@@ -11,6 +11,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"io"
+	"net"
 	"time"
 
 	"github.com/pion/dtls/v2/pkg/crypto/elliptic"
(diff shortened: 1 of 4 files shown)


Sean-Der:
@tonisole This looks great to me! I say we go with this design.

Would you mind starting a new example instead? It will make it more discoverable I think

Sorry I took so long to address this in the first place , excited to get your contribution merged :)

tonisole:
@Sean-Der Thanks for your feedback! I'm glad you like the design.

I agree that putting the Brute Force Detection examples in a separate folder will make it easier for others to find the specific examples they are looking for.

This way, we can keep the original examples intact and clearly differentiate between the two sets of examples. I’ll go ahead and create a new folder for the Brute Force Detection examples.

Sean-Der:
Merged! Great job @tonisole glad to see this land :)

I just fix some small lint issue and added a unit test! If you have anything else that could be better about this library would love to hear :)

tonisole:
Thank you @Sean-Der so much for merging my work and for your kind words! 😊

PR #649, opened 2024-07-16
Perform handshake on first read/write

kevmo314:
#### Description
Updates the connection to perform a handshake on first read/write instead of on accept. 
#### Reference issue
https://github.com/pion/dtls/issues/279

### DIFF
--- a/config.go
+++ b/config.go
@@ -4,7 +4,6 @@
 package dtls
 
 import (
-	"context"
 	"crypto/ecdsa"
 	"crypto/ed25519"
 	"crypto/rsa"
@@ -118,15 +117,6 @@ type Config struct {
 
 	LoggerFactory logging.LoggerFactory
 
-	// ConnectContextMaker is a function to make a context used in Dial(),
-	// Client(), Server(), and Accept(). If nil, the default ConnectContextMaker
-	// is used. It can be implemented as following.
-	//
-	// 	func ConnectContextMaker() (context.Context, func()) {
-	// 		return context.WithTimeout(context.Background(), 30*time.Second)
-	// 	}
-	ConnectContextMaker func() (context.Context, func())
-
 	// MTU is the length at which handshake messages will be fragmented to
 	// fit within the maximum transmission unit (default is 1200 bytes)
 	MTU int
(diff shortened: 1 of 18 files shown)


Sean-Der:
PR looks beautiful to me! Matches [Upstream](https://pkg.go.dev/crypto/tls#Conn.Handshake) and fixes biggest issue with lib at the moment :)

config.go
```diff
@@ -118,13 +118,10 @@ type Config struct {
...
-	// ConnectContextMaker is a function to make a context used in Dial(),
-	// Client(), Server(), and Accept(). If nil, the default ConnectContextMaker
-	// is used. It can be implemented as following.
+	// ConnectContextMaker is no longer used. It is kept for compatibility
+	// reasons and will be removed in a future release.
 	//
-	// 	func ConnectContextMaker() (context.Context, func()) {
-	// 		return context.WithTimeout(context.Background(), 30*time.Second)
-	// 	}
+	// Deprecated: Use the context parameter in [HandshakeContext] instead.
```
Sean-Der, line 124: I think we should kill this now. We are going with a major release, lets break now :)

conn.go
```diff
@@ -336,17 +331,35 @@ func ServerWithContext(ctx context.Context, conn net.PacketConn, rAddr net.Addr,
...
 	}
-	dconn, err := createConn(conn, rAddr, config, false)
-	if err != nil {
-		return nil, err
-	}
-	return handshakeConn(ctx, dconn, config, false, nil)
+	return createConn(conn, rAddr, config, false, nil)
+}
+
+// DialWithContext connects to the given network address and establishes a DTLS
```
Sean-Der, line 337: I am in support of getting rid of all of these.

examples/listen/cid/main.go
```diff
@@ -32,13 +26,9 @@ func main() {
 			fmt.Printf("Client's hint: %s \n", hint)
 			return []byte{0xAB, 0xC1, 0x23}, nil
 		},
-		PSKIdentityHint:      []byte("Pion DTLS Server"),
-		CipherSuites:         []dtls.CipherSuiteID{dtls.TLS_PSK_WITH_AES_128_CCM_8},
-		ExtendedMasterSecret: dtls.RequireExtendedMasterSecret,
-		// Create timeout context for accepted connection.
-		ConnectContextMaker: func() (context.Context, func()) {
```
Sean-Der, line 39: Is it easy/possible to do 30 second timeout still with new APIs?

Sean-Der:
Damn @kevmo314 really exceptional PR, I am so grateful for your help.

If you are in support I say we delete/break as much as we can now. Since we are doing a major release try to go with zero baggage. 

Also would be nice to have zero functionally change in examples. Users can reference them when updating their code.

examples/listen/cid/main.go
kevmo314: Yes, you would essentially call `conn.HandshakeContext(<30 second timeout context>)` manually before the first read/write. In other words, the sequence of calls would be something like:

```go
for {
  conn := lis.Accept()
  go func() {
    if err := conn.Handshake(<30 second timeout context>); err != nil {
      ...
    }
    // regular read/write stuff
  }()
}
```

kevmo314:
> Also would be nice to have zero functionally change in examples. Users can reference them when updating their code.

Sounds good, I'll update them to include the timeout context.

PR #671, opened 2024-08-21
Synchronize parallel calls to Conn.Close and Conn.handshake

Danielius1922:
#### Reference issue
Fixes #...

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -74,13 +74,13 @@ type Conn struct {
 
 	handshakeCompletedSuccessfully atomic.Value
 	handshakeMutex                 sync.Mutex
+	handshakeDone                  chan struct{}
 
 	encryptedPackets []addrPkt
 
 	connectionClosedByUser bool
 	closeLock              sync.Mutex
 	closed                 *closer.Closer
-	handshakeLoopsFinished sync.WaitGroup
 
 	readDeadline  *deadline.Deadline
 	writeDeadline *deadline.Deadline
@@ -256,6 +256,12 @@ func (c *Conn) HandshakeContext(ctx context.Context) error {
 		return nil
 	}
 
+	handshakeDone := make(chan struct{})
+	defer close(handshakeDone)
+	c.closeLock.Lock()
+	c.handshakeDone = handshakeDone
+	c.closeLock.Unlock()
+
 	// rfc5246#section-7.4.3
 	// In addition, the hash and signature algorithms MUST be compatible
 	// with the key in the server's end-entity certificate.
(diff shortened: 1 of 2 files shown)


Sean-Der:
Overall LGTM @Danielius1922!

Looks like a different race was found in CI. Once you can get it all to go green happy to test and can fix up any nits myself :)

Danielius1922:
> Overall LGTM @Danielius1922!
> 
> Looks like a different race was found in CI. Once you can get it all to go green happy to test and can fix up any nits myself :)

Yeah the issue is that `c.handshakeLoopsFinished.Wait()` (from Conn.Close) and `c.handshakeLoopsFinished.Add(2)` (from Conn.handshake) are called from different goroutines. Since `Close` can be called from a different goroutine, the calls on `c.handshakeLoopsFinished` need to be synchronized. 
I think the idea in `Conn.Close` is that it calls the internal `Conn.close` method, which might interrupt an ongoing handshake and then we wait using `c.handshakeLoopsFinished.Wait()`, so there is no ongoing handshake after the call to `Conn.Close`.

Handshakes themselves are serialized using `Conn.handshakeMutex`. I could try locking it in `Conn.Close`, I'm afraid of causing a deadlock if `Conn.handshake` on some code path invokes `Conn.Close`. 
Better safe than sorry here I think, so I'm gonna introduce a channel, assign/get it under a lock and wait on it outside the lock.

conn.go
```diff
@@ -405,7 +411,12 @@ func (c *Conn) Write(p []byte) (int, error) {
 // Close closes the connection.
 func (c *Conn) Close() error {
 	err := c.close(true) //nolint:contextcheck
-	c.handshakeLoopsFinished.Wait()
+	c.closeLock.Lock()
```
Danielius1922, line 414: poor `c.closeLock` gets locked and unlocked thrice in a call to `c.Close` (twice in `close` + the call here), I guess I could get the channel in `c.close` and return it? Doesn't fit semantically for `close` to return the channel, but it is an internal function, so it doesn't matter much.

conn.go
```diff
@@ -1146,8 +1161,13 @@ func (c *Conn) translateHandshakeCtxError(err error) error {
...
 
 func (c *Conn) close(byUser bool) error {
-	c.cancelHandshaker()
-	c.cancelHandshakeReader()
+	c.closeLock.Lock()
+	cancelHandshaker := c.cancelHandshaker
+	cancelHandshakeReader := c.cancelHandshakeReader
+	c.closeLock.Unlock()
+
+	cancelHandshaker()
```
Danielius1922, line 1169: @Sean-Der is it important to invoke these cancels at the beginning? if not I could move thes copying to line 1179 (after closeLock gets locked) and the invocations to 1187 (after it gets unlocked)

Danielius1922: Tests succeed locally when I do it like that, so maybe the order of calls here isn't imporant, in which case I can do it like that and avoid multiple locking/unlocking of the lock.

Sean-Der (approved):

Sean-Der:
Tagged `v3.0.2`

@Danielius1922 I added you to the org! Feel free to make branches and tags.

I think the concurrency of `Conn` can be cleaned up a lot. If you want to take a hack at it I am fully in support :)

Issue #524, opened 2023-02-01
Allow supplying external crypto.Signer for TLS signature


salrashid123:
## Summary

Allow users to supply a [crypto.Signer](https://pkg.go.dev/crypto#Signer) implementation instead of an actual private key to [dtls.v2.Config](https://godocs.io/github.com/pion/dtls/v2#Config).

This would allow an abstration allowing customers to use keys embedded into hardware (TPM) or KMS systems that implement that interface.

## Motivation

Right now users have to supply the raw private key to Config but with TPM, KMS or PKCS-11  systems, the key is not extractactable but is 'available' for use sometimes through a a `crypto.signer` interface:

eg for TPM
* [go-tpm-tools](https://pkg.go.dev/github.com/google/go-tpm-tools/client#Key.GetSigner)
* [salrashid123/signer](https://github.com/salrashid123/signer)   - [example](https://github.com/salrashid123/signer/blob/master/example/sign_verify_tpm/main.go#L172)

with this feature, a client on a device can use its embedded key for dtls connections

## Additional context

some additional refernces (untested at scale)

- JWT:
  * https://github.com/salrashid123/golang-jwt-signer
  * https://github.com/salrashid123/golang-jwt-pkcs11

- mTLS
  * https://github.com/salrashid123/go_tpm_https_embed
  * https://github.com/salrashid123/mtls_pkcs11

Sean-Der:
Hey @salrashid123 could I get your help on this? I am going to start working on a new major release and want to address all API breaks. 

Do you have a suggested API/have any pseudo-code?

salrashid123:
hi-

i'm not at all familiar with how the API is composed is used in practice (i just used it for a pretty odd [reason](https://github.com/salrashid123/mcbn)) but maybe able help in other ways.


 i did reverse the code in the library as-is to plumb through a signer ..it just involved mashing `crypto.Singer` with `crypto.PrivateKey` and what not ...meaning what i got below is just that...a hack so please look at it for ideas on what maybe much more appropriate for the new release

the  current lbirary allows you to pass a `tls.Certificates` through [Config](https://pkg.go.dev/github.com/pion/dtls/v2#Config).  the hack involved constructing a certificate where the private key is actually a crypto.signer and the cert was generated itself using that same tpm-backed signer.  Basicallly the certificate is backed by a TPM.

[...]

PR #679, opened 2024-11-02
Fix RSA signature verification issue

hoihochan:
Both `verifyKeySignature()` and `verifyCertificateVerify()` has a bug when handling RSA signature as it looks at the signature algorithm of the certificate to determine whether to verify with RSA PKCSv1.5. This will cause issues if the certificate's issuing CA uses something other than RSA (e.g. ECDSA) to sign the certificate.

Since DTLS v1.2 does not support RSA-PSS [1], we can just use RSA PKCSv1.5 verification directly if the public key of the certificate is RSA.

[1] https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-16

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -89,13 +89,11 @@ func verifyKeySignature(message, remoteKeySignature []byte, hashAlgorithm hash.A
 		}
 		return nil
 	case *rsa.PublicKey:
-		switch certificate.SignatureAlgorithm {
-		case x509.SHA1WithRSA, x509.SHA256WithRSA, x509.SHA384WithRSA, x509.SHA512WithRSA:
-			hashed := hashAlgorithm.Digest(message)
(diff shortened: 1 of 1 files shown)


Sean-Der:
Thank you so much @hoihochan !

How was the library otherwise? Any things I can improve/change?

Sean-Der (approved):

hoihochan:
> Thank you so much @hoihochan !
> 
> How was the library otherwise? Any things I can improve/change?

It's been good. We had an old fork of the library with this fix and I wanted to upstream the change so we can use the upstream library.

When I get time I would like to help on https://github.com/pion/dtls/issues/524

Sean-Der:
I would love your help!

Reach out if you need anything. Really appreciate any help :)

PR #690, opened 2025-01-24
Upgrade golangci-lint, more linters

JoeTurki:
#### Description

Introduces new linters, upgrade golangci-lint to version (v1.63.4)

#### Reference issue

https://github.com/pion/.goassets/pull/220


bench_test.go
```diff
@@ -37,6 +37,7 @@ func TestSimpleReadWrite(t *testing.T) {
 		}, false)
 		if sErr != nil {
 			t.Error(sErr)
+
```
daenney, line 40: There's a lot of changes like these, that feel like they're randomly introducing new lines between statements. Is that intentional? It doesn't improve readability from my point of view, but it introduces a ton of noise in this PR.

JoeTurki: @daenney no they are not random, we can't add random whitespace, this is `nlreturn`. `nlreturn checks for a new line before return and branch statements to increase code clarity`

Sean wanted me to enable all linters. And we had his open for while before I started working https://github.com/pion/.goassets/pull/220

We can disable that later, and i can go through repos and remove it, but imo it does make the code more readable.

JoeTurki: @daenney Just to clarify further: going through all the repos code manually, I noticed that there were actually more cases where spaces like these were already present than where they weren't. this linter (`nlreturn`) just helps standardize that behavior across the repos.

Again, Personally, I find having a blank line between branching and return statements makes the code easier to follow, but I understand this might be subjective. If the we decide it’s not a good fit, we can disable it later, and I'll go through and clean up the changes. And remove all the newlines.

For now, I think keeping it helps ensure consistency, I don't wanna merge this before I hear your input :)

Sean-Der: Sorry @daenney I should have communicated these linter changes more widely!

Since @JoeTurki has already gotten them rolled out 90% (to all the other Pion repos) are you ok with rolling forward on this?

daenney: Oh yeah, no worries, go ahead.

JoeTurki: @daenney Thank you <3

Issue #712, opened 2025-05-25
TestListenerCustomConnIDs sometimes times out or fails


zedkipp:
### Your environment.
  - Version: 806ff2f98a13d476a431d6d660dd04277bc89cb5
  - Other Information - go1.24.2, darwin/arm64

### What did you do?
```
$ cd internal/net/udp
# Even a low count value reliably reproduces a failure pretty reliably on my machine
$ go test -run TestListenerCustomConnIDs -count=100
```

### What did you expect?
The test to pass reliably

### What happened?
I'm observing the test sometimes fail with two different failure modes. I originally encountered the the timeout failure mode when running all repository Go tests locally. When attempting to reproduce the timeout, I found a second failure mode.

The first failure mode is the test failing with a timeout. Here's example output from this failure:
```
$ go test -run TestListenerCustomConnIDs -count=10 -failfast
goroutine profile: total 6
1 @ 0x10025f20c 0x10029c484 0x100360994 0x1003607c0 0x10035deac 0x1003e4a40 0x1002a5184
#	0x100360993	runtime/pprof.writeRuntimeProfile+0xb3							/usr/local/go/src/runtime/pprof/pprof.go:796
#	0x1003607bf	runtime/pprof.writeGoroutine+0x4f							/usr/local/go/src/runtime/pprof/pprof.go:755
#	0x10035deab	runtime/pprof.(*Profile).WriteTo+0x14b							/usr/local/go/src/runtime/pprof/pprof.go:377
#	0x1003e4a3f	github.com/pion/dtls/v3/internal/net/udp.TestListenerCustomConnIDs.TimeOut.func6+0x3f	/Users/zach/go/pkg/mod/github.com/pion/transport/v3@v3.0.7/test/util.go:24

1 @ 0x10029d6f8 0x1002380fc 0x100237c94 0x10031c498 0x10031e420 0x10031b734 0x10031e33c 0x10031d098 0x1003e4df0 0x100269324 0x1002a5184
#	0x10031c497	testing.(*T).Run+0x387		/usr/local/go/src/testing/testing.go:1859
#	0x10031e41f	testing.runTests.func1+0x3f	/usr/local/go/src/testing/testing.go:2279
#	0x10031b733	testing.tRunner+0xe3		/usr/local/go/src/testing/testing.go:1792
#	0x10031e33b	testing.runTests+0x3eb		/usr/local/go/src/testing/testing.go:2277
[...]

PR #713, opened 2025-05-26
Fix incorrect packet circular buffer read index after buffer resize

Zacho2:
#### Description

In #712 I encountered a test that failed in two peculiar ways. After debugging, there seemed to be a bug in the packet buffer resize logic that would result in one of the failure modes (connection reads erroneously blocking). This PR specifically addresses the test failure mode where the test times out due to erroneously blocking on reading a packet. 

The change fixes a bug that caused the packet circular buffer to have an incorrect read index after a buffer resize occurs when the write index is less than the read index. When this occurred, reading from a connection would erroneously block (packet(s) are in the buffer) and some buffered packets could be lost.

#### Reference issue
#712


Sean-Der:
Dang, that's some good debugging. Thank you @Zacho2

Sean-Der:
Would you mind making the PR against pion/transport also @Zacho2 ?

Zacho2:
@Sean-Der I would be happy to make that PR, I'm just not sure exactly what needs to change in pion/transport. Could you point me in the right direction?

I also re-pushed with a slight change to the read index fix (I realized both resize cases need to update the read index), and I simplified the test.

Sean-Der (approved):

Sean-Der:
I saw in `buffer.go` that this was copied from https://github.com/pion/transport/blob/master/packetio/buffer.go

Can you tell if it suffers from the same bug?

Zacho2:
The same bug does not seem present to me. The Buffer.grow() method seems to reset head (read) pointer to zero just like my commit. https://github.com/pion/transport/blob/43c8901250838c2740c30a7d7442deba1978fe1d/packetio/buffer.go#L115

Issue #729, opened 2025-09-20
DTLS v1.3 subtask: supported_versions extension


philipch07:
This issue is meant to track the progress regarding implementing the `supported_versions` extension for DTLS v1.3 (see the parent issue regarding the other tasks for the MVP feature set for DTLS v1.3).

PR #730, opened 2025-09-20
Add supported_versions extension feature from DTLS v1.3

philipch07:
#### Description
Currently a WIP.

This adds the `supported_versions` extension feature in accordance with [DTLS v1.3](https://www.rfc-editor.org/rfc/rfc9147.pdf) section 5.3. Note that it links to [TLS 1.3](https://www.rfc-editor.org/rfc/pdfrfc/rfc8446.txt.pdf) section 4.2.1.

#### Reference issue
Completes #729


pkg/protocol/extension/supported_versions.go
```diff
@@ -0,0 +1,120 @@
...
+// which versions of TLS it supports and by the server to indicate which
+// version it is using.
+//
+// https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.1
+type SupportedVersions struct {
+	// ClientHello's preference-ordered list.
+	Versions []protocol.Version
+
+	// ServerHello/HelloRetryRequest's selected version.
+	Selected protocol.Version
```
theodorsm: We should not have this `Selected` field here, as we should avoid state in an extension. This is usually done in the flighthandlers and the extension implementation should just focus on parsing or building byte packets. A good example of a how state is done in an extension is here: https://github.com/pion/dtls/blob/5611b1401b09ff9e3853b30f0a92ffb2aac1fe15/flight0handler.go#L70

pkg/protocol/extension/supported_versions_test.go
```diff
@@ -0,0 +1,243 @@
...
+	"testing"
+
+	"github.com/pion/dtls/v3/pkg/protocol"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSupportedVersions_ClientHello_RoundTrip(t *testing.T) {
+	ext := &SupportedVersions{
+		Versions: []protocol.Version{
+			{Major: 0xfe, Minor: 0xfc}, // DTLS v1.3
```
theodorsm: I think it would be nice to use the version variables (`Version1_3`) here.

theodorsm (changes requested):
@philipch07 Thanks for diving into this extension so fast! 

Removing the `Selected` field should simplify this implementation + tests a bit.

Additionally, we should check upon unmarshaling if the parsed major/minor version is supported by us, i.e is it one of version vars defined in `version.go`. We should discard versions not supported and only append supported versions.

pkg/protocol/extension/supported_versions.go
philipch07: That makes sense, thanks for explaining that; it was a little confusing while I was implementing it.

philipch07:
> Additionally, we should check upon unmarshaling if the parsed major/minor version is supported by us, i.e is it one of version vars defined in `version.go`. We should discard versions not supported and only append supported versions.

For discarding, I think this can cause an unmatched value with marshalling and unmarshalling: for example, lets say the list passed in is:

[`VersionX_Y`, `Version1_3`, `Version1_2`, `Version1_0`]

Then it would be marshalled (and later unmarshalled) as:

[`Version1_3`, `Version1_2`, `Version1_0`]

Which is no longer the same as the original input. Is that behavior still desirable? I'm not sure if it causes any issues, but it was just something that I noticed.

pkg/protocol/extension/supported_versions.go
```diff
@@ -0,0 +1,133 @@
...
+	// ClientHello's preference-ordered list.
+	Versions []protocol.Version
+}
+
+var errInvalidSupportedVersionsFormat = errors.New("invalid supported_versions format")
+
+func (s SupportedVersions) TypeValue() TypeValue { return SupportedVersionsTypeValue }
+
+func isSupported(v protocol.Version) bool {
+	return v.Equal(protocol.Version1_0) ||
```
JoTurk: We don't support 1.0.

pkg/protocol/extension/supported_versions.go
```diff
@@ -0,0 +1,133 @@
...
+
+		for !peek.Empty() {
+			var major, minor uint8
+			if !peek.ReadUint8(&major) || !peek.ReadUint8(&minor) {
+				return errInvalidSupportedVersionsFormat
+			}
+
+			v := protocol.Version{Major: major, Minor: minor}
+			// Only append supported versions.
+			if isSupported(v) {
```
JoTurk: I'm not sure if this filter logic should be handled here. This will treat unsupported versions as absent, and will not let us propagate appropriate handling logic.

JoeTurki (commented):
I think this should be a dumb packs/unpacks bytes and knows formats.
no is this version supported, no negotiation, no state. no policy leaking.

and all the handling/decision-making/logic should be in the handshake layer, like other extensions.

pkg/protocol/extension/supported_versions.go
philipch07: Oops, thanks for catching that!

pkg/protocol/extension/supported_versions.go
philipch07: I see, I may have misunderstood what @theodorsm was saying earlier regarding unmarshalling and filtering unsupported/unrecognized versions. My original understanding was that we should filter in the `unmarshal()` function, which is why I commented earlier about my concern regarding it no longer being RTT-able with the `VersionX_Y` example.

> the handling/decision-making/logic should be in the handshake layer, like other extensions.

That's in line with what @theodorsm was saying in his comment as well, so I think I misunderstood this.

@theodorsm when you get the chance, can you confirm what the desired behavior is? I don't mind changing it at all! I just want to confirm that we're all on the same page for this.

theodorsm: I was a bit quick with my feedback. To clarify, we should check if the value of the unmarshaled version is one of the defined versions by the spec (version 1.0, 1.2 and 1.3), all other values should be discarded (ex: 0xf0, 0xe8). An example where we do this is with parsing elliptic curve values: https://github.com/pion/dtls/blob/120a895e99cfee70febfee965ce67189ed8be48f/pkg/protocol/extension/supported_elliptic_curves.go#L59-L61 

I agree that we should not pick in the extension which version to use, but only parse supported values from the spec (does not mean that pion supports it though, and we can later reject it as part of the handshake logic).

pkg/protocol/extension/supported_versions.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		if !extData.Skip(1 + int(listLen)) {
+			return errInvalidSupportedVersionsFormat
+		}
+
+		return nil
+	}
+
+	// Otherwise, expect ServerHello/HelloRetryRequest selected_version, which should be exactly 2 bytes.
```
theodorsm, line 119: Isn't the rest of this function redundant? Wouldn't the peek above also unmarshal a list of only one version too?

pkg/protocol/extension/supported_versions.go
```diff
@@ -0,0 +1,131 @@
...
+type SupportedVersions struct {
+	// ClientHello's preference-ordered list.
+	Versions []protocol.Version
+}
+
+var errInvalidSupportedVersionsFormat = errors.New("invalid supported_versions format")
+
+func (s SupportedVersions) TypeValue() TypeValue { return SupportedVersionsTypeValue }
+
+func isSupported(v protocol.Version) bool {
```
theodorsm: This could be moved to the version.go file and be a "isValidVersion" (1.0, 1.2, 1.3) function. Or be a util function to parse version bytes and return Version1_0, Version1_2 or Version1_3.

philipch07: Yes, going off of this idea I decided to make 4 functions which are small and well defined.

pkg/protocol/extension/supported_versions.go
philipch07: Makes sense, thank you for the clarification!

pkg/protocol/extension/supported_versions.go
philipch07: No, although I wish there was a cleaner way to write this. I was able to clean up the `Marshal()` though, so that's quite a bit nicer now.

PR #737, opened 2025-09-30
Add HKDF funcs for Key Scheduling - DTLS v1.3

### LINKED ISSUE #740: DTLS v1.3 HKDF functions for key scheduling
Add the HKDF functions necessary for key scheduling which include:
- HKDF-Expand-Label
- HKDF-Extract
- Derive-Secret

See [DTLS v1.3](https://www.rfc-editor.org/rfc/rfc9147.pdf) section 5.9 and [TLS 1.3](https://www.rfc-editor.org/rfc/pdfrfc/rfc8446.txt.pdf) section 7.1.

philipch07:
#### Description
This adds the HKDF related functions for the key scheduling feature in accordance with [DTLS v1.3](https://www.rfc-editor.org/rfc/rfc9147.pdf) section 5.9. Note that it links to [TLS 1.3](https://www.rfc-editor.org/rfc/pdfrfc/rfc8446.txt.pdf) section 7.1.

This aims to add:
- [x] HKDF-Expand-Label
- [x] HKDF-Extract
- [x] Derive-Secret

Note that the architecture is still a WIP (see https://github.com/pion/dtls/issues/738) so the current file structures are subject to change.

#### Reference issue
Closes #740

### DIFF
--- a/pkg/crypto/keyschedule/keyschedule.go
+++ b/pkg/crypto/keyschedule/keyschedule.go
@@ -29,7 +29,8 @@ func HkdfExtract(hash func() hash.Hash, salt, ikm []byte) ([]byte, error) {
 	if hash == nil {
 		return nil, errMissingHashFunction
 	}
-	// The order of the ikm and salt arguments are different than the RFC.
+	// Note: Go's hkdf.Extract signature is (hash, ikm, salt),
+	// while RFC 5869 specifies HKDF-Extract(salt, IKM)
 	return hkdf.Extract(hash, ikm, salt)
 }
 
@@ -37,6 +38,12 @@ func HkdfExtract(hash func() hash.Hash, salt, ikm []byte) ([]byte, error) {
 func HkdfExpandLabel(hash func() hash.Hash, secret []byte, label string, context []byte, length int) ([]byte, error) {
 	fullLabel := []byte(DTLS13prefix + label)
 
+	if hash == nil {
+		return nil, errMissingHashFunction
+	}
+
+	// RFC 8446 section 7.1
+	// opaque label<7..255>
 	if len(fullLabel) < 7 {
 		return nil, errLabelTooSmall
 	} else if len(fullLabel) > 255 {
@@ -45,10 +52,6 @@ func HkdfExpandLabel(hash func() hash.Hash, secret []byte, label string, context
 
 	if len(context) > 255 {
 		return nil, errContextTooBig
-	}
-
-	if hash == nil {
-		return nil, errMissingHashFunction
 	}
 
 	var builder cryptobyte.Builder


theodorsm:
@JoeTurki, I agree that we should wait to merge this until we are closer to a WIP for DTLS 1.3. Unfortunately, this blocks much of the implementation, so we should provide a similar mock API meanwhile.

pkg/crypto/keyschedule/keyschedule.go
```diff
@@ -0,0 +1,93 @@
...
+const (
+	DTLS13prefix = "dtls13" // RFC 9147 section 5.9
+)
+
+// HkdfExtract implements RFC 5869 section 2.2.
+func HkdfExtract(hash func() hash.Hash, salt, ikm []byte) ([]byte, error) {
+	if hash == nil {
+		return nil, errMissingHashFunction
+	}
+	// The order of the ikm and salt arguments are different than the RFC.
```
adrianosela: nit: this comment implies that we are doing something against the RFC, rather than just the hkdf.Extract signature being how it is. I would get rid of the comment altogether, or make it more explicit e.g.

```
// Note: Go's hkdf.Extract signature is (hash, ikm, salt),
// while RFC 5869 specifies HKDF-Extract(salt, IKM)
```

pkg/crypto/keyschedule/keyschedule.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil, errLabelTooSmall
+	} else if len(fullLabel) > 255 {
+		return nil, errLabelTooBig
+	}
+
+	if len(context) > 255 {
+		return nil, errContextTooBig
+	}
+
+	if hash == nil {
```
adrianosela: nit: This check is cheaper than the above two, and the argument appears first in the func sig. Consider moving it up.

pkg/crypto/keyschedule/keyschedule.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	// The order of the ikm and salt arguments are different than the RFC.
+	return hkdf.Extract(hash, ikm, salt)
+}
+
+// HkdfExpandLabel implements RFC 8446 section 7.1 with RFC 9147 section 5.9's defined DTLS prefix.
+func HkdfExpandLabel(hash func() hash.Hash, secret []byte, label string, context []byte, length int) ([]byte, error) {
+	fullLabel := []byte(DTLS13prefix + label)
+
+	if len(fullLabel) < 7 {
```
adrianosela, line 47: nit: I know the 7 comes directly from the RFC, but would be useful to have a comment of why its 7. e.g. prefix is "dtls13" (6 chars) + at least 1 char label = 7 bytes minimum

theodorsm:
@adrianosela, thanks for the review! I have added some of the nits;) 

Regarding the tests, I chose to not add them: 
1. The are no ACVP vectors for DTLS 1.3 similar to what openssl's tls13 tests use: https://pages.nist.gov/ACVP/#kdfs. I checked boringSSL and wolfSSL without finding any similar values there either for DTLS. I leaning towards interoptability testing later for verification.
2. `hkdfExpandLabelDTLS13` is just a wrapper around `Expand` from `"golang.org/x/crypto/hkdf"`. I assume we can trust it as they do testing of the function: https://cs.opensource.google/go/go/+/master:src/crypto/hkdf/hkdf_test.go;drc=93b6475e1de0f055139bd0926048ee8e36aa7c42;l=318

philipch07:
@theodorsm thank you for cleaning up this PR (and the key share extension), and apologies for not getting around to them. The work that both you and @adrianosela have been doing in DTLS has been amazing!! :))

PR #738, opened 2025-10-05
DTLS 1.3 FSM architecture

### LINKED ISSUE #731: DTLS 1.3 config flag
This issue is part of a larger effort to implement an MVP of DTLS 1.3 (see parrent issue #727). We should have a config flag for enabling DTLS 1.3. The DTLS 1.3 implementation should keep the handshake/flight logic separate from the existing DTLS 1.2 implementation.

theodorsm:
#### Description

I have made a WIP DTLS 1.3 <s>configuration</s> and architecture. 

<s>As suggested by @JoTurk, it implementes a functional options pattern to provide flexibility in the future. A DTLS 1.3 configuration is created with `NewConfigVersion13(c Config, opts ...OptionVersion13) (*Config, error)` where an option/config can be implemented like this: 
```
func WithOption13(f func(b bool)) OptionVersion13 {
	return func(c *Config) error {
		c.comeConfig = b
		return nil
	}
}
*/
```
</s>

The codebase for DTLS 1.2 and DTLS 1.3 splits off in `conn.go` <s>with the new `handshake13` function and the new handshakeFSM in `handshaker_13.go` (note that neither reflects DTLS 1.3 logic yet, just a copy/paste)</s>. I suggest we split off as much code as possible suffixed with `_13` to keep the development of DTLS 1.3 separate from 1.2.

This PR contains a test of enabling DTLS 1.3 and verifies that an error is returned correctly, as we have not yet started to implement DTLS 1.3 flights.

I would appreciate some input from other maintainers on this! @JoTurk, @daenney, @Sean-Der. 

#### Reference issue
Fixes #731

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -20,6 +20,7 @@ import (
 	"github.com/pion/dtls/v3/pkg/crypto/signaturehash"
 	"github.com/pion/dtls/v3/pkg/protocol"
 	"github.com/pion/dtls/v3/pkg/protocol/alert"
+	"github.com/pion/dtls/v3/pkg/protocol/extension"
 	"github.com/pion/dtls/v3/pkg/protocol/handshake"
 	"github.com/pion/dtls/v3/pkg/protocol/recordlayer"
 	"github.com/pion/logging"
@@ -192,12 +193,12 @@ func createConn(
 	}
 
 	minVersion := config.minVersion
-	if !minVersion.Equal(protocol.Version1_2) && !minVersion.Equal(protocol.Version1_3) {
+	if !minVersion.Equal(protocol.Version1_3) {
 		minVersion = protocol.Version1_2
 	}
 
 	maxVersion := config.maxVersion
-	if !maxVersion.Equal(protocol.Version1_2) && !maxVersion.Equal(protocol.Version1_3) {
+	if !maxVersion.Equal(protocol.Version1_3) {
 		maxVersion = protocol.Version1_2
 	}
 
@@ -310,25 +311,6 @@ func (c *Conn) HandshakeContext(ctx context.Context) error { //nolint:cyclop
 	c.handshakeDone = handshakeDone
 	c.closeLock.Unlock()
 
-	if c.isVersion13Enabled() {
-		c.state.version = protocol.Version1_3
-		var initialFlight flightVal
-		if c.state.isClient {
-			initialFlight = flightVal(flight13_1)
-		} else {
-			initialFlight = flightVal(flight13_0)
-		}
-		initialFSMState := handshakePreparing
-
-		if err := c.handshake(ctx, initialFlight, initialFSMState); err != nil {
-			return err
-		}
-		c.log.Trace("Handshake DTLS 1.3 Completed")
-
-		return nil
-	}
-	c.state.version = protocol.Version1_2
-
 	// rfc5246#section-7.4.3
 	// In addition, the hash and signature algorithms MUST be compatible
 	// with the key in the server's end-entity certificate.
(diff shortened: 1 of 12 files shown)


theodorsm:
The goal here is not to be complete, but to provide a best-effort skeleton with my current understanding of how the DTLS 1.3 implementation would look. <s>It still contains some duplicate code in `conn.go` (the linter loves to complain about it), which shall be modified when we implement more of the DTLS 1.3 spec (such as the new record layer encoding #755 ).</s>

I still think the `_13` suffix is a good way of structuring to move on without large refactorings. This reduces the number of new public APIs (now private) we would need to expose when using a sub-package. I reduced the flight handlers files by grouping them into `flighthandlers_client_13` and `flighthandlers_server_13`. We can always decide to make a sub-package later if we see it necessary.

theodorsm:
The config is split to another PR #823

theodorsm:
There are some TODOs here, but they are mainly regarding the new [record demultiplexing](https://datatracker.ietf.org/doc/html/rfc9147#section-4.1). This is addressed in #793.

This PR is not supposed to be complete, but the changes are private so we can get it merged to unblock working on the DTLS 1.3 implementation further. It will affect #766

conn.go
```diff
@@ -310,6 +310,24 @@ func (c *Conn) HandshakeContext(ctx context.Context) error {
...
 	c.closeLock.Unlock()
 
+	if c.isVersion13Enabled() {
+		c.state.version = protocol.Version1_3
+		var initialFlight flightVal
+		if c.state.isClient {
+			initialFlight = flightVal(flight13_1)
+		} else {
+			initialFlight = flightVal(flight13_0)
+		}
```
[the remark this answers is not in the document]
theodorsm: What is actually used is negotiated during the handshake, but if the DTLS 1.3 is enabled at all we should either send a CH with supported_version or expect one as a server. If DTLS 1.2 is negotiated we can hand over the session to a FSM12.

conn.go
```diff
@@ -310,6 +310,25 @@ func (c *Conn) HandshakeContext(ctx context.Context) error {
 	c.handshakeDone = handshakeDone
 	c.closeLock.Unlock()
 
+	if c.isVersion13Enabled() {
+		c.state.version = protocol.Version1_3
+		var initialFlight flightVal
+		if c.state.isClient {
+			initialFlight = flightVal(flight13_1)
+		} else {
+			initialFlight = flightVal(flight13_0)
```
JoTurk: I think copilot is correct, I don't think should set state.version if multiple version are enabled to version1_3, this early.

Even if we got it to work somehow now, I think we should be very explicit here, And it shouldn't be a 1.3 DTLS with 1.2 fallback logic, we should be in:
1. DTLS 1.2 only mode.
2. DTLS 1.3 only mode.
3. dual stack 1.2/1.3 with version not selected yet (if we're early in the connection).

I'm concerned because after seeing all the bugs introduced by dual stack libraries (mainly the two bugs introduced by firefox) I'm worried we'll introduce similar bugs if we don't write this very explicitly.

This dual-stack pattern should make it easier to implement the client and the server path, and have the code expected to interact with DTLS 1.2 

Also it will make it easier to reasoon about

> In addition to the handshake messages that are deprecated by the TLS
   1.3 specification, DTLS 1.3 furthermore deprecates the
   HelloVerifyRequest message originally defined in DTLS 1.0.  DTLS
   1.3-compliant implementations MUST NOT use the HelloVerifyRequest to
   execute a return-routability check.  A dual-stack DTLS 1.2 / DTLS 1.3
   client MUST, however, be prepared to interact with a DTLS 1.2 server.

handshaker_13.go
```diff
@@ -0,0 +1,119 @@
...
+func (s *handshakeFSM13) Run(ctx context.Context, conn flightConn, initialState handshakeState) error {
+	state := initialState
+	defer func() {
+		close(s.closed)
+	}()
+	for {
+		s.cfg.log.Tracef("[handshake13:%s] %s: %s", srvCliStr(s.state.isClient), s.currentFlight.String(), state.String())
+		if s.cfg.onFlightState13 != nil {
+			s.cfg.onFlightState13(s.currentFlight, state)
+		}
```
JoTurk, line 83: I think we should call onFlightState13 after all the stats logic.

conn.go
```diff
@@ -1178,23 +1217,48 @@ func (c *Conn) isHandshakeCompletedSuccessfully() bool {
...
+		}
+		c.handshakeConfig.onFlightState13 = func(_ flightVal13, s handshakeState) {
+			currentFlight := c.fsm.(*handshakeFSM13).currentFlight //nolint:forcetypeassert
+			shouldClose := currentFlight.isLastSendFlight() || currentFlight.isLastRecvFlight()
+			//nolint:godox
+			// TODO: should be moved to FSM.
+			if shouldClose && c.setHandshakeCompletedSuccessfully() {
+				close(done)
+			}
+		}
```
JoTurk, line 1514: Related to the call place comment, I shouldn't this use the actual FSM state?

> The implementation reads an ACK from the peer: upon receiving an ACK for a partial flight (as mentioned in [Section 7.1](https://datatracker.ietf.org/doc/html/rfc9147#sending-acks)), the implementation transitions to the SENDING state, where it retransmits the unacknowledged portion of the flight, adjusts and re-arms the retransmit timer, and returns to the WAITING state. Upon receiving an ACK for a complete flight, the implementation cancels all retransmissions and either remains in WAITING, or, if the ACK was for the final flight, transitions to FINISHED.

state.go
```diff
@@ -90,6 +93,7 @@ type serializedState struct {
 	RemoteConnectionID    []byte
 	IsClient              bool
 	NegotiatedProtocol    string
+	version               uint16
```
JoTurk: version will need to be public for gob's marshal /  unmarshal.

JoTurk (commented):
Thank you. Great work.

Sorry it took me so long, I had to refresh my memory with the specs again because it has been a while, I think i have a fresh memory of rfc9147 and 6347 now :)

conn.go
theodorsm: I have added `prepareHandshakeStart` to try to handle this better. 

Both the DTLS 1.2 and 1.3 only mode should be quite clean, but dual-stack is a bit messier. The dual-stack mode negotiates the version without starting a FSM by sending a CH and waiting for the responding SH or waiting for a CH as a server, then pass the state/flights to the correct FSM. 

I think doing this slighly messy operation up front and then having clean FSM would be a good way to reduce such bugs. We should write some tests for this behavior tho (which I haven't really done here).

handshaker_13.go
theodorsm: I am not quite catching why that would be better? Is it that we can catch the state change earlier (and before the error return?)

conn.go
theodorsm: I agree, thus having the TODO in place. I think we can have this as a placeholder function, then moving it to the FSM during #818

state.go
theodorsm: Removed the versions from the serialized state. We can add it later when DTLS 1.3 functionality is public.

handshaker_13.go
JoTurk: because if the transition failed in the state logic we shouldn't call onFlightState13 i think, also i think it makes more sense to have the callback after the transition because it's a side effect, not a pre-action, but the rest of the comment is related to my other comment, the todo in the code.

conn.go
JoTurk: > We should write some tests for this behavior tho (which I haven't really done here).

Yeah and more DTLS 1.2 interops targeted at the dual-stack mode, because even if we follow the spec, we might break some incorrect DTLS 1.2 libraries.

theodorsm: Should we also consider a refactoring? Right now the `handshake` path starts the FSM together with a main read loop (and the reason for needing `readAndBufferNoFSM`). Perhaps we could decouple it better to make it easier to reason about when to launch the handshake and the current state. I am a little unsure how it would look, or if I am missing some reason why the current architecture is needed. I am somewhat afraid that such an effort will introduce more bugs and require a lot of time, on the other hand we already have seen some FSM bugs recently with the timers.

JoTurk: I think yes, I can start working on the refactor, Also i think we're in a close stage, so it's fine to start breaking APIs, I'll make a branch to lock the current state, and tag from there if we need to fix bugs or issues targeted at 1.2, what do you think?

theodorsm: @JoTurk, sounds like good plan, let's do it! Thanks:)

theodorsm: I have also added some dual-stack tests (negotiating DTLS 1.2) that are passing now.

handshaker_13.go
theodorsm: I still think it makes sense to have it where it is currently. We also need to consider the initalState, thus we run the onFlightState13 once when we enter the loop. onFlightState13 is only called after the err check when it loops around.  

If we move the callback after the switch we also have to call onFlightState13 before we enter the for loop, I think that is slightly more confusing having it called two places.

JoTurk: Okay but what if we processed with a flight callback side-effect i.e closing / calling done or handshake complete, then the state handler returned an error? and we already emitted the side effect for the state that didn't actually complete. 

This is more of a nit and not a blocking or anything feel free to mark this conversation as resolved, I'm not attached to it :) and I think this will be done when it moves to FSM right?

conn.go
JoTurk: @theodorsm sorry i missed this, this is great thank you.

flighthandlers_client_13.go
```diff
@@ -0,0 +1,165 @@
...
+			MasterKeyIdentifier: cfg.localSRTPMasterKeyIdentifier,
+		})
+	}
+
+	// connection ID
+
+	clientHello := &handshake.MessageClientHello{
+		Version:   protocol.Version1_2,
+		SessionID: state.SessionID,
+		Cookie:    state.cookie,
```
theodorsm, line 139: Should explicitly set to be empty.

legacy_cookie:
A DTLS 1.3-only client MUST set the legacy_cookie field to zero length. If a DTLS 1.3 ClientHello is received with any other value in this field, the server MUST abort the handshake with an "illegal_parameter" alert.

JoTurk (approved):
looks good to merge i think, all the issues and the design discussion can be fixed before we release DTLS 1.3.

PR #742, opened 2025-10-08
Add more tests for prf

philipch07:
#### Description
Added more tests to improve cov for `prf.go`.

#### Reference issue
N/A

### DIFF
--- a/pkg/crypto/prf/prf_test.go
+++ b/pkg/crypto/prf/prf_test.go
@@ -4,7 +4,11 @@
 package prf
 
 import (
+	"crypto/ecdh"
+	"crypto/rand"
 	"crypto/sha256"
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/pion/dtls/v3/pkg/crypto/elliptic"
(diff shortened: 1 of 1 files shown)


pkg/crypto/prf/prf_test.go
```diff
@@ -4,11 +4,16 @@
...
+	"crypto/ecdh"
+	"crypto/rand"
 	"crypto/sha256"
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/pion/dtls/v3/pkg/crypto/elliptic"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
JoTurk: nitpick: can we change the old tests in the file to use require or move the new tests to use assert if possible?

philipch07: Yes I can change them to use the same one. Is there any specific preference that Pion has in general?

JoTurk: require calls FailNow, while assert doesn't stop the test on failure, I personally prefer `assert` because it makes tests with goroutunes easier to debug, but generally in cases like this we should use one version in the same file.

philipch07: Good to know, thanks! Just fixed it.

PR #748, opened 2025-10-12
Move the signature hash errors into the same file

philipch07:
#### Description
The errors in the signature hash module were in a separate file. I'm assuming it's related to an older version of go. It can be condensed into the same file so we don't have an extra file just to define 4 errors.

#### Reference issue
N/A


philipch07:
> I'm not sure about this change, we use the `errors.go` pattern a lot for errors in the same package, yes it's not required if we have a single file in the package, but I'm not sure if we should change if we already did it, it's nice to have a place for all the errors defined by a package.

That's true. I think this PR is addressing the edge case since there's only really one file in this package. It doesn't really matter all that much though, perhaps it's easiest to keep things as is unless things get changed up more. Going forward, do you have a suggestion for whether we should still follow the `errors.go` pattern even if a package only has one file, or would you prefer for the errors to be defined at the top of the one file?

> Go doesn't care about files in the same package, it's all the same scope.

Yes, this makes sense.

JoeTurki:
Maybe using a lint rule (extreme), what do you think?

PR #749, opened 2025-10-12
Add the key share extension

### LINKED ISSUE #743: DTLS v1.3 key_share extension
This issue is meant to track the progress regarding implementing the `key_share` extension for DTLS v1.3 (see the parent issue regarding the other tasks for the MVP feature set for DTLS v1.3).

The extension isn't deeply described in [RFC 9147 (DTLS v1.3)](https://www.rfc-editor.org/rfc/rfc9147.pdf), but it refers to the implementation from section 4.2.8 (including 4.2.8.1 and 4.2.8.2) of [RFC 8446 (TLS v1.3)](https://www.rfc-editor.org/rfc/pdfrfc/rfc8446.txt.pdf).

philipch07:
#### Description
This adds the `supported_versions` extension feature in accordance with [DTLS v1.3](https://www.rfc-editor.org/rfc/rfc9147.pdf) which refers to [TLS 1.3](https://www.rfc-editor.org/rfc/pdfrfc/rfc8446.txt.pdf) section 4.2.8, 4.2.8.1, and 4.2.8.2.

Note about the ci failures:
This currently uses a global variable to test that my current logic is valid, so it will break all the DTLS v1.2 tests. At the moment, this is blocked by #738 as it requires a proper config/switch. I'm not sure what the best way of setting the toggle would be in `extensions.go`, but hopefully my current approach makes some sense.

#### Reference issue
Closes #743.

### DIFF
--- a/pkg/protocol/extension/key_share.go
+++ b/pkg/protocol/extension/key_share.go
@@ -117,7 +117,7 @@ func (k *KeyShare) Unmarshal(data []byte) error { //nolint:cyclop
 			var groupU16 uint16
 			var raw cryptobyte.String
 
-			if !peek.ReadUint16(&groupU16) || !peek.ReadUint16LengthPrefixed(&raw) {
+			if !peek.ReadUint16(&groupU16) || !peek.ReadUint16LengthPrefixed(&raw) || len(raw) == 0 {
 				return errInvalidKeyShareFormat
 			}
 
@@ -129,11 +129,9 @@ func (k *KeyShare) Unmarshal(data []byte) error { //nolint:cyclop
 
 			seenGroups[group] = struct{}{}
 
-			if elliptic.Curves()[group] {
-				entry.Group = group
-				entry.KeyExchange = append([]byte(nil), raw...)
-				k.ClientShares = append(k.ClientShares, entry)
-			}
+			entry.Group = group
+			entry.KeyExchange = append([]byte(nil), raw...)
+			k.ClientShares = append(k.ClientShares, entry)
 		}
 
 		// consume vector (2 bytes length + vecLen)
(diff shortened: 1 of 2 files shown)


pkg/protocol/extension/key_share.go
```diff
@@ -0,0 +1,192 @@
...
+
+	if hasClientShares {
+		seenGroups := []elliptic.Curve{}
+
+		for _, e := range k.ClientShares {
+			if slices.Contains(seenGroups, e.Group) {
+				return nil, errDuplicateKeyShare
+			}
+
+			seenGroups = append(seenGroups, e.Group)
```
philipch07: Do you think it would be better to have `seenGroups` be a map instead of a slice since we only use it to check for the existence of a group and we never have to traverse it?

pkg/protocol/extension/key_share.go
```diff
@@ -0,0 +1,192 @@
...
+				return errInvalidKeyShareFormat
+			}
+
+			group := elliptic.Curve(groupU16)
+
+			if slices.Contains(seenGroups, group) {
+				return errDuplicateKeyShare
+			}
+
+			seenGroups = append(seenGroups, group)
```
philipch07: And same for here?

pkg/protocol/extension/key_share.go
theodorsm: I was unfamiliar with map[]struct{} pattern in your original implementation. Had to read up on it, cool way of checking existence. Changed it back:)

theodorsm:
@philipch07 thanks for the feedback. Regarding naming, we could change it to something like `supported_groups_and_elliptic_curves`. I think the comment is enough, as it will be shown to a developer using this package in the generated documentation and by LSP servers.

pkg/protocol/extension/key_share.go
```diff
@@ -0,0 +1,190 @@
...
+
+			group := elliptic.Curve(groupU16)
+
+			if _, ok := seenGroups[group]; ok {
+				return errDuplicateKeyShare
+			}
+
+			seenGroups[group] = struct{}{}
+
+			if elliptic.Curves()[group] {
```
JoTurk: I think we should do the same thing as the other PR, we should return all groups regardless if we support them or not, and let the upper layer handle it.

pkg/protocol/extension/key_share.go
```diff
@@ -0,0 +1,190 @@
...
+				}
+			})
+		}
+	})
+
+	return builder.Bytes()
+}
+
+// Unmarshal decodes the extension.
+func (k *KeyShare) Unmarshal(data []byte) error { //nolint:cyclop
```
JoTurk, line 93: I think we should reject 0 length keys ( opaque key_exchange<1..2^16-1>;) https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8

pkg/protocol/extension/key_share.go
theodorsm: True, I removed the check.

pkg/protocol/extension/key_share.go
theodorsm: Good catch! During unmarshaling I added a check for zero length key_exchange by checking if `len(raw) == 0`.

theodorsm:
@JoTurk, thanks for the review! (no worries about forgetting) I have fixed your comments.

philipch07:
@theodorsm Thanks for updating this! Sorry I was busy recently. I would approve it but I can't since I'm the PR author 😅

JoTurk (approved):
Thank you for the fixes theo, and welcome back :)

PR #759, opened 2025-12-03
Backoff handshake retransmit

### LINKED ISSUE #758: No backoff on PSK mismatch creating flood of handshake between server and client.
### Your environment.
  - Version: v3.0.7
 
Image

### What did you do?

When trying out some cases for running a DTLS Reverse Proxy I noticed that I got an extreme amount of UDP traffic when my clients psk did not match the servers stored psk for the clients identity.

I tried with a pion based DTLS client and an MbedTLS based DTLS client and I got the same  result.

I have a more advanced server program in a local development branch but this intial test also showed the same issue: https://github.com/u9n/udp-edge/blob/master/proxy/main.go

### What did you expect?
Some error and maybe an abort on the handshake procedure.

But in discussion on discord we found the following from the RFC

```
Unlike TLS implementations, DTLS implementations SHOULD NOT respond
to invalid records by terminating the connection.  See Section 4.1.2.7 for details on this.
```
So apparently you should not tear down the connection on a fault. Since it caused a race of UDP messages between the server and client it would be good if there was a backoff for answering a client that continuously send the wrong psk (bad data). 

A way to notice if a client is behaving bad would also be good. The RFC also mentions this:

```
Implementations SHOULD detect when
   a peer is persistently sending bad messages and terminate the local
   connection state after such misbehavior is detected.
```

If there was an callback you could register on bad messages it would be possible to leave that detection to the application as it might be application specific.

### What happened?
When the PSKs where mismatched it resulted in `server: decrypt failed: dtls temporary: failed to decrypt packet: ccm: message authentication failed` and a continuous flight of `Server Hello/Server Key Exchange/Server Hello Done` and `Client Key Exchange / Change Cipher Specs / Encrypted Handshake message` between the server and client until the handshake timeout stopped

JoTurk:
#### Description
we currently spam retransmits under some conditions like mismatched PSK.
This matches openssl. 
#### Reference issue
Fixes #758

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -682,6 +682,89 @@ func TestPSKHintFail(t *testing.T) {
 	assert.ErrorIs(t, <-clientErr, pskRejected, "TestPSK: Client should fail with pskRejected error")
 }
 
+func TestPSKMismatchNoRetransmitLoop(t *testing.T) {
+	lim := test.TimeOut(time.Second * 20)
+	defer lim.Stop()
+
+	report := test.CheckRoutines(t)
+	defer report()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	var serverWrites atomic.Int32
+	var clientWrites atomic.Int32
+
+	ca, cb := dpipe.Pipe()
+	defer func() {
+		_ = ca.Close()
+	}()
+	defer func() {
+		_ = cb.Close()
+	}()
+
+	caCount := &connWithCallback{Conn: ca}
+	caCount.onWrite = func([]byte) {
+		clientWrites.Add(1)
+	}
+	cbCount := &connWithCallback{Conn: cb}
+	cbCount.onWrite = func([]byte) {
+		serverWrites.Add(1)
+	}
+
+	clientErr := make(chan error, 1)
+	serverErr := make(chan error, 1)
+
+	go func() {
+		conf := &Config{
+			PSK: func([]byte) ([]byte, error) {
+				return []byte("client-psk"), nil
+			},
+			PSKIdentityHint: []byte("Client Identity"),
+			CipherSuites:    []CipherSuiteID{TLS_PSK_WITH_AES_128_CCM_8},
+		}
+
+		c, err := testClient(ctx, dtlsnet.PacketConnFromConn(caCount), caCount.RemoteAddr(), conf, false)
+		if c != nil {
+			_ = c.Close() //nolint:contextcheck
+		}
+		clientErr <- err
+	}()
+
+	go func() {
+		conf := &Config{
+			PSK: func([]byte) ([]byte, error) {
+				return []byte("server-psk"), nil
+			},
+			CipherSuites: []CipherSuiteID{TLS_PSK_WITH_AES_128_CCM_8},
+		}
+
+		s, err := testServer(ctx, dtlsnet.PacketConnFromConn(cbCount), cbCount.RemoteAddr(), conf, false)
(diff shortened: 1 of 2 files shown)


conn_test.go
```diff
@@ -682,6 +682,89 @@ func TestPSKHintFail(t *testing.T) {
...
+
+	var serverHandshakeErr *HandshakeError
+	var clientHandshakeErr *HandshakeError
+	assert.ErrorAs(t, serverErrRes, &serverHandshakeErr)
+	assert.ErrorAs(t, clientErrRes, &clientHandshakeErr)
+
+	serverCount := serverWrites.Load()
+	clientCount := clientWrites.Load()
+
+	time.Sleep(2 * time.Second)
```
theodorsm, line 760: Should we increase this timer to 61 seconds to check if retransmits are not sent even with the max time?

handshaker.go
```diff
@@ -166,16 +166,27 @@ func srvCliStr(isClient bool) string {
 	return "server"
 }
 
+// normalizeRetransmitInterval protects against negative 0 and negative retransmit interval.
```
theodorsm: Small typo: _negative, 0 and negative_

theodorsm (changes requested):
Wow, cool that you jumped on this so quickly after the bug report! 

State-machines are a bit hard to reason about, and this is a weird bug I don't understand how happened. We should do some root-cause analysis on how the timer could even be 0 or negative without using the `FlightInterval` API. We should normalize [`FlightInterval` ](https://github.com/pion/dtls/blob/7ab1bc9b2afabcb5a2dda1de8854671c4b61371b/conn.go#L157) too. 

I think this is a good (temporary) bug fix for the retransmits, but I am left wondering how it happened. We should also provide a callback function for alerting later.

handshaker.go
JoTurk: Fixed the comment, thank you.

handshaker.go
```diff
@@ -290,12 +301,13 @@ func (s *handshakeFSM) wait(ctx context.Context, conn flightConn) (handshakeStat
...
-
-				return handshakeSending, nil
+				// ignore incoming retransmit hints, only rely on the timer-driven path below
+				// https://github.com/pion/dtls/issues/758
+				continue
 			}
 
 			nextFlight, alert, err := parse(ctx, conn, s.state, s.cache, s.cfg)
-			s.retransmitInterval = s.cfg.initialRetransmitInterval
+			s.retransmitInterval = normalizeRetransmitInterval(s.cfg.initialRetransmitInterval)
```
philipch07: Can we normalize the cfg's retransmitInterval just once instead of recomputing it every time in the loop?

conn_test.go
JoTurk: The test has a timeout of 20s and the clients have a timeout of 10s, 60s will be too much for the CI. we already have many issues with the CI.

handshaker.go
JoTurk: Good point, but that will be a behavior change that i don't want to touch in this fix pr.

conn_test.go
JoTurk: Yeah i had to revert to 2s, sadly more than that is too much for github CIs. the old behavior was 5k rts in that time frame.

handshaker.go
JoTurk: I removed normilization all together, will do it in another PR because it's unrelated to this fix.

theodorsm (approved):
LGTM. Also tried a test locally with longer timeouts (past 60 seconds) and it passed.

PR #761, opened 2025-12-11
Implement TLS_EMPTY_RENEGOTIATION_INFO_SCSV

JoTurk:
#### Description
We used to send `renegotiation_info` all the time, which was incorrect and caused some clients to break until the fix in https://github.com/pion/dtls/pull/711 this implements extra detection via `TLS_EMPTY_RENEGOTIATION_INFO_SCSV`. 

>  Both the SSLv3 and TLS 1.0/TLS 1.1 specifications require
   implementations to ignore data following the ClientHello (i.e.,
   extensions) if they do not understand it.  However, some SSLv3 and
   TLS 1.0 implementations incorrectly fail the handshake in such a
   case.  This means that clients that offer the "renegotiation_info"
   extension may encounter handshake failures.  In order to enhance
   compatibility with such servers, this document defines a second
   signaling mechanism via a special Signaling Cipher Suite Value (SCSV)
   "TLS_EMPTY_RENEGOTIATION_INFO_SCSV", with code point {0x00, 0xFF}.
   This SCSV is not a true cipher suite (it does not correspond to any
   valid set of algorithms) and cannot be negotiated.  Instead, it has
   the same semantics as an empty "renegotiation_info" extension, as
   described in the following sections.  Because SSLv3 and TLS
   implementations reliably ignore unknown cipher suites, the SCSV may
   be safely sent to any server.  The SCSV can also be included in the
   SSLv2 backward compatible CLIENT-HELLO (see Appendix E.2 of
   [RFC5246]).
   
https://datatracker.ietf.org/doc/html/rfc5746#section-3.3
#### Reference issue
fixes https://github.com/pion/webrtc/issues/3264

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -281,7 +281,18 @@ func testServer(
 	return conn, conn.HandshakeContext(ctx)
 }
 
-func sendClientHello(cookie []byte, ca net.Conn, sequenceNumber uint64, extensions []extension.Extension) error {
+func sendClientHello(
+	cookie []byte,
+	ca net.Conn,
+	sequenceNumber uint64,
+	extensions []extension.Extension,
+	cipherSuiteIDsOverride ...uint16,
+) error {
+	cipherSuites := cipherSuiteIDsOverride
+	if len(cipherSuites) == 0 {
+		cipherSuites = cipherSuiteIDs(defaultCipherSuites())
+	}
+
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):
amazing debugging, thanks for fixing this Joe :)

PR #762, opened 2025-12-15
Handshake fragments assembly refactoring

trs00:
- Determining the end of the fragment by looking at the fragment length, rather than handshake header's length
- Using totalBufferSize to avoid walks the linked list to determine the total buffer size
- Using totalFragmentCount to avoid list grow to an unbounded size with 0-length fragments
- Removing the need to try to reassemble(by recursively calling a function for each fragment) on every new packet

### DIFF
--- a/fragment_buffer.go
+++ b/fragment_buffer.go
@@ -65,7 +65,7 @@ func (f *fragmentBuffer) push(buf []byte) (isHandshake, isRetransmit bool, err e
 	}
 
 	frag := new(fragment)
-	for buf = buf[recordlayer.FixedHeaderSize:]; len(buf) != 0; frag = new(fragment) {
+	for buf = buf[recordlayer.FixedHeaderSize:]; len(buf) != 0; frag = new(fragment) { //nolint:gosec // G602
 		if err := frag.handshakeHeader.Unmarshal(buf); err != nil {
 			return false, false, err
 		}


Sean-Der:
Thank you so much @trs00! Changes look great to me.

Would you be able/interested in fixing the lints? If you don't have time I am happy to push to your branch directly :)

fragment_buffer.go
```diff
@@ -94,35 +114,26 @@ func (f *fragmentBuffer) pop() (content []byte, epoch uint16) {
...
-				if fragmentEnd != f.handshakeHeader.Length && f.handshakeHeader.FragmentLength != 0 {
-					if !appendMessage(fragmentEnd) {
-						return false
-					}
-				}
-
-				rawMessage = append(f.data, rawMessage...)
+	if frags.fragmentsLength != frags.handshakeLength {
+		return nil, 0
+	}
```
philipch07, line 119: I think both the previous implementation and this new implementation check for exact fragments. The [spec mentions](https://datatracker.ietf.org/doc/html/rfc6347#section-4.2.3) (second to last paragraph): 

> DTLS implementations MUST be able to handle overlapping fragment ranges. This allows senders to retransmit handshake messages with smaller fragment sizes if the PMTU estimate changes.

but that is likely out of scope for this PR. I guess it's still worth mentioning though, in case you happen to have any ideas for how to do that cleanly.

trs00: Yes, I think it is possible to handle overlaps by using list:
`fragmentByOffset map[uint32]*fragment -> fragmentByOffset *list.List`

on push()
- iterate fragments list and insert it in order of offset if it is not a duplicate
- check fragment for overlapping with previous/next fragments and adjust fragmentsLength

on pop()
- keep the same check for complete msg detection without a need to reiterate:
frags.fragmentsLength != frags.handshakeLength

So the difference is that with fragmentByOffset map we iterate through the fragments only once when we have all the fragments(without overlap support)
With the list approach we would need to iterate fragmentByOffset list on each fragment

philipch07: This idea sounds good + I like the translation of how this can be done with the map. My only remaining concern is dealing with fragment conflicts, I spent quite some time yesterday thinking about it but am unfortunately not familiar enough in this space to come up with a secure solution :(

The question I got stuck on was: how do we know which fragments are incorrect vs which are correct? For example, if we believe the first fragment received (in an overlap) must be correct, then could an attacker be able to poison the first fragment which causes subsequent fragments to think they're incorrect? And what if we instead believe the latest fragment with an overlap received is correct (and could an attacker poison the last fragment?) and so on and so forth.

(Btw the discussion for this doesn't block the pr at all, I'm just curious to know what your thoughts are!)

trs00: I think it might not be possible to authenticate individual fragments and it is necessary to wait until cryptographic verification at the end of handshake

philipch07: Makes sense, thank you!

trs00:
@JoeTurki may I ask if merge option is not available due to access restrictions or it is necessary to have 100% Patch coverage?

JoeTurki:
@trs00 Hello, I squash merged, If you plan to make more changes, I can send you an invite to the org so you can make branches directly, and merge PRs after approval :)

Thank you a lot.

PR #764, opened 2025-12-20
Fix gosec slice bounds warnings

asayyah:
Add nolint directives for safe slice accesses where bounds are guaranteed by fixed-size arrays or prior bounds checks.

### DIFF
--- a/pkg/protocol/extension/extension.go
+++ b/pkg/protocol/extension/extension.go
@@ -58,40 +58,41 @@ func Unmarshal(buf []byte) ([]Extension, error) { //nolint:cyclop
 	}
 
 	for offset := 2; offset < len(buf); {
-		if len(buf) < (offset + 2) {
+		bufView := buf[offset:] //nolint:gosec // offset bounded by loop condition
+		if len(bufView) < 2 {
 			return nil, errBufferTooSmall
 		}
 		var err error
-		switch TypeValue(binary.BigEndian.Uint16(buf[offset:])) {
+		switch TypeValue(binary.BigEndian.Uint16(bufView)) {
 		case ServerNameTypeValue:
-			err = unmarshalAndAppend(buf[offset:], &ServerName{})
+			err = unmarshalAndAppend(bufView, &ServerName{})
 		case SupportedEllipticCurvesTypeValue:
-			err = unmarshalAndAppend(buf[offset:], &SupportedEllipticCurves{})
+			err = unmarshalAndAppend(bufView, &SupportedEllipticCurves{})
 		case SupportedPointFormatsTypeValue:
-			err = unmarshalAndAppend(buf[offset:], &SupportedPointFormats{})
+			err = unmarshalAndAppend(bufView, &SupportedPointFormats{})
 		case SupportedSignatureAlgorithmsTypeValue:
-			err = unmarshalAndAppend(buf[offset:], &SupportedSignatureAlgorithms{})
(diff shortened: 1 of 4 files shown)


pkg/protocol/extension/extension.go
```diff
@@ -62,27 +62,27 @@ func Unmarshal(buf []byte) ([]Extension, error) { //nolint:cyclop
...
-			err = unmarshalAndAppend(buf[offset:], &RenegotiationInfo{})
+			err = unmarshalAndAppend(buf[offset:], &RenegotiationInfo{}) //nolint:gosec // bounds checked above
 		case ConnectionIDTypeValue:
-			err = unmarshalAndAppend(buf[offset:], &ConnectionID{})
+			err = unmarshalAndAppend(buf[offset:], &ConnectionID{}) //nolint:gosec // bounds checked above
 		case SupportedVersionsTypeValue:
-			err = unmarshalAndAppend(buf[offset:], &SupportedVersions{})
+			err = unmarshalAndAppend(buf[offset:], &SupportedVersions{}) //nolint:gosec // bounds checked above
 		default:
 		}
```
JoTurk, line 88: can we do something like this, instead of all of the nolints?

```go

bufView := buf[offset:]

if len(bufView) < 2 {
	return errBufferTooSmall
}

extType := TypeValue(binary.BigEndian.Uint16(bufView[:2]))

switch extType {
case ServerNameTypeValue:
	err = unmarshalAndAppend(bufView, &ServerName{})
case SupportedEllipticCurvesTypeValue:
	err = unmarshalAndAppend(bufView, &SupportedEllipticCurves{})
case SupportedPointFormatsTypeValue:
	err = unmarshalAndAppend(bufView, &SupportedPointFormats{})
case SupportedSignatureAlgorithmsTypeValue:
	err = unmarshalAndAppend(bufView, &SupportedSignatureAlgorithms{})
case UseSRTPTypeValue:
	err = unmarshalAndAppend(bufView, &UseSRTP{})
case ALPNTypeValue:
	err = unmarshalAndAppend(bufView, &ALPN{})
case UseExtendedMasterSecretTypeValue:
	err = unmarshalAndAppend(bufView, &UseExtendedMasterSecret{})
case RenegotiationInfoTypeValue:
	err = unmarshalAndAppend(bufView, &RenegotiationInfo{})
case ConnectionIDTypeValue:
	err = unmarshalAndAppend(bufView, &ConnectionID{})
case SupportedVersionsTypeValue:
	err = unmarshalAndAppend(bufView, &SupportedVersions{})
default:
}
```

asayyah: done!

pkg/protocol/handshake/message_client_hello.go
```diff
@@ -43,8 +43,7 @@ func (m *MessageClientHello) Marshal() ([]byte, error) {
 	}
 
 	out := make([]byte, handshakeMessageClientHelloVariableWidthStart)
-	out[0] = m.Version.Major
-	out[1] = m.Version.Minor
+	copy(out[0:2], []byte{m.Version.Major, m.Version.Minor})
```
JoTurk: Nit: even if it will prob never matter performance wise, and will likely be optimized by the compiler, the temporary slice is unnecessary. even if we'll have to add nolint. would prefer something like out[0], out[1] = m.Version.Major, m.Version.Minor // nolint...

but that's up to you.

asayyah: yeah that's a good point, fixed it! thank you!

PR #766, opened 2025-12-29
Allow intercepting and injecting DTLS packets

fippo:
which allows embedding them into STUN.

### DIFF
--- a/config.go
+++ b/config.go
@@ -228,9 +228,10 @@ type Config struct { //nolint:dupl
 	CertificateRequestMessageHook func(handshake.MessageCertificateRequest) handshake.Message
 
 	// OutboundHandshakePacketInterceptor is an optional callback that can be set to
-	// intercept outgoing raw handshake packets. It is called with the raw packet bytes.
+	// intercept outgoing raw handshake packets. It is called with the raw packet bytes
+	// and a boolean flag specifying if this is the last packet of a flight.
 	// The interceptor can decide to drop the packet by returning true.
-	OutboundHandshakePacketInterceptor func(packet []byte) bool
+	OutboundHandshakePacketInterceptor func(packet []byte, end bool) bool
 	// InboundHandshakePacketNotifier is an optional callback that can be set to
 	// receive notifications about incoming raw handshake packets. It is called with
 	// the raw packet bytes after the packet has been processed.
--- a/config.go
+++ b/config.go
@@ -227,15 +227,15 @@ type Config struct { //nolint:dupl
 	// message is sent from a server. The returned handshake message replaces the original message.
 	CertificateRequestMessageHook func(handshake.MessageCertificateRequest) handshake.Message
 
-	// OutboundHandshakePacketInterceptor is an optional callback that can be set to
+	// outboundHandshakePacketInterceptor is an optional callback that can be set to
 	// intercept outgoing raw handshake packets. It is called with the raw packet bytes
 	// and a boolean flag specifying if this is the last packet of a flight.
 	// The interceptor can decide to drop the packet by returning true.
-	OutboundHandshakePacketInterceptor func(packet []byte, end bool) bool
-	// InboundHandshakePacketNotifier is an optional callback that can be set to
+	outboundHandshakePacketInterceptor func(packet []byte, end bool) bool
+	// inboundHandshakePacketNotifier is an optional callback that can be set to
 	// receive notifications about incoming raw handshake packets. It is called with
 	// the raw packet bytes after the packet has been processed.
-	InboundHandshakePacketNotifier func(packet []byte)
+	inboundHandshakePacketNotifier func(packet []byte)
 
 	// OnConnectionAttempt is fired Whenever a connection attempt is made,
 	// the server or application can call this callback function.
--- a/config.go
+++ b/config.go
@@ -227,6 +227,11 @@ type Config struct { //nolint:dupl
 	// message is sent from a server. The returned handshake message replaces the original message.
 	CertificateRequestMessageHook func(handshake.MessageCertificateRequest) handshake.Message
 
+	// HandshakePacketInterceptor is a callback that can be set to intercept
+	// outgoing raw handshake packets. It is called with the raw packet bytes
+	// The interceptor can decide to drop the packet by returning true.
+	HandshakePacketInterceptor func(packet []byte) bool
+
 	// OnConnectionAttempt is fired Whenever a connection attempt is made,
 	// the server or application can call this callback function.
 	// The callback function can then implement logic to handle the connection attempt, such as logging the attempt,
--- a/config.go
+++ b/config.go
@@ -184,7 +184,7 @@ type Config struct { //nolint:dupl
 
 	// InsecureSkipVerifyHello, if true and when acting as server, allow client to
 	// skip hello verify phase and receive ServerHello after initial ClientHello.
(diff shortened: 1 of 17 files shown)


conn.go
```diff
@@ -474,9 +480,12 @@ func (c *Conn) writePackets(ctx context.Context, pkts []*packet) error {
 
 	var rawPackets [][]byte
 
+	// TODO: this is not quite correct.
```
fippo: "writePackets" is an odd name, this writes a series of records? Intercepting here is wrong since the goal is to intercept the handshake packets, not the individual records.
Theoretically handshake and non-handshake records could be mixed but I think that isn't allowed (at least in DTLS 1.2)

conn.go
```diff
@@ -963,6 +1046,14 @@ func (c *Conn) handleIncomingPacket(
...
 		return false, false, nil, nil
 	} else if isHandshake {
+		/*
+					if c.handshakePacketInterceptor != nil {
+			            // TODO: this seems to handle individual flights/records, not packets
+			            // which is what this function is called with despite its name.
+			            // We need the full packets for SPED so we can crc32 them (at least the ones
+			            // received in plain!)
+					}
+		*/
```
fippo: same problem as above, this is about records, not packets.

conn.go
fippo: Best way I came up with so far is to split the handling of handshake packets into a `writeHandshakePackets` (which seems like a general improvement) and call that from handshaker.go -- done in f501150589fd685f216f2ffa9fd58d6ce88e0782

conn.go
fippo: likewise, readAndBuffer already has a `hasHandshake` which could be used for an inbound interceptor (which should only be used to to let the piggybacking controller calculate and store the crc32).

fippo: 787aa438943cf152a4d36c16965fcc2a44e6b3ed adds an inbound notifier (which can not intercept packets)

conn.go
```diff
@@ -776,14 +860,19 @@ func (c *Conn) readAndBuffer(ctx context.Context) error { //nolint:cyclop
 		}
 	}
 	if hasHandshake {
+		if c.inboundHandshakePacketNotifier != nil {
+			// Would it be useful to know this was injected?
+			// Should this work on a copy?
```
fippo: flagging

theodorsm: I would perfer that we do not implement such logic here, thus the users of this functionality should handle it themselves. I think it's reasonable to assume that it's the same library that would both use inject and the notifier.

conn.go
```diff
@@ -732,20 +775,61 @@ var poolReadBuffer = sync.Pool{ //nolint:gochecknoglobals
 	},
 }
 
-func (c *Conn) readAndBuffer(ctx context.Context) error { //nolint:cyclop
-	bufptr, ok := poolReadBuffer.Get().(*[]byte)
-	if !ok {
-		return errFailedToAccessPoolReadBuffer
+func (c *Conn) InjectPacket(p []byte, rAddr net.Addr) {
```
theodorsm: It is not clear from the name how this packet it injected. I think `InjectIncomingPacket` is more precise.

conn.go
theodorsm: Non-handshake packets are still processed in the `writeHandshakePackets`. I assume it is only supposed to be done in the `writePackets` function after this refactor.

config.go
```diff
@@ -218,6 +218,15 @@ type Config struct {
 	// message is sent from a server. The returned handshake message replaces the original message.
 	CertificateRequestMessageHook func(handshake.MessageCertificateRequest) handshake.Message
 
+	// OutboundHandshakePacketInterceptor is an optional callback that can be set to
+	// intercept outgoing raw handshake packets. It is called with the raw packet bytes.
+	// The interceptor can decide to drop the packet by returning true.
+	OutboundHandshakePacketInterceptor func(packet []byte) bool
```
theodorsm: I suggest we could make the interceptors more general and allow for modification of handshake messages, similar to the message hooks above (I already know about some usecases, for example anti-fingerprinting).  

`OutboundHandshakePacketInterceptor func(packet []byte) []byte` and `InboundHandshakePacketInterceptor func(packet []byte) []byte`. Rather than a bool deciding if we drop a packet, we can drop a packet if the returned byte buffer is empty.

conn_test.go
```diff
@@ -3486,3 +3486,101 @@ func TestCloseWithoutHandshake(t *testing.T) {
...
+
+	ca, cb := dpipe.Pipe()
+	serverCert, err := selfsign.GenerateSelfSigned()
+	assert.NoError(t, err)
+
+	var client *Conn
+	server, err := Server(dtlsnet.PacketConnFromConn(cb), cb.RemoteAddr(), &Config{
+		Certificates: []tls.Certificate{serverCert},
+		OutboundHandshakePacketInterceptor: func(packet []byte) bool {
+			client.InjectPacket(packet, ca.RemoteAddr())
```
theodorsm: Could you make a seprate test for injected packets? Now the interceptor and the injection functionallity are tested closely together, I would suggest seperating them. Would there be any way of testing the interceptors without using the injection functionallity?

theodorsm (changes requested):
Thanks for working on this @fippo! I think it's useful functionallity to have for multiple usecases and I think the re-factor makes sense. I have made some comments to your PR.

theodorsm:
@Sean-Der ahhh I overlooked this API, thanks, that's much cleaner! Does this solve your usecase @fippo (if so, we should close this PR)? You can take inspiration from [pkg/net/net.go](https://github.com/pion/dtls/blob/master/pkg/net/net.go).

config.go
fippo: wouldn't such a use-case implement its own flight generator which allows way more flexibility?

conn.go
fippo: if you mean change cipher spec which doesn't have a handshake content type then yes, it is handled here since they need to go into the same network packet as part of a handshake flight (terminology... this one isn't great)

conn_test.go
fippo: Testing them in complete isolation is only going to result in weaker assertions like "number of handshake packets seen" whereas `TestInboundNotifier` asserts that what one connection is intercepting (or not) as part of the handshake is seen by the other connection.

JoTurk:
@Sean-Der @fippo What if we provide a sync BoringSSL-BIO-like interface to do the initial handshake and then resume / hand-off the connection once the handshake is done? I think we have everything we need in FSM12 and FSM13. And this can be done quickly.

fippo:
The BIO stuff is not an interface I would recommend to an enemy even! Doing it "manually" seems good enough.

JoTurk:
I really liked BIO when i was doing the Boringssl interop, before switching to Shim TCP / PacketedBio :)

Anyways, I actually meant something like sync blocking DoHandshake interface, and after the handshake is done we can hand-off the connection to DTLS (like the current ResumeWithOptions API), For DTLS 1.3 this will be at the Finished boundary and it will ignore all the post-finished state machine.

zshang-oai:
@JoTurk We ported SPED to boringssl + Pion webrtc and it's quite painful. +1 to what @fippo said.

JoTurk:
@zshang-oai can you please check this proposed API https://github.com/pion/dtls/pull/1094 you should be able to implement SPED with this API, without us exporting intercept and inject API.

fippo:
this is getting too complicated (unsurprisingly)

PR #770, opened 2026-01-09
Add Cookie extension with unit test and fuzzing

### LINKED ISSUE #769: DTLS 1.3: Cookie extension
Tracking the progress of the cookie extension for DTLS 1.3 defined in s[ection 4.2.2 of RFC 8446](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.2). 

According to [RFC 9147 section 5.1:](https://www.rfc-editor.org/rfc/rfc9147.html#section-5.1)
> The "cookie" extension is defined in [Section 4.2.2](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.2) of [[TLS13](https://www.rfc-editor.org/rfc/rfc9147.html#RFC8446)].

theodorsm:
#### Description

DTLS 1.3 has moved it's cookie handling to an extension. 

#### Reference issue

Fixes #769

### DIFF
--- a/pkg/protocol/extension/cookie.go
+++ b/pkg/protocol/extension/cookie.go
@@ -45,8 +45,7 @@ func (c *CookieExt) Marshal() ([]byte, error) {
 func (c *CookieExt) Unmarshal(data []byte) error { //nolint:cyclop
 	val := cryptobyte.String(data)
 	var extension uint16
-	val.ReadUint16(&extension)
-	if TypeValue(extension) != c.TypeValue() {
+	if !val.ReadUint16(&extension) || TypeValue(extension) != c.TypeValue() {
 		return errInvalidExtensionType
 	}
 


JoTurk (commented):
Should we make a branch for dtls 1.3? I'm just worried we might break something for dtls 1.2 or if we want to make a release for dtls 1.2. cherry-picking commits  might be hard after few months.
I know we talked about this before but we merged few bug fixes and enhancements since then. 
Up to you. I'm just suggesting :)

JoTurk:
@Sean-Der theo also talked about it here https://github.com/pion/dtls/pull/738

theodorsm:
@Sean-Der, @JoTurk Adding extension parsers shouldn't affect DTLS 1.2. I think it makes sense to have them in the master branch already. For more involved stuff we have to re-visit #738 (let's have a discussion about it there). [crypto/tls implements all extensions side by side for tls 1.2 and tls 1.3](https://cs.opensource.google/go/go/+/master:src/crypto/tls/handshake_messages.go;l=71;bpv=0;bpt=1)

theodorsm:
Also, fuzzing might be a bit overkill for this simple extension, but I think all unmarshalling of raw bytes should be fuzzed and this sets an easy example.

pkg/protocol/extension/cookie.go
```diff
@@ -0,0 +1,67 @@
...
+type CookieExt struct {
+	Cookie []byte
+}
+
+// TypeValue returns the extension TypeValue.
+func (c CookieExt) TypeValue() TypeValue {
+	return CookieTypeValue
+}
+
+var errCoookieExtFormat = errors.New("invalid cookie format")
```
JoTurk:
```suggestion
var errCookieExtFormat = errors.New("invalid cookie format")
```

pkg/protocol/extension/cookie.go
```diff
@@ -0,0 +1,67 @@
...
+	})
+
+	return b.Bytes()
+}
+
+// Unmarshal populates the extension from encoded data.
+func (c *CookieExt) Unmarshal(data []byte) error { //nolint:cyclop
+	val := cryptobyte.String(data)
+	var extension uint16
+	val.ReadUint16(&extension)
```
JoTurk:
```suggestion
	if !val.ReadUint16(&extension) {
		return errBufferTooSmall
	}
```

pkg/protocol/extension/cookie.go
```diff
@@ -0,0 +1,67 @@
...
+		return errInvalidExtensionType
+	}
+
+	var extData cryptobyte.String
+	if !val.ReadUint16LengthPrefixed(&extData) {
+		return errBufferTooSmall
+	}
+
+	var cookie cryptobyte.String
+	if !extData.ReadUint16LengthPrefixed(&cookie) {
```
JoTurk: maybe we should also reject extensions with trailing bytes?

pkg/protocol/extension/cookie.go
```diff
@@ -0,0 +1,67 @@
...
+		return errCoookieExtFormat
+	}
+
+	cookieLength := len(cookie)
+	if cookieLength == 0 || cookieLength > 0xfffd {
+		return errCoookieExtFormat
+	}
+
+	c.Cookie = cookie
```
JoTurk: also i think we should copy here like how we did in key_share, and other extensions.

pkg/protocol/extension/cookie.go
theodorsm: This was checked by `if cookieLength == 0 || cookieLength > 0xfffd` to verify the vector `opaque cookie<1..2^16-1>;`. I changed the check so the intention is more clear.

pkg/protocol/extension/cookie.go
theodorsm: OOOOOOOOopise ^^

PR #773, opened 2026-01-13
Add PreSharedKey extensions for DTLS 1.3

### LINKED ISSUE #771: DTLS 1.3 PSK extensions
This issue tracks the progress of the new extensions required by DTLS 1.3 to use Pre Shared Key (PSK).

To use PSK, a client must send a `pre_shared_key` ([RFC8446 section 4.2.11](https://www.rfc-editor.org/rfc/rfc8446#section-4.2.11)) and a `psk_key_exchange_modes` ([RFC8446 section 4.2.9](psk_key_exchange_modes)) extension.

theodorsm:
#### Description
DTLS 1.3 uses two new extensions to negotiate PreSharedKey (PSK) mode. 

_Progress_: 
- [x] `pre_shared_key` extension
- [x] `pre_shared_key` tests and fuzz
- [x] `psk_exchange_modes` extension
- [x] `psk_exchange_modes` tests and fuzz

#### Reference issue
Fixes #771

### DIFF
--- a/pkg/protocol/extension/pre_shared_key.go
+++ b/pkg/protocol/extension/pre_shared_key.go
@@ -136,7 +136,7 @@ func (p *PreSharedKey) Unmarshal(data []byte) error { //nolint:cyclop
 
 	// Ensure there is one binder value per identity in list,
 	// and check for trailing bytes.
-	if len(binders) != len(identities) || !extData.Empty() {
+	if len(p.Binders) != len(p.Identities) || !extData.Empty() {
 		return errPreSharedKeyFormat
 	}
 


pkg/protocol/extension/pre_shared_key.go
```diff
@@ -0,0 +1,127 @@
...
+				}
+				b.AddUint16LengthPrefixed(func(b *cryptobyte.Builder) {
+					b.AddBytes(pskIdentity.Identity)
+				})
+				b.AddUint32(pskIdentity.ObfuscatedTicketAge)
+			}
+		})
+		b.AddUint16LengthPrefixed(func(b *cryptobyte.Builder) {
+			for _, binder := range p.Binders {
+				if len(binder) < 32 {
```
adrianosela: can we define a const for this e.g. `minPSKBinderSize = 32`

pkg/protocol/extension/pre_shared_key.go
```diff
@@ -0,0 +1,127 @@
...
+// SPDX-License-Identifier: MIT
+
+package extension
+
+import (
+	"golang.org/x/crypto/cryptobyte"
+)
+
+// PreSharedKey implements the PreSharedKey extension in DTLS 1.3.
+// See RFC 8446 section 4.2.11. Pre-Shared Key Extension.
```
adrianosela: Can we add a quick-click URL to https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.11 for consistency with other extension types e.g.

```
// SupportedVersions is a TLS extension used by the client to indicate
// which versions of TLS it supports and by the server to indicate which
// version it is using.
//
// https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.1
```

pkg/protocol/extension/pre_shared_key.go
```diff
@@ -0,0 +1,127 @@
...
+type PreSharedKey struct {
+	// ClientHello: OfferedPsks
+	Identities []PskIdentity
```
adrianosela, line 18: can we add comments on each of these fields explaining what they represent?

pkg/protocol/extension/pre_shared_key.go
```diff
@@ -0,0 +1,127 @@
...
+
+package extension
+
+import (
+	"golang.org/x/crypto/cryptobyte"
+)
+
+// PreSharedKey implements the PreSharedKey extension in DTLS 1.3.
+// See RFC 8446 section 4.2.11. Pre-Shared Key Extension.
+type PreSharedKey struct {
```
adrianosela, line 16: can we also document the dual-purpose nature of this struct (ClientHello vs ServerHello)

pkg/protocol/extension/pre_shared_key.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	for !binders.Empty() {
+		var binder cryptobyte.String
+		if !binders.ReadUint8LengthPrefixed(&binder) || len(binder) < 32 {
+			return errPreSharedKeyFormat
+		}
+		p.Binders = append(p.Binders, PskBinderEntry(binder))
+	}
+
```
adrianosela, line 136: Other extensions check that there is no trailing data leftover. Would it be appropriate to do that here also?

adrianosela, line 136: Per the RFC, the number of binders must correspond 1:1 to the number of identities (in ClientHello).

Perhaps a good check to add?

pkg/protocol/extension/psk_key_exchange_modes.go
```diff
@@ -0,0 +1,78 @@
...
+
+package extension
+
+import (
+	"golang.org/x/crypto/cryptobyte"
+)
+
+// PskKeyExchangeModes implements the PskKeyExchangeModes extension in DTLS 1.3.
+// See RFC 8446 section 4.2.9. Pre-Shared Key Exchange Modes.
+type PskKeyExchangeModes struct {
```
adrianosela, line 14: Same feedback about adding a URL to the rfc section here.

adrianosela (commented):
Code looks good. Pointed out some minor validation checks missing in `(p *PreSharedKey) Unmarshal`, as well as some nice-to-have documentation.

pkg/protocol/extension/pre_shared_key.go
adrianosela: This would take care of my two comments:

```
  // PreSharedKey represents the "pre_shared_key" extension for DTLS 1.3.
  // This extension is used in both ClientHello and ServerHello messages,
  // but only the relevant fields should be populated for each context.
  // See RFC 8446 section 4.2.11.
  //
  // https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.11
  type PreSharedKey struct {
        Identities     []PskIdentity     // ClientHello only - offered PSK identities
        Binders        []PskBinderEntry  // ClientHello only - corresponding binder values
        SelectedIdentity *uint16         // ServerHello only - index of selected identity
  }
```

theodorsm: I added your suggestion.

pkg/protocol/extension/pre_shared_key.go
theodorsm: Nice catch! I added checks for both the 1:1 and trailing bytes.

pkg/protocol/extension/pre_shared_key.go
```diff
@@ -0,0 +1,142 @@
...
+	// ServerHello
+	if len(extData) == 2 {
+		var selected uint16
+		if !extData.ReadUint16(&selected) {
+			return errPreSharedKeyFormat
+		}
+		p.SelectedIdentity = selected
+
```
adrianosela, line 105: Missing trailing bytes check for serverhello here.

pkg/protocol/extension/pre_shared_key.go
```diff
@@ -0,0 +1,142 @@
...
+	for !binders.Empty() {
+		var binder cryptobyte.String
+		if !binders.ReadUint8LengthPrefixed(&binder) || len(binder) < minPSKBinderSize {
+			return errPreSharedKeyFormat
+		}
+		p.Binders = append(p.Binders, PskBinderEntry(binder))
+	}
+
+	if len(binders) != len(identities) || !extData.Empty() {
```
adrianosela: add a comment of why these 2 checks are needed.

pkg/protocol/extension/pre_shared_key.go
theodorsm: Shouldn't be needed considering we already check that the length is 2 and read an uint16.

pkg/protocol/extension/pre_shared_key.go
```diff
@@ -0,0 +1,144 @@
...
+		var binder cryptobyte.String
+		if !binders.ReadUint8LengthPrefixed(&binder) || len(binder) < minPSKBinderSize {
+			return errPreSharedKeyFormat
+		}
+		p.Binders = append(p.Binders, PskBinderEntry(binder))
+	}
+
+	// Ensure there is one binder value per identity in list,
+	// and check for trailing bytes.
+	if len(binders) != len(identities) || !extData.Empty() {
```
JoTurk: Is this a typo? I think we should check p.[binders/identities], aren't binders and identities already consumed here, and this will never be true?
```suggestion
	if len(p.Binders) != len(p.Identities) || !extData.Empty() {
```

theodorsm: Def. a typo, thanks!!

Issue #772, opened 2026-01-13
DTLS 1.3 Certificate messages


theodorsm:
This issue aims to track the implementation of the `certificate` and the `certificate_request` handshake messages for DTLS 1.3.

The new version of DTLS updates the structure of the `Certificate` (RFC 8446 section [4.4.2](https://www.rfc-editor.org/rfc/rfc8446#section-4.4.2)) and the `CertificateRequest` (RFC 8446 section [4.3.2](https://www.rfc-editor.org/rfc/rfc8446#section-4.3.2)) handshake messages. In DTLS 1.2 these are the defined in [RFC 5246 section 7.4.2](https://www.rfc-editor.org/rfc/rfc5246#section-7.4.2) and [RFC 5246 section 7.4.4](https://www.rfc-editor.org/rfc/rfc5246#section-7.4.4) respectively. 

These new messages should be implemented as `MessageCertificate13` in `pkg/protocol/handshake/message_certificate_13.go` and as `MessageCertificateRequest13` in `pkg/protocol/handshake/message_certificate_request_13.go`.

PR #774, opened 2026-01-14
DTLS 1.3 CertificateRequest + Certificate Messages

adrianosela:
## DTLS 1.3 CertificateRequest + Certificate Messages

Closes https://github.com/pion/dtls/issues/772

- Implements TLS 1.3 **Certificate Request** message as `CertificateRequest13` per [RFC 8446 section 4.3.2](https://www.rfc-editor.org/rfc/rfc8446#section-4.3.2)
- Implements TLS 1.3 **Certificate** message as `Certificate13` per  [RFC 8446 section 4.4.2](https://www.rfc-editor.org/rfc/rfc8446#section-4.4.2)
- Adds comprehensive unit tests for both

 cc:// @theodorsm

### DIFF
--- a/pkg/protocol/handshake/message_certificate_13_test.go
+++ b/pkg/protocol/handshake/message_certificate_13_test.go
@@ -264,6 +264,144 @@ func TestMessageCertificate13_EmptyCertData(t *testing.T) {
 	assert.ErrorIs(t, err, errInvalidCertificateEntry)
 }
 
+func TestMessageCertificate13_CertDataAtMaxBoundary(t *testing.T) {
+	// Test cert_data at exactly 2^24-1 bytes (max allowed per RFC)
+	// This will fail because serialized entry exceeds certificate_list limit
+	certData := make([]byte, 0xffffff) // 2^24-1 bytes
+	for i := range certData {
+		certData[i] = byte(i)
+	}
+
+	msg := &MessageCertificate13{
+		CertificateRequestContext: []byte{},
+		CertificateList: []CertificateEntry13{
+			{
+				CertificateData: certData,
+				Extensions:      []extension.Extension{},
+			},
+		},
+	}
+
+	_, err := msg.Marshal()
+	// Should fail with certificate list too long
+	// Serialized size = 3 (cert_data prefix) + 0xffffff (cert_data) + 2 (ext prefix) = 0x1000004
+	assert.ErrorIs(t, err, errCertificateListTooLong)
+}
+
+func TestMessageCertificate13_CertDataJustBelowMaxBoundary(t *testing.T) {
+	// Test cert_data at 2^24-1-5 bytes (should succeed for single cert with no extensions)
+	// Serialized size = 3 + (0xffffff-5) + 2 = 0xffffff (exactly at limit)
+	certData := make([]byte, 0xffffff-5) // 2^24-1-5 bytes
+	for i := range certData {
+		certData[i] = byte(i)
+	}
+
+	msg := &MessageCertificate13{
+		CertificateRequestContext: []byte{},
+		CertificateList: []CertificateEntry13{
+			{
+				CertificateData: certData,
+				Extensions:      []extension.Extension{},
+			},
+		},
+	}
+
+	data, err := msg.Marshal()
+	assert.NoError(t, err)
+
+	// Verify round-trip
+	out := &MessageCertificate13{}
+	err = out.Unmarshal(data)
+	assert.NoError(t, err)
+	assert.Equal(t, len(msg.CertificateList[0].CertificateData), len(out.CertificateList[0].CertificateData))
+}
+
+func TestMessageCertificate13_CertDataOneByteOverBoundary(t *testing.T) {
+	// Test cert_data at 2^24-1-4 bytes (one byte over the single-cert-no-ext limit)
+	// Serialized size = 3 + (0xffffff-4) + 2 = 0x1000000 (exceeds 2^24-1)
+	certData := make([]byte, 0xffffff-4) // 2^24-1-4 bytes
+	for i := range certData {
+		certData[i] = byte(i)
+	}
+
+	msg := &MessageCertificate13{
+		CertificateRequestContext: []byte{},
+		CertificateList: []CertificateEntry13{
+			{
+				CertificateData: certData,
+				Extensions:      []extension.Extension{},
+			},
+		},
+	}
+
+	_, err := msg.Marshal()
+	assert.ErrorIs(t, err, errCertificateListTooLong)
+}
+
+func TestMessageCertificate13_MultipleCertsExceedingBoundary(t *testing.T) {
+	// Test multiple certs that individually are fine but together exceed the limit
+	// Two certs of ~8MB each should exceed 2^24-1 total
+	certData1 := make([]byte, 0x800000) // 8MB
+	certData2 := make([]byte, 0x800000) // 8MB
+	for i := range certData1 {
+		certData1[i] = byte(i)
+		certData2[i] = byte(i + 1)
+	}
+
+	msg := &MessageCertificate13{
+		CertificateRequestContext: []byte{},
+		CertificateList: []CertificateEntry13{
+			{
+				CertificateData: certData1,
+				Extensions:      []extension.Extension{},
+			},
+			{
+				CertificateData: certData2,
+				Extensions:      []extension.Extension{},
+			},
+		},
+	}
+
+	_, err := msg.Marshal()
+	// Total serialized = 2 * (3 + 0x800000 + 2) = 2 * 0x800005 = 0x100000A (exceeds 0xffffff)
+	assert.ErrorIs(t, err, errCertificateListTooLong)
+}
+
+func TestMessageCertificate13_CertListAtExactBoundary(t *testing.T) {
+	// Test that we can marshal a certificate_list of exactly 2^24-1 bytes
+	// We need to account for all overhead: 3 (cert prefix) + cert_data + 2 (ext prefix)
+	// So cert_data should be 0xffffff - 5
+	certData := make([]byte, 0xffffff-5)
+	for i := range certData {
+		certData[i] = byte(i)
+	}
+
+	msg := &MessageCertificate13{
+		CertificateRequestContext: []byte{},
+		CertificateList: []CertificateEntry13{
+			{
+				CertificateData: certData,
+				Extensions:      []extension.Extension{},
+			},
+		},
+	}
+
+	data, err := msg.Marshal()
+	assert.NoError(t, err)
+
+	// Verify the certificate_list is exactly at the boundary
+	// data format: [1 byte context_len][context][3 bytes cert_list_len][cert_list]
+	// context is empty, so: [0x00][3 bytes][cert_list]
+	certListLen := int(data[1])<<16 | int(data[2])<<8 | int(data[3])
+	assert.Equal(t, 0xffffff, certListLen, "certificate_list should be exactly at 2^24-1")
+
+	// Verify round-trip
+	out := &MessageCertificate13{}
+	err = out.Unmarshal(data)
+	assert.NoError(t, err)
+	assert.Equal(t, len(msg.CertificateList[0].CertificateData), len(out.CertificateList[0].CertificateData))
+}
+
 func TestMessageCertificate13_UnmarshalBufferTooSmall(t *testing.T) {
 	// Define (invalid) serialized messages (data too small)
 	tests := []struct {


pkg/protocol/handshake/message_certificate_13.go
```diff
@@ -0,0 +1,185 @@
...
+// Type returns the handshake message type.
+func (m MessageCertificate13) Type() Type {
+	return TypeCertificate
+}
+
+const (
+	handshakeMessageCertificate13ContextLengthFieldSize = 1
+	handshakeMessageCertificate13ContextMaxLength       = 255
+	handshakeMessageCertificate13CertLengthFieldSize    = 3
+	handshakeMessageCertificate13ExtLengthFieldSize     = 2
```
philipch07: Perhaps these can be named to something a little shorter? The references are rather verbose below 😅

pkg/protocol/handshake/message_certificate_13.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// Parse certificate_list
+	m.CertificateList = []CertificateEntry13{}
+	for i := 0; offset < certificateListEnd; i++ {
+		entry, newOffset, err := parseCertificate13Entry(data, offset, certificateListEnd)
+		if err != nil {
+			return err
+		}
+		offset = newOffset
+		m.CertificateList = append(m.CertificateList, entry)
```
philipch07: Currently `parseCertificat13Entry()` returns `CertificateEntry13{}` when it returns an error, but the `CertificateEntry13{}` isn't used due to the guard. Perhaps we can simplify `parseCertificat13Entry()` to `return nil, offset, err` instead?

pkg/protocol/handshake/message_certificate_request_13.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Type returns the handshake message type.
+func (m MessageCertificateRequest13) Type() Type {
+	return TypeCertificateRequest
+}
+
+const (
+	handshakeMessageCertificateRequest13ContextMaxLength = 255
+	handshakeMessageCertificateRequest13MinLength        = 3
```
philipch07: Similarly, this is maybe a bit verbose.

pkg/protocol/handshake/message_certificate_13.go
adrianosela: I don't disagree, but they are usefully explicit as there are (and there will be) a lot of files in this package.

Do you have any suggestions?

adrianosela: How about i just drop the `handshakeMessage` prefix?

philipch07: Hmm, maybe we can rename the prefix to something like `cert13`?

adrianosela: done, and `certReq13` for the other msg consts.

pkg/protocol/handshake/message_certificate_13.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// Start with certificate_request_context (1-byte length prefix)
+	out := []byte{byte(len(m.CertificateRequestContext))}
+	out = append(out, m.CertificateRequestContext...)
+
+	// Build certificate_list
+	certificateList := []byte{}
+	for _, entry := range m.CertificateList {
+		// Add cert_data as a 3-byte length prefix
+		certDataLen := len(entry.CertificateData)
```
theodorsm, line 82: We should check for boundaries `<1..2^24-1>`. 

```
if certDataLen == 0 || certDataLen > 0xffffff {
    return nil, errInvalidCertificateEntry 
}
```

pkg/protocol/handshake/message_certificate_13.go
```diff
@@ -0,0 +1,185 @@
...
+		extensionsData, err := extension.Marshal(entry.Extensions)
+		if err != nil {
+			return nil, err
+		}
+		certificateList = append(certificateList, extensionsData...)
+	}
+
+	// Add certificate_list with 3-byte length prefix
+	certificateListLenBytes := make([]byte, cert13CertLengthFieldSize)
+	util.PutBigEndianUint24(certificateListLenBytes, uint32(len(certificateList))) //nolint:gosec // G115
```
theodorsm, line 106: Should check for max length boundary <0..**2^24-1**> of `certificateList`

pkg/protocol/handshake/message_certificate_13.go
```diff
@@ -0,0 +1,185 @@
...
+	return out, nil
+}
+
+// parseCertificate13Entry parses a single certificate entry from the data.
+func parseCertificate13Entry(data []byte, offset, endOffset int) (*CertificateEntry13, int, error) {
+	// Read cert_data length
+	if offset+cert13CertLengthFieldSize > endOffset {
+		return nil, offset, errInvalidCertificateEntry
+	}
+	certDataLen := int(util.BigEndianUint24(data[offset : offset+cert13CertLengthFieldSize]))
```
theodorsm: Should check for min length boundary <**1**..2^24-1> of cert_data

pkg/protocol/handshake/message_certificate_13.go
```diff
@@ -0,0 +1,185 @@
...
+	if offset+cert13ExtLengthFieldSize > endOffset {
+		return nil, offset, errInvalidCertificateEntry
+	}
+	extensionsLen := int(binary.BigEndian.Uint16(data[offset : offset+cert13ExtLengthFieldSize]))
+
+	// Read and unmarshal extensions
+	if offset+cert13ExtLengthFieldSize+extensionsLen > endOffset {
+		return nil, offset, errInvalidCertificateEntry
+	}
+	extensionsData := data[offset : offset+cert13ExtLengthFieldSize+extensionsLen]
```
theodorsm: To be on the defensive side, we should take a copy: 

```
extensionsData := append([]byte, data[offset : offset+cert13ExtLengthFieldSize+extensionsLen]...)
// or 
extensionsData make([]byte, cert13ExtLengthFieldSize+extensionsLen)
extensionsData := copy(extensionsData, data[offset : offset+cert13ExtLengthFieldSize+extensionsLen])
```

pkg/protocol/handshake/message_certificate_13_test.go
```diff
@@ -0,0 +1,309 @@
...
+	marshalUnmarshalMessageCertificate13AndVerifyMatch(t, msg, out)
+
+	// Verify certificate is valid
+	cert, err := x509.ParseCertificate(out.CertificateList[0].CertificateData)
+	require.NoError(t, err)
+	assert.Equal(t, x509.ECDSAWithSHA256, cert.SignatureAlgorithm)
+}
+
+func TestMessageCertificate13_WithContext(t *testing.T) {
+	// Build (valid) message with multiple extensions with non empty context
```
theodorsm: Is this test supposed to have more extensions or is this comment wrong?

pkg/protocol/handshake/message_certificate_request_13.go
```diff
@@ -0,0 +1,121 @@
...
+	builder.AddUint8LengthPrefixed(func(b *cryptobyte.Builder) {
+		b.AddBytes(m.CertificateRequestContext)
+	})
+
+	// Marshal extensions (includes 2-byte length prefix, like in TLS 1.2)
+	extensionsData, err := extension.Marshal(m.Extensions)
+	if err != nil {
+		return nil, err
+	}
+	builder.AddBytes(extensionsData)
```
theodorsm, line 81: Check boundaries `<2..2^16-1>`

pkg/protocol/handshake/message_certificate_request_13.go
```diff
@@ -0,0 +1,121 @@
...
+	// Read certificate_request_context
+	var contextData cryptobyte.String
+	if !str.ReadUint8LengthPrefixed(&contextData) {
+		return errInvalidCertificateRequestContext
+	}
+	m.CertificateRequestContext = make([]byte, len(contextData))
+	copy(m.CertificateRequestContext, contextData)
+
+	// Read extensions (remaining bytes)
+	if len(str) == 0 {
```
theodorsm: `if len(str) < 2` as extensions<**2**..2^16-1>;

pkg/protocol/handshake/message_certificate_request_13_test.go
```diff
@@ -0,0 +1,188 @@
...
+					{Hash: hash.SHA256, Signature: signature.ECDSA},
+				},
+			},
+		},
+	}
+	marshalUnmarshalMessageCertificateRequest13AndVerifyMatch(t, msg)
+}
+
+func TestMessageCertificateRequest13_MultipleExtensions(t *testing.T) {
+	// Build (valid) message with multiple extensions with multiple extensions
```
theodorsm: Typo? _with multiple extensions_ twice?

pkg/protocol/handshake/message_certificate_13_test.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMessageCertificate13_Type(t *testing.T) {
+	m := &MessageCertificate13{}
+	assert.Equal(t, TypeCertificate, m.Type())
+}
+
+func TestMessageCertificate13_SingleCertNoExtensions(t *testing.T) {
```
theodorsm, line 140: I am missing an unmarshal test of a fully raw message, like in: https://github.com/pion/dtls/blob/91214624d12c49031af056799189acc3e072e750/pkg/protocol/handshake/message_certificate_test.go#L24

pkg/protocol/handshake/message_certificate_request_13_test.go
theodorsm, line 1: I am missing an unmarshal test of a fully raw certificate request message, like: https://github.com/pion/dtls/blob/91214624d12c49031af056799189acc3e072e750/pkg/protocol/handshake/message_certificate_request_test.go#L23

pkg/protocol/handshake/message_certificate_13_test.go
adrianosela: Copypasta fail - not supposed to have multiple exts

pkg/protocol/handshake/message_certificate_13.go
adrianosela: For testing the upper bound, should I allocate a 16 MB slice... or is testing the zero case sufficient you think @theodorsm ?

pkg/protocol/handshake/message_certificate_13.go
adrianosela: I added a check for non zero, for the upper bound decoding it off of 3 bytes already means its < 2^24.

pkg/protocol/handshake/message_certificate_13.go
theodorsm: I think only testing for the zero case is sufficent.

theodorsm: The actually max size should be `0xffffff - cert13CertLengthFieldSize`. I forgot about the length prefix for the vector, sorry!

pkg/protocol/handshake/message_certificate_13.go
```diff
@@ -0,0 +1,194 @@
...
+	if !str.ReadBytes(&extensionsDataAlias, int(extensionsLen)) {
+		return nil, errInvalidCertificateEntry
+	}
+
+	// Prepend the length for extension.Unmarshal (which expects it)
+	// Take defensive copy to avoid aliasing issues
+	extensionsWithLen := make([]byte, 2+len(extensionsDataAlias))
+	extensionsWithLen[0] = byte(extensionsLen >> 8)
+	extensionsWithLen[1] = byte(extensionsLen)
+	copy(extensionsWithLen[2:], extensionsDataAlias)
```
theodorsm: We can do it a bit simpler with: 

```
    // verify length of extensions
	if len(str) > 0xffff + 2 {
		return nil, errInvalidCertificateEntry
	}

	extensions, err := extension.Unmarshal(append([]byte{}, str...))
```

adrianosela: Right... also `extension.Unmarshal` copies internally, so no need for taking a copy.

adrianosela:
Nice catch on the cert's bounds

pkg/protocol/handshake/message_certificate_13.go
JoTurk: > Right... also extension.Unmarshal copies internally, so no need for taking a copy.

Our extensions are public and they need to be safe on their own. And we copy in all the new extensions.

pkg/protocol/handshake/message_certificate_13.go
JoTurk: Correct me if I'm wrong, but I think 0xFFFFFF was correct. I think from reading TLS/DTLS specs, variable-length vectors <floor..ceiling> constrain the vector contents length, and the length field is encoded separately:

>  Variable-length vectors are defined by specifying a subrange of legal
   lengths, inclusively, using the notation <floor..ceiling>.  When
   these are encoded, the actual length precedes the vector's contents
   in the byte stream.  The length will be in the form of a number
   consuming as many bytes as required to hold the vector's specified
   maximum (ceiling) length.  A variable-length vector with an actual
   length field of zero is referred to as an empty vector.

1. https://datatracker.ietf.org/doc/html/rfc8446#section-3.4

from boringssl code: https://boringssl.googlesource.com/boringssl/%2B/master/ssl/tls13_both.cc#541

adrianosela: Yeah @JoTurk is right here. The n-byte length fields are independent from the vector whose length they describe.

theodorsm: @JoTurk, @adrianosela  Yes, I agree that the length fields are independent, but to me the confusion arises from the parent vector: `CertificateEntry certificate_list<0..2^24-1>`.
As the vector holds a list of vectors, then the length prefix of any vector in this list is included in the `2^24-1` boundary no?  It then seems to be that the max length for any certificate data must be `0xffffff - cert13CertLengthFieldSize`? I would say this problem/confusion is avoided when using the cryptobyte library as it does to the upper bound checking for us. For example, this would return an error: 
```
	certificateData := make([]byte, 0xffffff)
	for i := range certificateData {
		certificateData[i] = byte(i)
	}
	var builder cryptobyte.Builder
	builder.AddUint24LengthPrefixed(func(b *cryptobyte.Builder) {
		b.AddUint24LengthPrefixed(func(b *cryptobyte.Builder) {
			fmt.Printf("Length of certData: %x\n", len(certificateData))
			b.AddBytes(certificateData)
			fmt.Printf("Length of certData vector with prefix: %x\n", len(b.BytesOrPanic()))
		})
		fmt.Printf("Length of certificate_list vector with prefix: %x\n", len(b.BytesOrPanic()))
	})
	_, err := builder.Bytes()
	if err != nil {
		fmt.Println(err.Error())
	}

// Output: 
Length of certData: ffffff
Length of certData vector with prefix: 1000002
Length of certificate_list vector with prefix: 1000005
cryptobyte: pending child length 16777218 exceeds 3-byte length prefix
```

JoTurk: I agree with @adrianosela  points about the extension size.

Also in isolation we shouldn't check for the list bound here. because if we do, we'll return errInvalidCertificateEntry when an entry is only out of bounds because it causes the parent list to exceed its limit, in that case we should return errCertificateListTooLong, which is already checked at the end of the loop. So subtracting -3 or -5 here is really a logic / spec bug.

theodorsm (approved):
I agree with the conclusion about the upper bound checking, and thanks for the discussion about it!

This PR looks really good now, especially with all the good tests. 

@adrianosela thanks for working on this and the quick iterations :100:.

PR #786, opened 2026-01-22
Introduce options patterns

### LINKED ISSUE #765: Config pattern
This is a tracking issue for deprecating the Config struct and introducing the config pattern to `pion/dtls`. This should also make it straightforward to configure DTLS 1.3. #731 

In 4.1 and 4.2, we updated mdns, RTP, and ICE. and for the next release we plan to update SCTP and DTLS.

JoTurk:
#### Description
(not ready for review, I'll double review this when i wake up).
This introduces options patterns to DTLS. it's a bit tricky because dtls options are a bit complex:
1. Like ICE we apply the defaults first, then we do validations in the options calls.
2. Like MDNS we have client and servers options, and overlapping options between the two.
3. like transport we're introducing `WithOptions` helpers, and in the next major release we're going to depreciate this suffix, and change the current constructors to options first.

Updated all examples and the e2e test.

resolves: #765

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -99,6 +99,9 @@ type Conn struct {
 	handshakeConfig *handshakeConfig
 }
 
+// createConn creates a new DTLS connection.
+// Caller is responsible for validating the config before calling this function.
+//
 //nolint:cyclop
 func createConn(
 	nextConn net.PacketConn,
@@ -107,10 +110,6 @@ func createConn(
 	isClient bool,
 	resumeState *State,
 ) (*Conn, error) {
-	if err := validateConfig(config); err != nil {
-		return nil, err
-	}
-
 	if nextConn == nil {
 		return nil, errNilNextConn
 	}
@@ -317,6 +316,8 @@ func (c *Conn) HandshakeContext(ctx context.Context) error {
 }
 
 // Dial connects to the given network address and establishes a DTLS connection on top.
+//
+// Deprecated: Use DialWithOptions instead.
 func Dial(network string, rAddr *net.UDPAddr, config *Config) (*Conn, error) {
 	// net.ListenUDP is used rather than net.DialUDP as the latter prevents the
 	// use of net.PacketConn.WriteTo.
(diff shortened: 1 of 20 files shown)


JoTurk:
Okay i reviewed this few times and i think it's consistent with our other options patters, I'll add more coverage before merge but i think this is ready for review.


@philipch07 @adrianosela @theodorsm

conn.go
```diff
@@ -107,10 +107,6 @@ func createConn(
 	isClient bool,
 	resumeState *State,
 ) (*Conn, error) {
-	if err := validateConfig(config); err != nil {
```
adrianosela, line 110: nit: consider adding a comment above this function, and `serverWithConfig` stating that callers are responsible for config validation.

JoTurk: sounds good, we moved the validation to the public functions Server and Client, in the options API we do the validation for each option per option call, but yeah it's a good idea to add a comment here.

options.go
```diff
@@ -0,0 +1,630 @@
...
+func (o sharedOption) applyClient(c *dtlsConfig) error { return o(c) }
+
+// WithCertificates sets the certificate chain to present to the other side of the connection.
+// For functional options, an explicitly empty slice is not allowed.
+func WithCertificates(certs ...tls.Certificate) Option {
+	return sharedOption(func(c *dtlsConfig) error {
+		if len(certs) == 0 {
+			return errEmptyCertificates
+		}
+		c.certificates = append([]tls.Certificate(nil), certs...)
```
adrianosela: nit: maybe we can make it more obvious that these are defensive copies? e.g.

```
// defensiveCopy copies a slice. This prevents the caller from mutating
// the config after construction. Returns empty slice if input is empty.
func defensiveCopy[T any](t ...T) []T {
	return append([]T{}, t...)
}
```

Just an idea, but feel free to ignore, append([]T(nil), slice...) is a well-known idiom anyways.

adrianosela (approved):
This is amazing - 🚢 it!

options.go
JoTurk: I agree. I'll add this.

JoTurk: This will be the first use of generics in pion btw xd.

options.go
```diff
@@ -0,0 +1,636 @@
...
+		ConnectionIDGenerator:         c.connectionIDGenerator,
+		PaddingLengthGenerator:        c.paddingLengthGenerator,
+		HelloRandomBytesGenerator:     c.helloRandomBytesGenerator,
+		ClientHelloMessageHook:        c.clientHelloMessageHook,
+		ServerHelloMessageHook:        c.serverHelloMessageHook,
+		CertificateRequestMessageHook: c.certificateRequestMessageHook,
+		OnConnectionAttempt:           c.onConnectionAttempt,
+	}
+
+	if len(c.certificates) > 0 {
```
adrianosela, line 126: I just noticed that we are defensively copying twice. Not a big deal - but we don't need to AFAICT. (we are doing a copy when setting the option on `dtlsConfig`, so no need to copy again here)

cc:// @JoTurk

JoTurk: I think we should remove the other copy, and make it copy for the old API, i'll make a PR for more cleanup and to update the tests. hopefully we can remove the other API sooner so we don't have to maintain both.

options.go
JoTurk: update: it didn't last more than ~6 months, back to 0 templates. https://github.com/pion/dtls/pull/956

PR #793, opened 2026-02-05
Add DTLS 1.3 Unified Header

### LINKED ISSUE #755: DTLS 1.3 record layer encoding
This issue aims to implement [RFC9147 section 4](https://datatracker.ietf.org/doc/html/rfc9147#name-the-dtls-record-layer).

theodorsm:
#### Description

This PR implements the new DTLS 1.3 Unified Header that is part of the new record layer encoding.

https://datatracker.ietf.org/doc/html/rfc9147#name-the-dtls-record-layer:
```
    0 1 2 3 4 5 6 7
    +-+-+-+-+-+-+-+-+
    |0|0|1|C|S|L|E E|
    +-+-+-+-+-+-+-+-+
    | Connection ID |   Legend:
    | (if any,      |
    /  length as    /   C   - Connection ID (CID) present
    |  negotiated)  |   S   - Sequence number length
    +-+-+-+-+-+-+-+-+   L   - Length present
    |  8 or 16 bit  |   E   - Epoch
    |Sequence Number|
    +-+-+-+-+-+-+-+-+
    | 16 bit Length |
    | (if present)  |
    +-+-+-+-+-+-+-+-+
```

#### Reference issue
Partly fixes #755


pkg/protocol/recordlayer/header.go
```diff
@@ -52,10 +63,28 @@ func (h *Header) Marshal() ([]byte, error) {
 
 // Unmarshal populates a TLS RecordLayer Header from binary.
 func (h *Header) Unmarshal(data []byte) error {
-	if len(data) < FixedHeaderSize {
+	if len(data) < 1 {
 		return errBufferTooSmall
 	}
+
 	h.ContentType = protocol.ContentType(data[0])
+	if protocol.IsDTLS13Ciphertext(h.ContentType) {
```
theodorsm, line 71: Hmm, we should also check if we have enabled DTLS 1.3 before parsing and returning something that would be unvalid for DTLS 1.2. Tied to TODOs in #738

theodorsm: Maybe it's enought with a comment for the function, considering the returned Header has `h.Version = protocol.Version1_3`

JoTurk:
Can you please look at the AI's comments, I think some of them are worth fixing, I'll re read the rfcs and the code and review all your pending prs, sorry it might take a bit because I caught a fever :)

pkg/protocol/recordlayer/header_13.go
```diff
@@ -0,0 +1,132 @@
...
+		contentType |= UnifiedHeaderSeqBit
+		head.AddUint16(u.SequenceNumber)
+	} else {
+		head.AddUint8(uint8(u.SequenceNumber)) //nolint:gosec
+	}
+
+	if u.Length > 0 {
+		contentType |= UnifiedHeaderLengthBit
+		head.AddUint16(u.Length)
+	}
```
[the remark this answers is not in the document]
theodorsm: The length can be controlled by not initializing Length or by setting it to 0 explicitly, there is no need for a seperate boolean.

pkg/protocol/recordlayer/header.go
```diff
@@ -52,10 +63,28 @@ func (h *Header) Marshal() ([]byte, error) {
...
+		if err != nil {
+			return err
+		}
+		h.Version = protocol.Version1_3
+		h.Epoch = uint16(uh.EpochLow)
+		h.SequenceNumber = uint64(uh.SequenceNumber)
+		h.ConnectionID = uh.ConnectionID
+		h.ContentLen = uh.Length
+
+		return nil
```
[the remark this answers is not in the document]
theodorsm: If a unified header is parsed, Version 1.3 is set. This will be an indicator callers can use to handle length correctly. If the length is 0, this means that the record consumes the entire rest of the datagram in the lower level transport. Since we now also calculate the correct variable length of the header with `Size()`, the content length can be derived.

pkg/protocol/recordlayer/errors.go
```diff
@@ -16,9 +16,22 @@ var (
...
+	errUnsupportedProtocolVersion = &protocol.FatalError{
+		Err: errors.New("unsupported protocol version"), //nolint:err113
+	}
+	errInvalidContentType = &protocol.TemporaryError{
+		Err: errors.New("invalid content type"), //nolint:err113
+	}
+	errCIDTooBig = &protocol.InternalError{
+		Err: errors.New("connection ID size is too big"), //nolint:err113
+	}
+	errInvalidUnifiedHeaderFormat = &protocol.FatalError{
```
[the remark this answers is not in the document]
theodorsm: Decode errors should be fatal: https://datatracker.ietf.org/doc/html/rfc8446#section-6.2

pkg/protocol/recordlayer/header_13.go
```diff
@@ -0,0 +1,132 @@
...
+
+// Unmarshal populates a DTLS 1.3 Unified Header from binary.
+func (u *UnifiedHeader) Unmarshal(data []byte) error {
+	str := cryptobyte.String(data)
+
+	var ct uint8
+	if !str.ReadUint8(&ct) || !protocol.IsDTLS13Ciphertext(protocol.ContentType(ct)) {
+		return errInvalidContentType
+	}
+
```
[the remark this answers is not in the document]
theodorsm: ConnectionID should not be cleared as it should be initialized with a byte buffer with expected length.

pkg/protocol/recordlayer/header.go
```diff
@@ -52,10 +65,30 @@ func (h *Header) Marshal() ([]byte, error) {
...
+	if len(data) < 1 {
 		return errBufferTooSmall
 	}
+
 	h.ContentType = protocol.ContentType(data[0])
+	if protocol.IsDTLS13Ciphertext(h.ContentType) {
+		var uh UnifiedHeader
+		uh.ConnectionID = make([]byte, len(h.ConnectionID))
+		err := uh.Unmarshal(data)
+		if err != nil {
```
JoTurk: My concern is related to my comment in the other PR, I think this should handled based on negotiated version or and state/epoch in mixed mode, rather than trying to assume based on the header itself, Also i think we should have explicit 1.2 and 1.3 receive paths (but i think this is out of scope of this PR, https://github.com/pion/dtls/blob/fd27a52a64265097a621432090926a651109b4cb/conn.go#L823 but this is more https://github.com/pion/dtls/issues/755 )

I think we should do something similar to boringssl's use_dtls13_record_header. https://pigweed.googlesource.com/third_party/github/google/boringssl/%2B/HEAD/ssl/dtls_record.cc

theodorsm: I agree about having more explicitness in the paths. I have removed this check in the DTLS 1.2 header. Let's add an explicit RecordLayer13 that uses the UnifiedHeader for ciphertext (in a different PR)

pkg/protocol/recordlayer/header_13.go
```diff
@@ -0,0 +1,149 @@
...
+		head.AddBytes(u.ConnectionID)
+	}
+
+	if u.SequenceNumber > math.MaxUint8 {
+		contentType |= UnifiedHeaderSeqBit
+		head.AddUint16(u.SequenceNumber)
+	} else {
+		head.AddUint8(uint8(u.SequenceNumber)) //nolint:gosec
+	}
+
```
JoTurk: > The S bit (0x08) indicates the size of the sequence number. 0 means an 8-bit sequence number, 1 means 16-bit. Implementations MAY mix sequence numbers of different lengths on the same connection.

I think it's a good idea t store the S value in the struct and rely on it for setting the sequence size rather than the actual value, e.g unmarshal and marshal should round-trippable (if we he hold the S value in the struct), right now the user of this strict can't force it to be so..

pkg/protocol/recordlayer/header_13.go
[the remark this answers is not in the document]
JoTurk: Similar comment to S bit i guess.

JoTurk (approved):
Thank you, sorry i was taking a short break, before things really start, I read the spec few times, my comments are nits feel free to ignore them but I think they are kinda important for tests, and for long term. so we don't have to break the api in the future.

Thank you and sorry again!

pkg/protocol/recordlayer/header_13_test.go
```diff
@@ -0,0 +1,163 @@
...
+package recordlayer
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUnifiedHeader(t *testing.T) {
+	uh := UnifiedHeader{SequenceNumber: 0xaabb, SeqBit: true, Length: 42, LengthBit: true, EpochLow: 15}
```
theodorsm, line 13: The usage of the UnifiedHeader is now very explicit and I would argue it's easier to forget to set the bit flags (we just have to be aware). But I guess it's worth it with the control and round trip correctness.

Interestingly, boringssl only uses 16-bit seq numbers: https://pigweed.googlesource.com/third_party/github/google/boringssl/%2B/HEAD/ssl/dtls_record.cc#541

theodorsm:
@JoTurk, thanks for the review! No worries, I have been out sick this week. 

I have added the bits to the struct, but now the interface is very explicit. We can try this interface out and see if it becomes messy, if it does, we can move the bits to be implicit again.

pkg/protocol/recordlayer/header_13_test.go
JoTurk: I wonder if we should do the same as boringssl.

JoTurk:
@theodorsm Sorry about this, I hope you're doing better.
Yeah we can test it for a bit and change anything if we don't like it, since we're not tagging from main until 1.3 is ready :)

PR #820, opened 2026-03-28
Add optimizations for connection & handshake

noboruma:
#### Description
Add `MarshalInto` functions to allow for buffer pre-allocation of headers
Add pooling around timers
Avoid unnecessary copies in compact scenarios where the packets are already MTU size
Add cheap cache for MessageCertificate
Avoid unnecessary copy/allocations in Content ApplicationData

~~On top of marhsall revamp I am also proposing:~~
~~- Timer pooling for handshaker (12 only)~~
~~- Multi-sockets listener support~~
~~- acceptFilter based on net.Addr~~


noboruma:
@theodorsm @JoTurk I have updated the API
Any blocker on your side? At our scale (200k DTLS connections) we see a huge difference with this change. CPU and RAM usage are benefiting a lot here, as you can see here:
Screenshot 2026-04-10 at 22 06 33

^ Here is a 20h run. On the left the stable piece is with the proposed changes, on the right after 12h is when using `origin/main` directly (there is 1 reboot in between).

JoTurk:
Direction looks good to me, thank you so much for up streaming this btw. I'll need to review the details tho.
Can you fix the failing tests?

noboruma:
@JoTurk I have fixed the tests, only the API break is being reported but this one should be ignored?

JoTurk:
> I have fixed the tests, only the API break is being reported but this one should be ignored?

No sadly, we can't break the API yet (We plan to do it with DTLS 1.3 is ready), we'll have to do a work around. or introduce another API.

JoTurk:
@noboruma Just a heads-up, we've started introducing breaking changes on main for DTLS 1.3, and we won't be cutting tags from main anytime soon until v4. If you're fine waiting, you can keep the breaking API and proceed with review and merging into main.

If you'd prefer to get this feature into v3 now instead, feel free to re-target the PR to the dtls-1.2 branch.

thank you.

JoTurk:
sorry for all the conflicts, I'll fix this and get it merged for v4.

noboruma:
@JoTurk I have rebased on top of `v3`
I actually have a bunch more of optimizations for `v3` that I would like to push there as well

JoTurk:
@noboruma sounds good, if it targets v3 we can't break the API so the `API / check / Compatibility` must pass, we can have another API in parallel and in v4 we keep the optimized API (I didn't really review the PR yet).

Sean-Der:
@noboruma I am sorry this got dropped in all the work with 1.3 ?

Would you be ok with me merging this into /v4 ? I will take care of fixing everything. Backporting to /v3 I am hesitant on because we just don't have enough people to maintain multiple major releases :(

Sean-Der:
Also would you be open to doing a blogpost about your performance work? It would be a great write up/people could learn a bunch from what you have done.

Also would love to give you credit for the work you have done :) 

'Making Pion Go Fast'... something like that.

JoTurk:
@Sean-Der he talked to me in discord about rebasing this for v4.

Sean-Der:
Nice, wanna make this part of v4 announcement?

if we can also release with significant performance how cool is thst :)

if you don’t have time happy to take over PR!

JoTurk:
he originally wanted to get this merged with v3 this is why it took too long but since we're releasing v4 it would be easier to rebase it and remove the old API. @noboruma we are likely going to release v4 this week.

JoTurk:
@Sean-Der @noboruma i think getting this merged (and the old API removed) should be blocker for v4, and yes it should be part of the announcement :)

JoTurk:
image
This is a common Github bug, I'll close and open the pr to fix the CI.

internal/net/udp/packet_conn.go
```diff
@@ -23,6 +23,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/noboruma/gensync"
+	"github.com/noboruma/gensync/gensync"
```
JoTurk: Sorry we would like to avoid depending on non-stdlib libraries (We apply this policy to all of our libraries). 

for many reasons, mainly because we usually support really old Go versions (there is already a version mismatch), and we like to be able to modify and update anything in our libraries runtime code.

pkg/protocol/handshake/message_server_key_exchange.go
```diff
@@ -36,40 +36,94 @@ func (m MessageServerKeyExchange) Type() Type {
...
+	}
+
+	offset := 0
 	if m.IdentityHint != nil {
-		out = append([]byte{0x00, 0x00}, m.IdentityHint...)
-		binary.BigEndian.PutUint16(out, uint16(len(out)-2)) //nolint:gosec //G115
+		binary.BigEndian.PutUint16(out[offset:], uint16(len(m.IdentityHint))) //nolint:gosec //G115
+		offset += 2
+		n := copy(out[offset:], m.IdentityHint)
+		offset += n
```
JoTurk, line 87: This can panic if out is empty, which can happen with an invalid `MessageServerKeyExchange` since MarshalTo can return 0.

pkg/protocol/ack.go
```diff
@@ -29,18 +31,34 @@ func (a ACK) ContentType() ContentType {
...
-		}
-	})
+	return out, err
+}
+
+// MarshalTo encodes the Handshake.
+func (a *ACK) MarshalTo(out []byte) (int, error) {
+	n := 0
+	out[0] = byte(16 * len(a.Records) >> 8) //nolint:gosec // G115
+	out[1] = byte(16 * len(a.Records))      //nolint:gosec // G115
```
JoTurk: This can panic too.

pkg/protocol/return_routability_check.go
```diff
@@ -31,12 +31,24 @@ func (r ReturnRoutabilityCheck) ContentType() ContentType {
...
+
+	return out, err
+}
+
+// MarshalTo encodes to wire format into a pre-allocated buffer.
+func (r *ReturnRoutabilityCheck) MarshalTo(out []byte) (int, error) {
 	out[0] = byte(r.MessageType)
 
-	return append(out, r.Cookie[:]...), nil
+	return copy(out[1:], r.Cookie[:]), nil
```
JoTurk: ~~This can panic~~

pkg/protocol/application_data.go
```diff
@@ -24,6 +24,18 @@ func (a *ApplicationData) Marshal() ([]byte, error) {
 	return a.Data, nil
 }
 
+// MarshalTo encodes the ApplicationData to binary into a pre-allocated buffer.
+func (a *ApplicationData) MarshalTo(out []byte) (int, error) {
+	copy(out, a.Data)
+
+	return len(a.Data), nil
```
JoTurk: This will silently truncate if the buffer is small, and report len(a.data).

pkg/protocol/handshake/message_certificate_verify.go
```diff
@@ -31,27 +31,38 @@ func (m MessageCertificateVerify) Type() Type {
...
-		return nil, dtlserrors.ErrInvalidSignHashAlgorithm
+		return 0, dtlserrors.ErrInvalidSignHashAlgorithm
 	}
 
-	out := make([]byte, 1+1+2+len(m.Signature))
-
-	out[0] = byte(m.HashAlgorithm)
-	out[1] = byte(m.SignatureAlgorithm)
+	out[0] = byte(m.HashAlgorithm)                                //nolint:gosec // G115
+	out[1] = byte(m.SignatureAlgorithm)                           //nolint:gosec // G115
```
JoTurk, line 64: This will panic.

JoTurk (commented):
I think for all the similar comments we'll need to just make sure we have MarshalSize  and zero check before enter any MarshalTo.

pkg/protocol/return_routability_check.go
JoTurk: Sorry line 49 can panic, but this line will return incorrect value, we write 1 + cookie but we only return the result of copying cookie.

pkg/protocol/handshake/message_client_hello.go
```diff
@@ -38,46 +40,95 @@ func (m MessageClientHello) Type() Type {
 	return TypeClientHello
 }
 
+func (m *MessageClientHello) cacheMarshalExtensions() error {
+	if m.marchalledExtensions == nil && m.marchalledExtensionsErr == nil {
+		m.marchalledExtensions, m.marchalledExtensionsErr = extension.MarshalList(m.Extensions)
```
JoTurk: the `Extensions` field is public, if the caller change it, the cache will be incorrect, we should either evaluate we can change the extension field to getters and setters (that updates the cache), (this applies to all the similar cache). 

sadly, we have to mutate extensions in many places:

https://github.com/pion/dtls/blob/dccb7be0a6329aebb65b4d4d7d9d5d719b6698f8/internal/negotiation/retry.go#L104

JoTurk:
I don't see any other issues other than the few comments, I think this is ready to merge once the issues are fixed, thank you a lot

pkg/protocol/return_routability_check.go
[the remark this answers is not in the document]
JoTurk: duplicated comment.

pkg/protocol/application_data.go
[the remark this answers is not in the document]
JoTurk: duplicated comment.

pkg/protocol/ack.go
```diff
@@ -29,18 +31,34 @@ func (a ACK) ContentType() ContentType {
...
-			b.AddUint64(rec.SequenceNumber)
-		}
-	})
+	return out, err
+}
+
+// MarshalTo encodes the Handshake.
+func (a *ACK) MarshalTo(out []byte) (int, error) {
+	n := 0
+	out[0] = byte(16 * len(a.Records) >> 8) //nolint:gosec // G115
```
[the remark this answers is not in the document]
JoTurk: Also duplicated comment, I wish copilot was able to see if someone commented on the same issue.

JoTurk:
(I'm sorry about all the duplicated copilot comments, it doesn't seem to be able to look if someone reviewed the same issues)

pkg/protocol/recordlayer/header.go
```diff
@@ -86,7 +100,7 @@ func (h *Header) Unmarshal(data []byte) error {
 	return nil
 }
 
-// Size returns the total size of the header.
-func (h *Header) Size() int {
+// MarshalSize returns the total size of the header.
+func (h *Header) MarshalSize() int {
 	return FixedHeaderSize + len(h.ConnectionID)
 }
```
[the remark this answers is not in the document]
noboruma: Here I believe `Size()` is self-contained in this repo - it was renamed everywhere and therefore it should be safe

pkg/protocol/recordlayer/header_13.go
```diff
@@ -139,7 +139,7 @@ func (u *UnifiedHeader) Unmarshal(data []byte) error {
 	return nil
 }
 
-func (u *UnifiedHeader) Size() int {
+func (u *UnifiedHeader) MarshalSize() int {
```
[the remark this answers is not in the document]
noboruma: Same as previous comment

pkg/protocol/recordlayer/header.go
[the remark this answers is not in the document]
JoTurk: ^ I agree.

JoTurk (approved):
I think this is ready to merge now.
1.  Applied the optimization to outbound record construction (this removed many allocations) and DTLS 1.3 ciphertext.
2. removed redundant copies.
3. removed some redundant allocations (extension length is now allocation free, ClientHello had a few temporary allocations, some other...)
4. replaced the extensions API with a Getter / Setter so that the extension cache is cleaner.
5. restored the AEAD nonce-buffer pooling that was removed by mistake.
6. Added some validations and made sure we don't allocate when marshal fails.
7. I fixed a few panic and truncation issues. Bound checks, Partial marshaling...

(corrected some marshals and some minor issues)

Sean-Der:
@JoTurk LGTM! Things I noticed (but don't feel strongly)

* MarshalTo returns partial results to users. If we hit an error it doesn't just bail. Like [this](https://github.com/pion/dtls/blob/21aefa02789c8fe2f43e8dcea890e5d149032e87/pkg/protocol/handshake/message_server_hello.go#L77-L83) I would just return an error right away and `0`. Otherwise users might accidentally see non-zero and think buffers are ok

```
  - MessageServerHello.MarshalTo: writes prepared prefix despite error (https://github.com/pion/dtls/blob/21aefa02789c8fe2f43e8dcea890e5d149032e87/pkg/protocol/handshake/message_server_hello.go#L74-L84)
  - MessageClientHello.MarshalTo: same partial-write branch (https://github.com/pion/dtls/blob/21aefa02789c8fe2f43e8dcea890e5d149032e87/pkg/protocol/handshake/message_client_hello.go#L80-L90)
  - MessageEncryptedExtensions.MarshalTo: marshals the prepared prefix after preparation fails (https://github.com/pion/dtls/blob/21aefa02789c8fe2f43e8dcea890e5d149032e87/pkg/protocol/handshake/message_encrypted_extensions.go#L54-L67)
  - PrepareList: returns previously prepared extensions with an error (https://github.com/pion/dtls/blob/21aefa02789c8fe2f43e8dcea890e5d149032e87/pkg/protocol/extension/raw.go#L123-L155)
  - MarshalListTo: returns the current offset when a later extension fails (https://github.com/pion/dtls/blob/21aefa02789c8fe2f43e8dcea890e5d149032e87/pkg/protocol/extension/raw.go#L211-L247)
  - Handshake.MarshalTo: adds the child’s partial byte count to the header count (https://github.com/pion/dtls/blob/21aefa02789c8fe2f43e8dcea890e5d149032e87/pkg/protocol/handshake/handshake.go#L156-L178)
```

* Getter/Setter indirection. I bet Go will optimized it away, but adds more code and uncertainty. Unless you want to do validation in them in the future? I think it would make sense then. I only noticed because of lots of changes around withExtensions etc...

JoTurk:
@Sean-Der 
- I added the partial write to fix a bug where parent serializers such as Handshake.MarshalTo don't aggregate child counts, and miss writes (this fixes an awkward bug in hello marshal).
- I added the Setter / Getter to remove "CachedExtensions and Extensions" it was awkward that users were able to modify extensions cache manually (earlier version of the PR had extension cache that didn't update when we set Extensions), I forgot to remove the Setter / Getter when I moved this cache at a different layer, I'll remove them.

PR #823, opened 2026-04-13
Add min and max version API

### LINKED ISSUE #731: DTLS 1.3 config flag

theodorsm:
#### Description

This PR adds a private API for setting the minimum and max version of DTLS, similar to how it is done in [`crypto/tls`](https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/crypto/tls/common.go;l=791): 

```
    // MinVersion contains the minimum TLS version that is acceptable.
    //
    // By default, TLS 1.2 is currently used as the minimum. TLS 1.0 is the
    // minimum supported by this package.
    //
    // The server-side default can be reverted to TLS 1.0 by including the value
    // "tls10server=1" in the GODEBUG environment variable.
    MinVersion uint16

    // MaxVersion contains the maximum TLS version that is acceptable.
    //
    // By default, the maximum version supported by this package is used,
    // which is currently TLS 1.3.
    MaxVersion uint16
```

The API should be kept private until the DTLS 1.3 MVP implementation is done.

#### Reference issue
Fixes #731

### DIFF
--- a/options.go
+++ b/options.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/pion/dtls/v3/pkg/crypto/elliptic"
+	"github.com/pion/dtls/v3/pkg/protocol"
 	"github.com/pion/dtls/v3/pkg/protocol/handshake"
 	"github.com/pion/logging"
 )
@@ -80,6 +81,8 @@ type dtlsConfig struct { //nolint:dupl
 	certificateRequestMessageHook func(handshake.MessageCertificateRequest) handshake.Message
 	onConnectionAttempt           func(net.Addr) error
 	listenConfig                  net.ListenConfig
+	minVersion                    protocol.Version
+	maxVersion                    protocol.Version
 }
 
 // applyDefaults applies default values to the config.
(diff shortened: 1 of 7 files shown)


options.go
```diff
@@ -561,6 +566,34 @@ func WithClientHelloMessageHook(fn func(handshake.MessageClientHello) handshake.
...
+	return sharedOption(func(c *dtlsConfig) error {
+		if protocol.IsSupportedVersion(version) {
+			c.maxVersion = version
+
+			return nil
+		}
+
+		return errUnsupportedProtocolVersion
+	})
+}
```
JoTurk, line 595: Maybe we should export this as a single a single option (optional? min, optional? max)

JoTurk (approved):

options.go
daenney: Doing a quick Github search, `tls.MinVersion` for Go results in about 1.2k hits, `MaxVersion` in a little less than half of that.

In my own experience, it's also relatively rare to want to set a ceiling on the supported protocols, whereas a floor is fairly common.

I would suggest keeping the options separate.

JoTurk: Fair point, I agree. 

(I was thinking about WebRTC and other users that might not want DTLS 1.3 because of lack of support, fingerprinting and bugs like this https://bugzilla.mozilla.org/show_bug.cgi?id=1314849 (solved now). Also I was thinking about exporting a smaller API surface).

Welcome back :)

theodorsm:
Thanks! Good to see you around again @daenney :)

PR #829, opened 2026-04-22
DTLS 1.3: Add flights 0 and 1

### LINKED ISSUE #825: DTLS 1.3 HelloRetryRequest and cookie exchange
This issue aims to track the implementation efforts of the full DTLS 1.3 handshake defined in [RFC9147 Section 5.7](https://datatracker.ietf.org/doc/html/rfc9147#section-5.7)

  - [x] Generate and parse HelloRetryRequest
  - [x] Apply the synthetic message_hash transcript transition
  - [x] Generate the second ClientHello with the cookie
  - [x] Reject a second HelloRetryRequest
  - [x] Add an end-to-end client/server HelloRetryRequest test
  - [x] Test invalid, missing, and mismatched cookies
  - [x] Test retransmission and packet reordering around the cookie exchange
  - [x] Verify application data after the retried handshake

theodorsm:
#### Description

This PR aims to implement flight 0 (waiting server) and flight 1 (sending the first CH) for DTLS 1.3.

Blocked by #738  (copied code / commits)

#### Reference issue
Partly fixes #825


flighthandlers_client_13.go
```diff
@@ -0,0 +1,171 @@
...
+			Group: keypair.Curve, KeyExchange: keypair.PublicKey,
+		})
+	}
+	state.localKeyEntries = entries
+	extensions = append(extensions, &extension.KeyShare{
+		ClientShares: entries,
+	})
+
+	extensions = append(extensions, &extension.SupportedVersions{
+		Versions: []protocol.Version{protocol.Version1_3, protocol.Version1_2},
```
theodorsm: This should reflect config

flighthandlers_client_13.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	// connection ID
+
+	// Pre_shared_key must be last extension
+
+	clientHello := &handshake.MessageClientHello{
+		Version:   protocol.Version1_2,
+		SessionID: state.SessionID,
+		Cookie:    state.cookie,
```
theodorsm: Should explicitly set to be empty.

>legacy_cookie:
>A DTLS 1.3-only client MUST set the legacy_cookie field to zero length. If a DTLS 1.3 ClientHello is received with any other value in this field, the server MUST abort the handshake with an "illegal_parameter" alert.

JoTurk:
@theodorsm I'm going to rebase this and the rest of the flight PRs now that FMS is merged.

JoTurk:
> @theodorsm I'm going to rebase this and the rest of the flight PRs now that FMS is merged.

Done! Now I'm going to review it and merge or add commits.

flighthandler_13.go
```diff
@@ -9,26 +9,37 @@ import (
...
 
 //nolint:unused
 func (f flightVal13) getFlightParser13() (flightParser13, error) {
-	return nil, errFlightUnimplemented13
+	switch f {
+	case flight13_0:
+		return flight13_0Parse, nil
+	default:
+		return nil, errFlightUnimplemented13
+	}
```
JoTurk, line 38: I'm implementing wait/parse (after this is merged).

flighthandlers_server_13.go
```diff
@@ -24,3 +38,177 @@ package dtls
...
+		if !slices.Contains(state.remoteGroups, entry.Group) {
+			return 0, &alert.Alert{Level: alert.Fatal, Description: alert.IllegalParameter}, errInvalidGroupInKeyShare
+		}
+		groups = append(groups, entry.Group)
+	}
+	state.namedCurve, _ = findMatchingGroup(groups, cfg.ellipticCurves)
+
+	if cfg.insecureSkipHelloVerify {
+		nextFlight = flight13_4
+	}
```
JoTurk: we should generate key pair after selecting group. I'm fixing this.

flighthandlers_client_13.go
```diff
@@ -34,13 +37,82 @@ import (
...
+	for _, val := range sh.Extensions {
+		switch ext := val.(type) {
+		case *extension.SupportedVersions:
+			// nolint:godox
+			// TODO: negotiate version
+			state.remoteVersions = ext.Versions
+		case *extension.CookieExt:
+			state.cookie = ext.Cookie
+		case *extension.KeyShare:
+			state.remoteKeyEntries = ext.ClientShares
```
JoTurk: this should be selected group.

flighthandlers_server_13.go
```diff
@@ -24,3 +38,177 @@ package dtls
...
+
+			continue
+		}
+		if c := cipherSuiteForID(CipherSuiteID(id), cfg.customCipherSuites); c != nil {
+			cipherSuites = append(cipherSuites, c)
+		}
+	}
+
+	// nolint:godox
+	// TODO: check for DTLS 1.3 cipher suites
```
JoTurk, line 98: Going to finish this todo and filter out non-dtls 1.3 ciphers (in another PR).

JoTurk (changes requested):

flighthandlers_client_13.go
```diff
@@ -34,6 +37,78 @@ import (
...
+		case *extension.KeyShare:
+			if ext.SelectedGroup != nil {
+				state.remoteKeyEntries = []extension.KeyShareEntry{
+					{Group: *ext.SelectedGroup},
+				}
+			}
+		}
+	}
+
+	state.handshakeRecvSequence = seq
```
JoTurk, line 142: This is part of the dual-state issue, and ik it's a todo but we should reject HRR without the expected selected version.

flighthandlers_server_13.go
```diff
@@ -24,3 +38,181 @@ package dtls
...
+	for _, entry := range state.remoteKeyEntries {
+		// Clients MUST NOT offer any KeyShareEntry values
+		// for groups not listed in the client's "supported_groups" extension.
+		// Servers MAY check for violations of these rules and abort the
+		// handshake with an "illegal_parameter" alert if one is violated.
+		if !slices.Contains(state.remoteGroups, entry.Group) {
+			return 0, &alert.Alert{Level: alert.Fatal, Description: alert.IllegalParameter}, errInvalidGroupInKeyShare
+		}
+		groups = append(groups, entry.Group)
+	}
```
JoTurk: Multiple problems here, we should build the groups from the overlap of server/client groups.
And if there is a common group but non usable initial share we should send HelloRetryRequest.

>  the client has not provided a sufficient "key_share" extension
   (e.g., it includes only DHE or ECDHE groups unacceptable to or
   unsupported by the server), the server corrects the mismatch with a
   HelloRetryRequest and the client needs to restart the handshake with
   an appropriate "key_share" extension, as shown in Figure 2.  If no
   common cryptographic parameters can be negotiated, the server MUST
   abort the handshake with an appropriate alert

and

> This vector MAY be empty if the client is requesting a
   HelloRetryRequest.  Each KeyShareEntry value MUST correspond to a
   group offered in the "supported_groups" extension and MUST appear in
   the same order.  However, the values MAY be a non-contiguous subset
   of the "supported_groups" extension and MAY omit the most preferred
   groups.  Such a situation could arise if the most preferred groups
   are new and unlikely to be supported in enough places to make
   pregenerating key shares for them efficient.

theodorsm: Very good point. I changed the group logic here.

HRR will always be sent unless explicilty skipped with `insecureSkipHelloVerify` (should we keep this flag for DTLS 1.3?) as we need to send a cookie to the client.

I updated the HRR flight generation to check if a `KeyShare` has been recieved (HRR MUST not contain ext included in the CH), this will cover both the empty vector case and failed groups negotation: https://github.com/pion/dtls/pull/832/changes#diff-ac0226cf6f69c71886a0d5c61981a0e1386b2f94df9b4bc05e106cf10015d7c5R82

I am unsure how to takle the ordering: does the server reject wrong ordering with an alert or is the "MUST" just for the client? 
I dropped the server MAY reject entries not in `supported_groups`, we should check what openssl does (to align on "MAY").

flighthandlers_client_13.go
theodorsm: This is already handled above: https://github.com/pion/dtls/pull/829/changes#diff-17521c6a8aec576ae412b4d816cf562376d9b49814476e1f5bed536e76c046dbR71-R73

JoTurk: Sorry im not talking about legacy version, I'm talking about verifying supported versions extension, 

> The server's extensions MUST contain "supported_versions".
   Additionally, it SHOULD contain the minimal set of extensions
   necessary for the client to generate a correct ClientHello pair.  As
   with the ServerHello, a HelloRetryRequest MUST NOT contain any
   extensions that were not first offered by the client in its
   ClientHello, with the exception of optionally the "cookie" (see
   [Section 4.2.2](https://www.rfc-editor.org/rfc/rfc8446.html#section-4.2.2)) extension.

but i think you wanted to verify extensions later?

flighthandlers_server_13.go
JoTurk: > I dropped the server MAY reject entries not in supported_groups, we should check what openssl does (to align on "MAY").

Sounds good I'll check what other implementations does here.

flighthandlers_client_13.go
theodorsm: Ahh ofc sorry

I was unsure at the time how it would be with the negotiation of extensions (from both client and server even if this TODO is only client side), there are a lot of edge cases with different extensions (for example keyshare that we just discussed). I think it will be easier to reason about when we have the scaffolding for more flights (HRR and SH) and we have to circle back to it when we use more extensions (like pre_shared_key) 

But we can add this case already, feel free to commit such a change

theodorsm: Actually there might not be more cases: 

> The HelloRetryRequest extensions defined in this specification are:
>  -  supported_versions (see Section 4.2.1)
>  -  cookie (see Section 4.2.2)
>  -  key_share (see Section 4.2.8)

Are other extensions negotiated? It's been a while since I have loaded the RFC into my RAM (brain)

JoTurk: > Are other extensions negotiated? 

There are some draft extensions I think. but yeah nothing else.

> But we can add this case already, feel free to commit such a change

Actually this turned out to be a bit more complicated than what I thought and full of edge cases because our supported versions extension didn't have a state for if the value is single-value or vector. And we had no way to reject vector versions in ServerHello.

https://github.com/pion/dtls/pull/829/changes/da13f71b7fca3b67bc0090f050d7551bc653016a

JoTurk: did the same for ClientHello in https://github.com/pion/dtls/pull/829/commits/2f3a3cfbc1d8c9a82c656f485c53a5132af61df3

JoTurk:
I see other issues, but i need to confirm more against the spec, and there many code with missing test coverage, I'll comment, and push fixes.

PR #835, opened 2026-05-16
Fix not sending MasterKeyIdentifier in the use_srtp extension

theodorsm:
#### Description

Currently, the `use_srtp` extension is missing the configured `MasterKeyIdentifier` in flight 3 (the second ClientHello after a HelloVerifyRequest) 

Also updated the comment above `UseSRTP` to point to the correct RFC.

### DIFF
--- a/flight3handler.go
+++ b/flight3handler.go
@@ -307,7 +307,8 @@ func flight3Generate(
 
 	if len(cfg.localSRTPProtectionProfiles) > 0 {
 		extensions = append(extensions, &extension.UseSRTP{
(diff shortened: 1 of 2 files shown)


JoTurk (approved):
Thank you, Will be cool to have a regression test.

theodorsm:
@JoTurk thanks for the quick review. We are lacking proper tests for flights currently. I want to improve that together with the new flights for dtls 1.3. I'll add this as a regressions test if I make such a PR in the future, I suspect there are more subtle bugs hiding around too.

PR #838, opened 2026-06-02
Fix panic unmarshaling ECDHE_PSK ServerKeyExchange

Sean-Der:
Missing bounds check a invalid ServerKeyExchange message could cause crash.

Reported by Karolina GORNA

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,11 @@ Native [DTLS 1.2][rfc6347] implementation in the Go programming language.
 
 A long term goal is a professional security review, and maybe an inclusion in stdlib.
 
+> [!WARNING]
+> This branch is currently frozen from tagging while DTLS 1.3 work is underway.
+>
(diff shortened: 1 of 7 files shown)


JoTurk (commented):
@Sean-Der should target `dtls-1.2` instead of `main` since we have breaking changes for dtls 1.3 in main :) maybe we should change the default branch until dtls 1.3 is released.

JoTurk:
I'll also keep main in sync with dtls-1.2 :)

Sean-Der:
@JoTurk `master` is going to be `v4`?

Can I start a `v3` branch? Will `dtls-1.2` be long lived?

JoTurk:
@Sean-Der main is going to be v4 yeah, dtls-1.2 is just `v3`, and not planned to be long lived, after dtls 1.3 is released. we can rename it that if you think it's clearer.

Sean-Der:
Nice, ok ill will make `v3` branch!

I don't know if its a Go idiom, but just been the pattern so far
```
sean@SeanLaptop:~/go/src/github.com/pion/webrtc$ git branch -r | grep origin/v
  origin/v2
  origin/v3
```

PR #841, opened 2026-06-06
Restore Firefox compatibility

aler9:
#### Description

Patch #801 silently changed the way this library reacts to unknown signature algorithms - they were silently skipped before, now they cause a fatal error. This broke compatibiliy with remote peers that are advertising unhandled algorithms (like SHA256+DSA), and going all way to the top prevents Firefox from interacting with pion/webrtc.

This patch fixes the issue by restoring the old behavior of skipping unknown signature algorithms.

#### Reference issue

https://github.com/bluenviron/mediamtx/issues/5838

### DIFF
--- a/pkg/protocol/handshake/message_certificate_request.go
+++ b/pkg/protocol/handshake/message_certificate_request.go
@@ -109,9 +109,8 @@ func (m *MessageCertificateRequest) Unmarshal(data []byte) error { //nolint:cycl
 
 		scheme := binary.BigEndian.Uint16(data[offset+i : offset+i+2])
 		var alg signaturehash.Algorithm
-		err := alg.Unmarshal(tls.SignatureScheme(scheme))
-		if err != nil {
-			return errInvalidSignHashAlgorithm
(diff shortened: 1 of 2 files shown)


JoTurk:
hello @aler9. 
damn, thank you for the fix, firefox also broke compatibility with us recently https://bugzilla.mozilla.org/show_bug.cgi?id=2033783 and it was one of the reasons we decided to freeze dtls 1.3 and new work from getting released as `v3` and until dtls and browser interop is ready (I'll try to finish them this weekend) https://github.com/pion/browser-tests https://github.com/pion/dtls-interop

Can you make this PR target `v3` branch? because main has breaking changes for dtls 1.3.

aler9:
@JoTurk i've rebased against v3.

JoTurk (approved):
Thank you @aler9 and sorry about the bad patch, I'll release this, and retract the bad version.

JoTurk:
@aler9 released in v3.1.4.

PR #844, opened 2026-06-16
sync main with v3

JoTurk:
#### Description
Sync v3 with the missing commits from main.


JoTurk:
this github bug is so annoying
image

Issue #727, opened 2025-09-14
DTLS 1.3 MVP features


theodorsm:
This issue aims to collect features needed for an MVP DTLS 1.3 implementation and serve as the starting point for a more rigorous implementation down the line (discussed in #188).

- [x] Config API
- [x] supported_version extension
- [x] cookie extension
- [x] signature_algorithms/cert extension update
- [x] Mandatory-to-Implement cipher suites (TLS_AES_128_GCM_SHA256, TLS_AES_256_GCM_SHA384 and TLS_CHACHA20_POLY1305_SHA256).
- [x] Mandatory-to-Implement signatures (rsa_pkcs1_sha256, rsa_pss_rsae_sha256 and ecdsa_secp256r1_sha256).
- [x] key_share extension
- [x] PSK extensions
- [x] EncryptedExtensions
- [x] Record layer encoding update
- [x] Encrypted sequence numbers
- [x] Key derivation 
- [x] Finished message / Transcript Hash update
- [x] Certificate/Request message update
- [x] 1-RTT handshake flights
- [x] Complete ACK reliability (#818)
- [x] Complete HelloRetryRequest/cookie exchange (#825)
- [x] Complete MVP post-handshake processing: NewSessionTicket and KeyUpdate (#824)
- [x] Fix record-layer CID-state validation (#846)
- [ ] Complete DTLS 1.3 interoperability testing
- [ ] Add interoperability testing for dual-stack mode against DTLS 1.2-only implementations

PR #859, opened 2026-06-24
Send and require signature_algorithms

JoTurk:
#### Description
RFC 8446: defines signature_algorithms vs signature_algorithms_cert. signature_algorithms is required for certificate authentication and missing it requires missing_extension.
RFC 8446: a ClientHello without pre_shared_key must contain both signature_algorithms and supported_groups, and  ClientHellos must be rejected with missing_extension.
RFC 8446: missing_extension as alert 109.
add signature_algorithms to flight13_3Generate in retry/cookie ClientHello and errors.

This is part of handling all the different kind of cases with extensions now since flights are merged, related to https://github.com/pion/dtls/pull/829#discussion_r3448002648 
#### Reference issue
part of #825 refs #727

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -66,6 +66,10 @@ var (
 	//nolint:err113
 	errInvalidClientHello = &FatalError{Err: errors.New("invalid ClientHello")}
 	//nolint:err113
+	errMissingClientHelloExtension = &FatalError{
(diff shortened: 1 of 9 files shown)


JoTurk:
(I'll improve test coverage later, I don't want to add a lot of coverage here because many things need to change.)

Sean-Der (approved):
LGTM thanks @JoTurk :D

Issue #848, opened 2026-06-19
Refactor this library errors to use standard Go error handling


JoTurk:
Since we're planning for a major release, remove the custom error handling and make it possible to remove `err113`  and other related nolint. use standard wrapping and error helpers.

PR #860, opened 2026-06-25
Move errors to internal/errors

Sean-Der:
Reduce duplicate definition of errors across packages


JoTurk (commented):
@Sean-Der good call i think.

While we're at it, I would love if we can get rid of InternalError, TemporaryError, TimeoutError, and HandshakeError and use standard error wrapping https://github.com/pion/dtls/issues/848 what are your thoughts on this?

Sean-Der:
lets do it!

JoTurk (approved):
nice 🔥

Issue #851, opened 2026-06-20
DTLS 1.3 - Integrate key-exchange flights into handshake transcript


JoTurk:
This issue tracks integrating key-exchange flights into handshake transcript blocked by https://github.com/pion/dtls/pull/829 https://github.com/pion/dtls/pull/832 https://github.com/pion/dtls/pull/834

PR #864, opened 2026-06-26
Add transcript append helpers for outbound handshake

JoTurk:
#### Description
transcript append helpers for outbound handshake messages and populate cipher suite and compression method fields for HelloRetryRequest ServerHello. 
#### Reference issue
1/2 part of #851 and blocks cipher suites and record work. refs #727

### DIFF
--- a/handshaker_13.go
+++ b/handshaker_13.go
@@ -5,6 +5,8 @@ package dtls
 
 import (
 	"context"
+	"crypto/sha256"
+	"errors"
 	"time"
 
 	dtlserrors "github.com/pion/dtls/v3/internal/errors"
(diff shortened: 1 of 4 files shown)


Sean-Der (approved):

handshaker_13.go
```diff
@@ -276,3 +278,120 @@ func (s *handshakeFSM13) wait(ctx context.Context, conn flightConn) (handshakeSt
...
+
+	sender := transcriptSenderForSide13(isClient)
+	for _, p := range pkts {
+		h, canonical, ok, err := canonicalOutboundHandshake13(p)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			continue
+		}
```
Ibn-Butt-Toota, line 321: Is it necessary to have both `ok` and `err`? Or is this something that will be cleaned up in the next part?

handshaker_13.go
```diff
@@ -276,3 +278,120 @@ func (s *handshakeFSM13) wait(ctx context.Context, conn flightConn) (handshakeSt
...
+) error {
+	id := transcriptMessageID13{
+		sender: sender,
+		seq:    h.Header.MessageSequence,
+	}
+	if sh, ok := h.Message.(*handshake.MessageServerHello); ok && isHelloRetryRequest(sh) {
+		duplicate, err := transcript.hasCanonical13(id, canonical)
+		if err != nil || duplicate {
+			return err
+		}
```
Ibn-Butt-Toota, line 368: Tiny nit, maybe this can be written as:

```suggestion
		if duplicate, err := transcript.hasCanonical13(id, canonical); err != nil || duplicate {
			return err
		}
```
which would scope the `duplicate` variable to the condition? Just a style thing, either way is fine.

Ibn-Butt-Toota (approved):

handshaker_13.go
JoTurk: Yeah it's already hard to read, I think this version will make it harder. I wanted to make it:

```go
duplicate, err := transcript.hasCanonical13(id, canonical)
if err != nil {
	return err
}
if duplicate {
	return nil
}
```

JoTurk: It will change after we do cache anyways.

handshaker_13.go
JoTurk: yes, ok controls sean or not sean. err = invalid canonical or sean but changed bytes. both will need different handling.

PR #893, opened 2026-07-05
Only pass extensions declared length

JoTurk:
#### Description
The extension loop calls the unmarshal with all the remaining payload which includes next extensions, and this is a foot gun because extensions can incorrectly try to read or verify the remainings.

Example https://github.com/pion/dtls/pull/806#discussion_r2866986724

### DIFF
--- a/pkg/protocol/extension/extension.go
+++ b/pkg/protocol/extension/extension.go
@@ -73,53 +73,56 @@ func Unmarshal(buf []byte) ([]Extension, error) { //nolint:cyclop
 
 	for offset := 2; offset < len(buf); {
 		bufView := buf[offset:] //nolint:gosec // offset bounded by loop condition
-		if len(bufView) < 2 {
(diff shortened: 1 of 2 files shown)


pkg/protocol/extension/extension.go
```diff
@@ -73,53 +73,56 @@ func Unmarshal(buf []byte) ([]Extension, error) { //nolint:cyclop
 
 	for offset := 2; offset < len(buf); {
 		bufView := buf[offset:] //nolint:gosec // offset bounded by loop condition
-		if len(bufView) < 2 {
+		if len(bufView) < 4 {
```
JoTurk, line 76: this isn't a new change, we verify this inside all the extensions unmarshal and at the end of the loop. but since we're going to read type and length in the loop we need to also verify it here before we enter unmarshal.

Sean-Der (approved):

PR #898, opened 2026-07-06
Unblock reads when conn is closed

JoTurk:
#### Description
I noticed that the CI sometimes fails with timeout (esp TestNetTest/pingpong) and i thought it was flaky slow CI, but i noticed the same issue locally and during interop tests. 
And by tracking the issue it turns out that conn can actually block on write if the connection is closed. This was fixed a while ago for SCTP users by doing writeDeadline https://github.com/pion/sctp/pull/465 but it doesn't fix it at DTLS side.
Did unrelated refactor to make the linter happy.
The back-port for v3: https://github.com/pion/dtls/pull/899

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -194,6 +194,7 @@ type Conn struct {
 	handshakeCompletedSuccessfully atomic.Bool
 	handshakeMutex                 sync.Mutex
 	handshakeDone                  chan struct{}
+	writeLock                      sync.Mutex
 
 	encryptedPackets []addrPkt
 
@@ -838,6 +839,8 @@ func (c *Conn) Read(buff []byte) (n int, err error) { //nolint:cyclop
 
 	for {
 		select {
+		case <-c.closed.Done():
+			return 0, io.EOF
 		case <-c.readDeadline.Done():
 			return 0, dtlserrors.ErrDeadlineExceeded
 		case out, ok := <-c.decrypted:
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -940,57 +946,149 @@ func (c *Conn) RemoteSRTPMasterKeyIdentifier() ([]byte, bool) {
 }
 
 func (c *Conn) writePackets(ctx context.Context, pkts []*dtlsflight.Packet) error {
+	c.writeLock.Lock()
+	defer c.writeLock.Unlock()
```
boks1971, line 950: This lock is only used here, so not a concern. But, it is holding lock across network i/o. Is that okay?

boks1971 (approved):

conn.go
JoTurk: wirtePackets always had a lock since it was made (but it used conn's shared locks), this moves the write-life-time lock to a dedicated write owned lock which should actually improve the holding time 

https://github.com/pion/dtls/blame/d8bc7dbd9805d45c19189ebe33c774f44e1237c4/conn.go#L871-L873

JoTurk: We still use the shared lock in prepareRawPackets, but we don't hold it through wirtePackets  as we used to do. This is why the Github review makes it seems like we introduced locks to writePackets.

boks1971: a ha, thank you Jo. Yeah, the diff was hard to read and I missed it. Sorry about the false alarm and thank you for the clarification.

Issue #736, opened 2025-09-30
DTLS v1.3 Key Schedule/Derivation


philipch07:
This issue is meant to track the progress regarding implementing the key scheduling/derivation for DTLS v1.3 (see the parent issue regarding the other tasks for the MVP feature set for DTLS v1.3).

The relevant section of [RFC 9147 (DTLS v1.3)](https://www.rfc-editor.org/rfc/rfc9147.pdf) is section 5.9. It then refers to [RFC 8446 (TLS v1.3)](https://www.rfc-editor.org/rfc/pdfrfc/rfc8446.txt.pdf).

PR #911, opened 2026-07-08
Share  record protection between handshake and application traffic secrets

JoTurk:
#### Description
An important refactor needed for #736 so application init can replace existing handshake record protection with allow-reinitialize. and record-protection can be shared between handshake and application records.


JoTurk:
I'll add tests for the new branches in here when i wake up and merge this.

JoTurk:
Added tests.

Ibn-Butt-Toota (approved):

PR #950, opened 2026-08-02
removed unused cookieLength and sessionLength

ashleywma:
Hi im new to Pion and I saw there were two unused variables and I thought they were okay to remove since they are not exported.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -42,8 +42,6 @@ import (
 
 const (
 	initialTickerInterval = time.Second
-	cookieLength          = 20
-	sessionLength         = 32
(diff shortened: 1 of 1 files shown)


JoTurk:
Hello, and welcome to pion sorry these aren't used now but we're going to use them after post handshake cleanup, dtls is currently confusing because we have pending dtls 1.3 work

JoTurk:
Sorry I was just looking from my phone and I thought this is the flight code and not conn, yeah we moved this logic to internal

JoTurk (approved):
Thank you

PR #956, opened 2026-08-03
Polish Go naming and cloning idioms

Sean-Der:
Rename the queued-packet buffer guard to describe its ownership role and replace abbreviated locals with descriptive names.

Use standard clone helpers and consistent empty checks. Clarify comments without changing protocol behavior.


JoTurk (approved):
I went over many of the important clone changes i don't think it will case any issues, We can also enable this modernazie setting next time we upgrade golangci https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_appendclipped and if someone wants to keep append or empty slices they can just do a nolint.

Issue #824, opened 2026-04-21
DTLS 1.3 Post Handshake messages


theodorsm:
This issue aims to track the implementation efforts for the Post-Handshake messages introduced for DTLS 1.3, see [RFC 9147 Section 5.8.4](https://datatracker.ietf.org/doc/html/rfc9147#section-5.8.4) and [RFC 8446 section 4.6](https://datatracker.ietf.org/doc/html/rfc8446#section-4.6). 

  ## Completed                                                                                                            
                                                                                                                                                                                                                                                    
  - [x] Marshal/unmarshal NewSessionTicket                                                                                
  - [x] Marshal/unmarshal KeyUpdate                                                                                                                                                                                                                 
  - [x] Marshal/unmarshal NewConnectionId                                                                                                                                                                                                           
  - [x] Marshal/unmarshal RequestConnectionId                                                                                                                                                                                                       
  - [x] Add post-handshake state-manager scaffolding                                                                      

  ## MVP work remaining

  - [x] Drive the post-handshake manager from FSM StateFinished
  - [x] Send the server's queued NewSessionTicket
  - [x] Track outbound records and retire them from received ACKs
  - [x] Retransmit unacknowledged post-handshake fragments
  - [x] Receive and store NewSessionTicket on the client
  - [x] Implement KeyUpdate send/receive
  - [x] Ratchet read/write traffic secrets independently
  - [x] Add loss, duplication, and reordering tests

  ## Post-MVP

  - [x] NewConnectionId / RequestConnectionId integration (#775)
  - [x] Post-handshake client authentication

PR #970, opened 2026-08-05
Handle DTLS 1.3 ACK retransmissions

Sean-Der:
Track outbound handshake records by logical fragment and retire them as ACKs arrive. Retransmit only missing fragments, stop completed flights, and cover partial, duplicate, lost, and reordered ACK behavior.

Implements the handshake-scoped portions of #818 without wiring the post-handshake processing tracked by #824.


Ibn-Butt-Toota (approved):
LGTM!

PR #971, opened 2026-08-05
Implement Post-handshake handling and NewSessionTicket

JoTurk:
#### Description
~~work in progress: Implementing NewSessionTicket, the first user of all the postHandshake scaffold we added recently, hopefully no need for refactors, We need to wire a lot of things we implemented for post-handshake.~~

Edit: Sadly had to make some refactors, for example: postHandshake now extends handshakeContext (I tried to avoid this becuase when I made the postHandshake struct I didn't want it to turn into a mini-FSM), sadly this unavoidable because FSM can't pass states and configs every-time postHandshake needs it and the work around was worse (hooks/callbacks).

Was tested and implemented against BoringSSL: https://github.com/pion/dtls-interop/pull/7

From the todos in #824 This completes:

- [x] `Drive the post-handshake manager from FSM StateFinished`
- [x] `Send the server's queued NewSessionTicket`
- [x] `Track outbound records and retire them from received ACKs`
- [x]  `Receive and store NewSessionTicket on the client`: Just without PSK handling, I moved PSK into its own ticket for post-MVP #972 

And as for: `Retransmit unacknowledged post-handshake fragments`: we do re-transmission but currently we resend the entire packet and not just frags, I'll make a follow up PR.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -4017,7 +4017,7 @@ func TestDTLS13ServerSendsFinalACK(t *testing.T) {
 	go func() { errs <- server.HandshakeContext(ctx) }()
 	require.NoError(t, <-errs)
 	require.NoError(t, <-errs)
-	assert.Equal(t, int32(1), applicationEpochWrites.Load())
+	assert.GreaterOrEqual(t, applicationEpochWrites.Load(), int32(1))
 
 	var rawACK []byte
 	select {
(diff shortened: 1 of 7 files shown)


JoTurk:
This is just missing new unit tests.

FrantaBOT (approved):

PR #973, opened 2026-08-06
Run dtls-interop in CI

### LINKED ISSUE #830: Move DTLS interop runners to another repo
I was going to add Boringssl and other interop runners, but BoringSSL and others requires us to add code to the codebase which makes it harder to navigate the repo and harder for new contributors.
This issue tracks moving the interop runners to another repo, we should still run them in the CI.

JoTurk:
#### Description
run https://github.com/pion/dtls-interop in the CI.
#### Reference issue
Closes #830


JoTurk:
fails because https://github.com/pion/dtls/pull/971 isn't merged and it should work there after this is merged.

PR #982, opened 2026-08-08
Derive traffic secert update

JoTurk:
#### Description
https://datatracker.ietf.org/doc/html/rfc8446#section-7.2
We missed this while we were implementing #736 and It's now required to implement post-handshake #824 keyUpdate.
the CI interp will fail because I added an interop for keyUpdate https://github.com/pion/dtls-interop/pull/11


FrantaBOT (approved):

JoTurk:
@FrantaBOT thank you for the review, I cleaned the RFC reference comment, when we're done, I'm going to normalize all the RFC references and add it in the parts that are missing any reference.

Issue #983, opened 2026-08-08
DTLS 1.3 - epoch-aware record layer state


JoTurk:
This issue extends from #755 For the sake of simplicity (and because I didn't really know the details of https://datatracker.ietf.org/doc/html/rfc9147#section-8 ) the initial implementation of the record layer encoding wasn't epoch-state aware, and only maintained a single active epoch and record-protection per direction.
This issue tracks adding per-epoch full DTLS 1.3 record state, current/next/previous read protection, retained receive and send protections, and all the refactors / rework needed to implement post-handshake keyUpdate in #824 

Part of #727

PR #984, opened 2026-08-08
Add epoch-aware traffic keys state

JoTurk:
#### Description
This adds the state and structs needed to bind and hold per-epoch DTLS 1.3, required for #983


FrantaBOT (approved):

JoTurk:
thank you for the review @FrantaBOT

PR #991, opened 2026-08-08
Retain previous read keys

JoTurk:
#### Description
A quick follow up on #983, retains previous read keys, this is the last plumbing PR before we can work on UpdateKey itself #824


JoTurk:
Also refactored the openCiphertextRecord cipher tests and made them more useful, mainly because dogsled was complaining about all the _, _, _, now that we added epoch to the returned values. this is why the diff is bigger than what it supposed to be.

treejadey (approved):

PR #994, opened 2026-08-08
Implement outbound keyUpdate

JoTurk:
#### Description
This Implements outbound keyUpdate this is a really small change since we already implemented all of the hard multi epoch record state and keyring handling.
part of #824


JoTurk:
I avoided adding pointless unit-tests here this is why the coverage is bad, once the inbound keyupdate is done, i'll add e2e test to cover both sides.

FrantaBOT (approved):

JoTurk:
thank you for the review @FrantaBOT

Issue #996, opened 2026-08-09
DTLS 1.3  - unified serialization of application and post-handshake writes.


JoTurk:
The RFCs impose some ordering constraints, for example for KeyUpdate:
1. requested keyUpdate must be sent prior to sending next application records: https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.3
2. new keys can't be used until keyUpdate is acked https://datatracker.ietf.org/doc/html/rfc9147#section-8

The current implementation uses `PauseApplicationData` / `ResumeApplicationData`
as a temporary hacky workaround to get stuff working. After KeyUpdate #824, I'll refactor
this into clear unified serialization of application and post-handshake writes.

part of #727

PR #997, opened 2026-08-09
Add public UpdateKeys API

JoTurk:
#### Description
This is the last PR for post-handshake key-update before I'll start refactoring and cleaning the implementation for example #996  I'll also check edge cases in the spec and our implementation. 

This API adds a blocking sync UpdateKeys that returns when flight was ACKed and when we commit the next write generation. Or if a failure happened.

The other option we can consider for this API is to break it into a non-blocking Start + Wait this will be similar to BoringSSL https://boringssl.googlesource.com/boringssl/+/HEAD/include/openssl/ssl.h?pli=1#327 but it will be a bit complex to use and for our side to add and maintain.

@Sean-Der what do you think?

### DIFF
--- a/internal/handshake/fsm13.go
+++ b/internal/handshake/fsm13.go
@@ -170,6 +170,58 @@ func (s *fsm13) Done() <-chan struct{} {
 	return s.closed
 }
 
+type KeyUpdater interface {
+	UpdateKeys(context.Context, handshake.KeyUpdateRequest) error
+}
+
+// UpdateKeys queues a reliable DTLS 1.3 KeyUpdate and waits until its ACK has
+// committed the next local write generation.
+func (s *fsm13) UpdateKeys(ctx context.Context, request handshake.KeyUpdateRequest) error {
+	if request != handshake.KeyUpdateNotRequested && request != handshake.KeyUpdateRequested {
+		return dtlserrors.ErrInvalidKeyUpdate
+	}
+
+	result := make(chan error, 1)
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):

internal/handshake/fsm13.go
```diff
@@ -170,6 +170,58 @@ func (s *fsm13) Done() <-chan struct{} {
...
+}
+
+// UpdateKeys queues a reliable DTLS 1.3 KeyUpdate and waits until its ACK has
+// committed the next local write generation.
+func (s *fsm13) UpdateKeys(ctx context.Context, request handshake.KeyUpdateRequest) error {
+	if request != handshake.KeyUpdateNotRequested && request != handshake.KeyUpdateRequested {
+		return dtlserrors.ErrInvalidKeyUpdate
+	}
+
+	result := make(chan error, 1)
```
Sean-Der, line 184: Could/should we use a context instead of a `chan`?

I spent a lot of time debugging/fixed chan double closes so I tend to avoid now :)

JoTurk: Yes :)

JoTurk: done in another commit, because it needed to touch all the commands and post-handshake :)

https://github.com/pion/dtls/blob/676e2a7dde42339c1d47f3edfa19848024598f5b/internal/handshake/fsm13.go#L184-L190

Issue #1000, opened 2026-08-11
DTLS 1.3 - address all the todos and unimplemented errors


JoTurk:
The DTLS 1.3 code has a lot of todo and unimplemented errors mainly because we implemented everything at the same time and some parts of the spec were blocked other unfinished stuff. 
Now we have most of the core finished we should implement all the TODOs and test them against live interop.

part of #727

PR #1003, opened 2026-08-11
Handle duplicated Flight 4

JoTurk:
#### Description
Another issue I found while testing against WolfSSL sometimes it's too impatient and retransmit flight4 just before receiving our flight 5. which causes Pion to try to go to the nonexistent flight5 parse (it's a send only flight), so we return "ErrFlightUnimplemented13" we should return unexpected_message there I'll do this as part of #1000. 
This fix makes Pion treats flight4 retransmission as duplicate and retransmit flight 5.  part of #727


JoTurk:
(This adds nolint cyclop to handleReceivedFlight and makes it a hard to read, I'll extract it after this is merged)

FrantaBOT (approved):

JoTurk:
Thank you so much for all the fast reviews @FrantaBOT

Issue #846, opened 2026-06-18
DTLS 1.3 - validate the CID presence bit against negotiated state


JoTurk:
- [x] Track `cidExpected` separately from CID length
  - [x] Reject C=1 when no CID was negotiated
  - [x] Reject C=0 when a CID is required
  - [x] Test zero-length and non-zero-length negotiated CIDs
  - [x] Fuzz both header forms

Issue #775, opened 2026-01-14
DTLS 1.3 Connection ID updates


theodorsm:
- [x] Implement NewConnectionId message encoding
  - [x] Implement RequestConnectionId message encoding
  - [x] Process both messages in the post-handshake manager (#824)
  - [x] Update active local/remote CID state
  - [x] Handle retirement and sequencing rules
  - [x] Add end-to-end CID rotation tests

JoTurk:
This is done, We just need to finish post handshake CID commands.

PR #1004, opened 2026-08-11
Implement static 1.3 CID negotiation

JoTurk:
#### Description
Implements static  CID negotiation against WolfSSL https://github.com/pion/dtls-interop/pull/15 I tested this with a local record-layer change and it works against wolfssl, I'll open a PR after this is merged for record-layer CID changes (should be small).

most of #846 and part of #775

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -591,8 +591,6 @@ func (c *Conn) Write(payload []byte) (int, error) {
 		return 0, err
 	}
 
-	//nolint:godox
-	// TODO: check for version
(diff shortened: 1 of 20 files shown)


JoTurk:
I added 3 nolints but I'll remove them in a small refactor after this, the test coverage is bad because I don't wanna lock the implementation before it's done, and I'm going to add tests after everything.

FrantaBOT (approved):

Issue #1005, opened 2026-08-12
DTLS 1.3 - review the dual and single state life cycle and edge cases


JoTurk:
Before we release 1.3 we should:
- [ ] Review the state before version negotiation and make sure that we don't send or receive unexpected versioned packets or extensions during the early state.
- [ ] in single modes we shouldn't accept or send packets from incorrect version (currently we accept 13 datagrams in the 1.2 mode and vice versa).
- [ ] We should add interop coverage for 1.2 to make sure we don't break 1.2 clients or servers (Against real implementations and not just the spec).

part of #727

JoTurk:
Since many public APIs use protocol/version now, and we export it in `State`, we should change it from a mutable struct to:

```go
type Version uint16

const (
    Version1_0 Version = 0xfeff
    Version1_2 Version = 0xfefd
    Version1_3 Version = 0xfefc
)
```

major and minor can be exported as methods.

The current API:

https://github.com/pion/dtls/blob/fd27a52a64265097a621432090926a651109b4cb/pkg/protocol/version.go#L8-L20

PR #1006, opened 2026-08-12
Integrate CID into record-layer and conn

JoTurk:
#### Description
Follow up on https://github.com/pion/dtls/pull/1004 this finishes the CID integration into record-layer and conn, and makes the CID interop works against WolfSSL. After CID we'll need to do dual-state and single modes hardening as part of  #1005  
The remaining CID work is the post-handshake CID messages.
part of #846  and #775 
I'll rework the exported pkg code as part of #1007

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1099,6 +1099,10 @@ func (c *Conn) sealRecordContent(
 		SeqBit:         true,
 		LengthBit:      true,
 	}
+	if state13, ok := c.state.(*dtlsstate.State13); ok &&
+		state13.CID.Negotiated && state13.CID.Send.UseCID {
+		header.ConnectionID = bytes.Clone(state13.CID.Send.Active)
+	}
 
 	ciphertext, err := generation.Protection.Seal(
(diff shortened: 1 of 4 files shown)


Sean-Der (approved):

internal/flight/flight12/flight3handler.go
```diff
@@ -381,3 +380,8 @@ func flight3Generate( //nolint:cyclop
 		},
 	}, nil, nil
 }
+
+func isExtendedMasterSecretRequested(policy dtlsconfig.ExtendedMasterSecretType) bool {
+	return policy == dtlsconfig.RequestExtendedMasterSecret ||
+		policy == dtlsconfig.RequireExtendedMasterSecret
+}
```
JoTurk: This was committed by mistake, this is just to remove the nolint, I'll do in a separate commit

JoTurk:
Removed the flight code change (to remove nolint:cyclop) that was added here by mistake.

PR #1016, opened 2026-08-13
Make receive buffer size configurable

### LINKED ISSUE #642: Receive Packet Length of 8192 is undocumented, inconsistent (?)
### Your environment.
  - Version: *Release or SHA*
  - Browser: *include version*

### What did you do?

A working example of what I'm trying to do is [in this gist](https://gist.github.com/jspc/716d826faef4bfdb8c9b95fef26233b6)

I have some _potentially_ large payloads I want to be able to transmit with this package. I say potentially, because the payloads are unbounded.

The way I've done this is via a pretty naive fragmenter; since the maximum datagram size is 64k and we have tls overheads, split the data in chunks of slighty-smaller-than-64k, and send them on. The server side of this, as implemented with this package, works fine; I've verified this with wireshark as I was trying to debug this issue.

The problem is when I come to receive the data, also with this package. Setting a buffer of slighty-smaller-than-64k returns an `io.EOF`. Not a subset of data, but an `EOF`. Through trial and error I figured out that I was able to roughly read 8k at a time. Anything larger and I get nothing.

I found [this](https://github.com/pion/dtls/blob/a5d1fac9900f1453716496731650ec20bfcf6cf5/internal/net/udp/packet_conn.go#L33) in the repo, which looks precisely like what I'm hitting.

I guess, then, what I'm asking is:

1. Does that 8k come from somewhere specific? I'm more than happy to be told I don't know anything about networking and that 8k is a magic number than can't be tweaked
2. If not, should it be made a var and settable in code? Perhaps via the `dtls.Config` struct?
3. If not, then should the maximum size we can send be set to 8k too?

I'm also, again, more than happy to be told I'm using the package wrong and that there's a smarter way for fragmenting data.

### What did you expect?

To be able to receive packets of more than 8k, or for this to be documented somewhere. 

Or, if it is documented, perhaps either clearer, or for the server to enforce the same limit.

### What happened?

I was unable to, and in fact received an `io.EOF` instead.

the-sarge:
The buffers used to read incoming datagrams are hard-coded to 8192 bytes in two places: the conn read loop and the udp listener's demux loop. A datagram bigger than that gets silently truncated by the OS and the receiver just sees a timeout or EOF...even though the send side produces datagrams up to 64K.

This adds a `WithReceiveBufferSize` option that sizes both read paths:

```go
dtls.DialWithOptions("udp", addr, dtls.WithReceiveBufferSize(65536))
```

The default stays 8192, so no change unless you opt in.

Added a test that round-trips an 8500-byte payload over loopback UDP. It fails against main with the same timeout from the issue.

Fixes #642

### DIFF
--- a/config.go
+++ b/config.go
@@ -36,6 +36,7 @@ var defaultCurves = []elliptic.Curve{ //nolint:gochecknoglobals
 type connConfigValues struct {
 	logger                      logging.LeveledLogger
 	maximumTransmissionUnit     int
+	receiveBufferSize           int
 	paddingLengthGenerator      func(uint) uint
 	replayProtectionWindow      int
 	initialRetransmitInterval   time.Duration
@@ -74,10 +75,11 @@ func newConnConfigValues(config *dtlsConfig) (connConfigValues, error) {
 
 	return connConfigValues{
 		logger:                      newConnLogger(config),
-		maximumTransmissionUnit:     effectiveMTU(config.MTU),
-		paddingLengthGenerator:      effectivePaddingLengthGenerator(config.PaddingLengthGenerator),
+		maximumTransmissionUnit:     config.MTU,
+		receiveBufferSize:           config.ReceiveBufferSize,
+		paddingLengthGenerator:      config.PaddingLengthGenerator,
 		replayProtectionWindow:      effectiveReplayProtectionWindow(config.ReplayProtectionWindow),
-		initialRetransmitInterval:   effectiveFlightInterval(config.FlightInterval),
+		initialRetransmitInterval:   config.FlightInterval,
 		minVersion:                  minVersion,
 		maxVersion:                  maxVersion,
 		cipherSuites:                cipherSuites,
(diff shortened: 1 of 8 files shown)


JoTurk:
Hello, thank you for the PR, I was thinking about this issue, I don't think making the buffer configurable is right fix because custom size is just custom truncation point and it doesn't fix the problem that we truncate bigger datagrams than our buffer.
I was thinking that we should do either of / or both of:
1. way bigger buffer but pooled between peers, we'll still set a limit per peer to prevent abuse but we shouldn't truncate big datagrams.
2. we just set the receive buffer to the max UDP datagram ~65kb or the the max DTLS 1.3 record of ~16KB (but this is tricky because one datagram can have multiple records).

But also changing the receive buffer here will require a similar change in pion/ice because we have the same limit there, @Sean-Der what do you think?

the-sarge:
Yes - I probably should have checked before basing the design on a 2 year old comment. ;)

How about:

- Always allocate/read up to the maximum UDP datagram size. Agree that 16 KiB limit is insufficient because one datagram can contain multiple DTLS records.

- Continue pooling buffers. I already use a shared sync.Pool for the read path, and the listener demux uses 1 receive buffer. I could change those to 64 KiB and remove the public option, then bind to prevent abuse as you suggested, without truncating datagrams.

I could look at ice too if you'd like.

JoTurk:
@the-sarge I think we'll have overflow bugs with many of our ciphers and record layer code if we push to 65535, i think the safest max value would around 65472? (CBC-SHA256).
Anyways do you know any use case for increasing the buffer size? My ISP doesn't allow me to send anything bigger than 9k even with IP fragmentation (I have jumbo frames).

Sean-Der (approved):

Issue #1009, opened 2026-08-12
Clean extensions ownership and refactor them into shared, DTLS 1.2, and DTLS 1.3 packages


JoTurk:
This is an initial proposal for the extension refactoring and ownership cleanup.

The current `pkg/protocol/extension` package has several problem (some are inherited from v3):

- `extension.Unmarshal` selects a type using only the numeric extension type.
- Unknown extensions are silently discarded.
- Some extensions like `CertificateAuthorities` and `PostHandshakeAuth` are missing from the unmarshal.
- Several DTLS 1.3 types combine multiple formats into union-like structs:
  - `SupportedVersions`
  - `KeyShare`
  - `PreSharedKey`
  - `EarlyDataIndication`
- Extension placement is not validated against the handshake message against the spec.
- Some shared extensions have different request and response payloads. For example, the current `ServerName` cannot parse the empty server ack.
- the name `SupportedEllipticCurves` is not accurate since the registry includes non-elliptic and hybrid named groups.

### split idea:

```text
pkg/protocol/extension/
  type.go
  raw.go
  list.go

  server_name.go
  alpn.go
  use_srtp.go
  srtp_protection_profile.go
  connection_id.go
  supported_groups.go
  signature_algorithms.go
  signature_algorithms_cert.go

  dtls12/
      supported_point_formats.go
      extended_master_secret.go
      renegotiation_info.go

  dtls13/
      supported_versions.go
      cookie.go
      key_share.go
      pre_shared_key.go
      psk_key_exchange_modes.go
      early_data.go
      certificate_authorities.go
      oid_filters.go
      post_handshake_auth.go
```

### Public API

The root package should have the global type namespace and extension framing.

```go
type Type uint16

type Raw struct {
  Type Type
  Data []byte
}

type Extension interface {
  ExtensionType() Type
  MarshalData() ([]byte, error)
}

type PayloadUnmarshaller interface {
  UnmarshalData([]byte) error
}

func ParseList(data []byte) ([]Raw, error)
func MarshalList(extensions []Extension) ([]byte, error)
```

`Raw` should implement `Extension`, allowing unknown extensions to remain in a decoded list and be marshalled again.

`ExtensionType()` instead of `Type()` so unknown `Raw` can retain a `Type` field.

#### Framing api

`ParseList` should:

- Validate the outer extension-list length.
- Validate each extension header and payload length.
- Preserve extension order.
- Preserve duplicate entries for later contextual validation.
[...]

PR #1020, opened 2026-08-14
Fix SRTP not being negotiated over DTLS 1.3

kulikov0:
I tried to use WebRTC data channels with DTLS 1.3, and the connection failed, reporting that no SRTP Protection Profile was chosen. This occurs because `flight4handler.go` generates an empty EncryptedExtensions block, leaving the client's `SelectedSRTPProtectionProfile()` at 0.

Linked to #727 and #1000

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -1110,6 +1110,75 @@ func TestSRTPConfiguration(t *testing.T) {
 	}
 }
 
+func TestSRTPConfigurationDTLS13(t *testing.T) {
+	// Check for leaking routines
+	report := test.CheckRoutines(t)
+	defer report()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	clientMKI := []byte("ClientSRTPMKI")
+	serverMKI := []byte("ServerSRTPMKI")
+
+	ca, cb := dpipe.Pipe()
+	type result struct {
+		c   *Conn
+		err error
+	}
+	resultCh := make(chan result)
+
+	go func() {
+		opts := []ClientOption{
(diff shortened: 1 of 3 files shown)


JoTurk:
Hello, I'm sorry DTLS 1.3 isn't released yet, if you want to use it from main you can set the SRTP profiles with WithSRTPProtectionProfiles (currently only works if pion is the server becase we add it to ClientHello) this is what i do to test against chrome / firefox. But it still has small gaps and not intended for prod use.

This was  planned to be done after the extenstions / use-srtp cleanup in #1009 Because i want the extension ownership to be clear, and that we validate the orders and the requirements in the spec. And I don't want to piggy back this logic in the flight code. Sorry for closing this PR.

JoTurk:
hello @kulikov0  this issue #1021 explains more why we don't want to have duplicated SRTP selection or extensions logic in general in the flights code, we'll extract the SRTP selection logic from 1.2 and share it in the same place between 1.2 and 1.3 for EncryptedExtensions (currently we only send it with ClientHello so we can test against browsers with Pion as a server but that will be refactored too). This should be the final part before the extensions side of 1.3 is ready and i have been testing it for days.
Sorry again for closing your PR.

kulikov0:
Okey-dokey. In any case, thank you for the project. I will use the fork in my security research

JoTurk:
hello @kulikov0 after many PRs and cleanups we now negotiate SRTP in the encryptedExtensions using the same shared 1.2 and 1.3 logic https://github.com/pion/dtls/pull/1035

kulikov0:
Amazing! Thanks @JoTurk !

Issue #1021, opened 2026-08-14
Extension validation and negotiation


JoTurk:
This is a follow up to #1009 part of #825 , #1000 , #727 and #775 

Current problems with extenstions handling:
- we decode extensions blocks multiple times, and extensions logic are all over the place (handshake, flights, sometimes they are the same logic but duplicated).
- we don't validate the complete block before negotiation or state mutation.
- we don't negotiate against the final post-hook message.
- we don't have an immutable decision that we commit once.
- we don't share common extensions logic between 1.2 and 1.3.
- we shouldn't have any extensions logic in `flight12` and `flight13` and it should just have the versions and the handshake flow
- we don't reject known extensions in illegal contexts.
- we don't reject duplicate types, including unknown types, and we don't enforce the spec orders.
- we should rework our alerts so malformed payloads, illegal placement, and missing requirements produce the correct alerts.
- we should share profile selection and response validation across DTLS 1.2 and 1.3 (selection in DTLS 1.2 ServerHello and DTLS 1.3 EncryptedExtensions but same logic across both versions).
- CID-in-HRR isn't correct and inconsistent

PR #1023, opened 2026-08-15
Validate extension blocks by context

JoTurk:
#### Description
extensions registry, placement/duplicate/dependency rules, alert classification according to specs, part of the work to normalize extensions between 1.2 and 1.3, and handle extension edge cases and rules for 1.3, part of #1021 , #727 and #1005


JoTurk:
I added temporary code to make this compatible with handshake, cache and flight code until i change them in later PRs.

treejadey (approved):

PR #1028, opened 2026-08-16
Reject unsolicited ServerHello extensions

JoTurk:
#### Description
Enforces both the 1.2 and 1.3 specs and rejects unsolicited ServerHello extensions in responses  (currently only enforced in the dual mode #1005  ). does a cleanup after #1027 and removes some temporary code i added to get snapshots merged (e.g the use of stdlib "reflects" in the snapshot code, I moved the nil check to the new API before Type so we don't have to use reflects #1009 ). part of #1021


Ibn-Butt-Toota (approved):
wow...

Issue #1026, opened 2026-08-16
Return Routability Check for DTLS 1.2 and 1.3


JoTurk:
This issue tracks the implementation of https://datatracker.ietf.org/doc/rfc9853/

- [x] Initial CID offer/response negotiation for DTLS 1.2 and 1.3
  - Post-handshake CID rotation remains tracked by #775
- [x] RRC extension type 61, dependent on connection_id https://github.com/pion/dtls/pull/1047 https://github.com/pion/dtls/pull/1048
- [x] RRC content type 27 and message codecs https://github.com/pion/dtls/pull/1049
  - [x] PathChallenge
  - [x] PathResponse
  - [x] PathDrop parsing
  - [x] Graceful handling of unknown message types
- [x] Basic address-validation procedure https://github.com/pion/dtls/pull/1050
- [x] Per-candidate anti-amplification accounting https://github.com/pion/dtls/pull/1050
- [x] DTLS 1.2, DTLS 1.3, and dual-stack tests https://github.com/pion/dtls/pull/1051

skipped from this implemantion:
- enhanced validation mode.
- application-provided address validation.

part of #727

PR #1048, opened 2026-08-21
Negotiate Return Routability Check

JoTurk:
#### Description
Negotiates Return Routability Check with CID part of #1026

### DIFF
--- a/internal/state/common.go
+++ b/internal/state/common.go
@@ -52,6 +52,9 @@ type Common struct {
 	// RemoteCIDOffered reports whether the peer sent a connection_id
 	// extension.
 	RemoteCIDOffered bool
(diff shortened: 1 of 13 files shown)


internal/state/common.go
```diff
@@ -128,6 +131,7 @@ func (s *Common) RecordLocalClientHello(snapshot negotiation.ClientHelloSnapshot
 func (s *Common) ResetConnectionIDs() {
 	s.SetLocalConnectionID(nil)
 	s.RemoteConnectionID, s.LocalCIDOffered, s.RemoteCIDOffered = nil, false, false
+	s.RRCNegotiated = false
```
JoTurk, line 134: I'm not sure if resetting `RRCNegotiated` here is correct.

treejadey (approved):

PR #1052, opened 2026-08-21
Add path migration policies

JoTurk:
#### Description
This removes `WithConnectionIDGenerator` and adds `WithConnectionID(generator, policy)`.
Policy controls the CID migration policy, as required by:
https://datatracker.ietf.org/doc/html/rfc9146#section-6 and  https://datatracker.ietf.org/doc/html/rfc9147#section-11

> There is a strategy for ensuring that the new peer address is able to receive and process DTLS records. No strategy is mandated by this specification, but see note (*) below.

Pion@v3 assumed that applications that enable CID use a transport like ICE with CID validation, now we support RRC #1026 native DTLS applications can have path migration without using another transport.

This adds 3 different polices when CID is enabled:
1. migration is rejected and disabled, but CID is still enabled.
2. migration is enabled and not verified for applications that use transports like ICE to verify paths.
3. RRC-based migration, where we use RRC to ensure that the new peer address is able to receive and process DTLS records. (RRC is only sent and negotiated if this option is used).

I did some refactors to avoid adding multiple conditions in 1.2 and 1.3 flight codes. and extracted the CID code to common helpers.

(Interop will fail because of the API change).

### DIFF
--- a/config.go
+++ b/config.go
@@ -46,6 +46,7 @@ type connConfigValues struct {
 	certificateSignatureSchemes []signaturehash.Algorithm
 	ellipticCurves              []elliptic.Curve
 	serverName                  string
+	cidPathMigrationPolicy      cidPathMigrationPolicy
 }
 
 func newConnConfigValues(config *dtlsConfig) (connConfigValues, error) {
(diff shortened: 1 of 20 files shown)


Ibn-Butt-Toota (approved):

JoTurk:
@Sean-Der @theodorsm I'm going to merge this so I can fix the interop, please let me know if you have any concerns with the API, or if it doesn't make sense.

PR #1054, opened 2026-08-23
Use unique record numbers for fragments

JoTurk:
#### Description
Each fragment correctly allocates a fresh record sequence number, but we reused the packet sequence number in the CID header, which isn't correct and cases peers to reject fragments as replay duplicates.

part of #775


JoTurk:
This issue exists in v3 since v3.0.0, but i guess nobody reported or caught it because the 1.2 CID implementation was hardly tested.

Ibn-Butt-Toota (approved):

PR #1055, opened 2026-08-23
Return zero bytes from write on failure

JoTurk:
#### Description
One more "breaking change" to get before we tag v4, Write returns len(payload) regardless of whether the write succeeds or fails. This is so confusing.


Ibn-Butt-Toota (approved):

Ibn-Butt-Toota:
I'm always confused

PR #1073, opened 2026-08-25
Use the received handshake lease in DTLS 1.2 too

fippo:
recvHandshakeLease pauses the reader while an FSM transition needs exclusive access to the received records. fsm13 retains it on receive and releases it in send(), once the answering flight has been written.

fsm12 instead closed state.Done right after parsing, in both wait() and finish(), so the reader resumed one state early and could process the next datagram while prepare() was still generating the answer from the same handshake cache.

fsm12 now uses the lease at the same points as fsm13: retained on receive, released in send(), released on the paths that produce no flight, and released from Run() as a backstop. The one deliberate difference is finish(), where a 1.2 server goes on to resend its last flight and therefore keeps the lease until send().

Covered by the 1.2 counterparts of
TestHandshakeFSM13ServerFlight4KeepsReaderPausedThroughQueueDrain.

### DIFF
--- a/internal/handshake/fsm12.go
+++ b/internal/handshake/fsm12.go
@@ -61,6 +61,7 @@ type fsm12 struct {
 	cfg                *dtlsconfig.HandshakeConfig
 	closed             chan struct{}
 	establishment      *Establishment
+	received           recvHandshakeLease // keeps the reader paused across a prepare/send transition
 }
 
 func NewFSM12(
@@ -85,6 +86,8 @@ func NewFSM12(
 }
 
 func (s *fsm12) Run(ctx context.Context, conn Conn, initialState State) error {
+	defer s.received.release()
+
 	return runHandshakeFSM(
 		ctx,
 		conn,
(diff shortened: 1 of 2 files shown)


fippo:
found while doing sped stuff which needs to wait for the response packet. It *sometimes* did but not always.

Having the FSMs for 1.2 and 1.3 be more similar is nice.

JoTurk (commented):
Context for when it was added and why it wasn't applied for 1.2 https://github.com/pion/dtls/pull/937 
FSM12 doesn't work the same as FSM13 (we tried to make them similar but the specs are really different)
I don't think this fix fits 1.2 because we'll be holding the DTLS reader until WritePackets returns, for 1.2 it should be fixed at the cache layer (at prepare and before WritePackets) and not the I/O.

JoTurk:
and tldr from the PR I linked (and the failing CI), the main reason FSM13 needed recvHandshakeLease, while 1.2 does not, is ACK. if an ACK arrives while we are still processing and sending a flight, we must not process that ACK before the flight's has been committed, this isn't an issue with FSM12 and the problem you described in the PR is just logical race around handshake-cache, and applying I/O serialization here is a bit extreme fix IMO :)

fippo:
Thank you, will try a different approach - which looks *a lot* more complicated

PR #1075, opened 2026-08-27
Optimize buffer used by PacketConn

noboruma:
Optimize `PacketConn` by:
- Avoiding unbound buffer growth of buffers
- Removing locks between reads and writes

On top of that:
- Add option to set the listener backlog size
- Add option to set the `PacketConn` ringbuf size

Also add a quick function to quickly drain packets at the `Conn` layer.

### DIFF
--- a/internal/net/buffer.go
+++ b/internal/net/buffer.go
@@ -36,6 +36,15 @@ type AddrPacket struct {
 	data bytes.Buffer
 }
 
+type readNotification struct {
+	ch      chan struct{}
+	waiters atomic.Int64
+}
+
+func newReadNotification() *readNotification {
+	return &readNotification{ch: make(chan struct{})}
+}
+
 // PacketBuffer is a circular buffer for network packets. Each slot in the
 // buffer contains the remote address from which the packet was received, as
 // well as the packet data.
@@ -46,7 +55,7 @@ type PacketBuffer struct {
 	write, read         atomic.Uint64
 	readLock, writeLock sync.Mutex
 
-	notify   chan struct{}
+	notify   atomic.Pointer[readNotification]
 	closedCh chan struct{}
 	closed   atomic.Bool
 
@@ -61,13 +70,15 @@ func NewPacketBufferWithSize(bufSize int) *PacketBuffer {
 		bufSize = 1
 	}
 
-	return &PacketBuffer{
+	buffer := &PacketBuffer{
 		readDeadline: deadline.New(),
 		packets:      make([]AddrPacket, bufSize),
-		notify:       make(chan struct{}, 1),
 		closedCh:     make(chan struct{}),
 		growable:     growable,
 	}
+	buffer.notify.Store(newReadNotification())
+
+	return buffer
 }
 
 // NewPacketBuffer creates a new PacketBuffer with default buffer size.
@@ -120,10 +131,7 @@ func (b *PacketBuffer) WriteTo(pkt []byte, addr net.Addr) (int, error) {
 
 	b.write.Add(1)
 
-	select {
-	case b.notify <- struct{}{}:
-	default:
-	}
+	b.notifyReaders()
 
 	return n, nil
 }
@@ -184,13 +192,36 @@ func (b *PacketBuffer) ReadFrom(packet []byte) (n int, addr net.Addr, err error)
 }
 
 func (b *PacketBuffer) waitForRead() error {
+	notify := b.notify.Load()
+	notify.waiters.Add(1)
+
+	if b.notify.Load() != notify || b.read.Load() < b.write.Load() || b.closed.Load() {
+		notify.waiters.Add(-1)
+
+		return nil
+	}
+
+	var err error
 	select {
 	case <-b.readDeadline.Done():
-		return ErrTimeout
-	case <-b.notify:
-		return nil
+		err = ErrTimeout
+	case <-notify.ch:
 	case <-b.closedCh:
-		return nil
+	}
+
+	notify.waiters.Add(-1)
+
+	return err
+}
+
+func (b *PacketBuffer) notifyReaders() {
+	notify := b.notify.Load()
+	if notify.waiters.Load() == 0 {
+		return
+	}
+
+	if b.notify.CompareAndSwap(notify, newReadNotification()) {
+		close(notify.ch)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


options.go
```diff
@@ -393,6 +395,30 @@ func WithReceiveBufferSize(size int) Option {
...
+func WithListenerPacketConnSize(size int) Option {
+	return sharedOption(func(c *dtlsConfig) error {
+		if size <= 0 {
+			return dtlserrors.ErrInvalidReceiveBufferSize
+		}
+		c.PacketConnBufferSize = size
+
+		return nil
+	})
+}
```
JoTurk: sorry this is the wrong layer to add this, Options controls DTLS configurations (usually for both client/server) , this seems to be PacketConn / buffer configuration, We can expose our packetconn with constructors + config if users don't want to implement it themself.

internal/net/buffer.go
```diff
@@ -31,204 +32,144 @@ var ErrTimeout = dtlserrors.ErrNetBufferTimeout
...
-			}
+			addr = slot.addr
 
-			// If we were full before reading and have successfully read, we are
-			// no longer full.
-			if b.full {
-				b.full = false
-			}
+			bufferPool.Put(slot.data)
+			slot.data = nil
```
JoTurk: doesn't this break the packetconn interface? https://pkg.go.dev/net

> Multiple goroutines may invoke methods on a PacketConn simultaneously.

JoTurk (commented):
Thank you

internal/net/buffer.go
noboruma: Ah good point, I will make reads & writes serialized separately for now - still a win from today where reads & writes are serialized together

options.go
noboruma: Is there a better place where I can set those?

internal/net/buffer.go
```diff
@@ -31,212 +32,180 @@ var ErrTimeout = dtlserrors.ErrNetBufferTimeout
...
-	b.mutex.Lock()
-
-	if b.closed {
-		b.mutex.Unlock()
-
+	if b.closed.Load() {
 		return 0, io.ErrClosedPipe
 	}
+	b.writeLock.Lock()
+	defer b.writeLock.Unlock()
```
JoTurk: Close set closed outside of the lock, and we check closed here outside of the lock, This will cause a race between close and WriteTo and we can write to a closed connection, this pattern had caused deadlocks for us in other libraries (if it un-did the writeDeadline).

internal/net/buffer.go
```diff
@@ -31,212 +32,180 @@ var ErrTimeout = dtlserrors.ErrNetBufferTimeout
...
-		b.packets = newBuf
+	slot := &b.packets[write%uint64(len(b.packets))]
 
-		// Update read/write pointers and mark buffer as not full.
-		b.read = 0
-		b.write = n
-		b.full = false
-	}
+	buf := bufferPool.Get()
+	slot.data = buf.(*bytes.Buffer) //nolint:forcetypeassert
```
JoTurk: nit: please avoid adding nolint here.

options.go
JoTurk: We should let users construct PacketListener and pass it to DTLS.

internal/net/buffer.go
```diff
@@ -31,212 +32,180 @@ var ErrTimeout = dtlserrors.ErrNetBufferTimeout
...
 
+// NewPacketBuffer creates a new PacketBuffer with default buffer size.
+func NewPacketBuffer() *PacketBuffer {
+	return NewPacketBufferWithSize(defaultPacketSize)
+}
+
+//nolint:gochecknoglobals
+var bufferPool = sync.Pool{
+	New: func() any {
+		return bytes.NewBuffer(make([]byte, 4*1024))
```
JoTurk: This means we allocate 4kb for every slot, even if the dtls packet is much smaller. We used to allocate 0, is this intentional?

noboruma: I just set it arbitrarily, but now that you mentioned it I noticed I was initializing this wrong (it should have been cap with a len of 0) - I am reverting to a 0 sized buffer to keep it consistent with previous logic.

JoTurk:
While i was looking at internal/net I saw many problems (unrelated to this PR and the other), the stream-to-packet adapters don't really make sense and are hacky, CID / validation, MTU / receive / replay / records bounds aren't explicit.

I ignored this package with the V4 work, but I'll give it a short visit after I merge both of your PRs.

internal/net/buffer.go
```diff
@@ -31,212 +32,182 @@ var ErrTimeout = dtlserrors.ErrNetBufferTimeout
...
 
-	b.mutex.Unlock()
+	b.write.Add(1)
 
-	if notify != nil {
-		close(notify)
+	select {
+	case b.notify <- struct{}{}:
+	default:
 	}
```
JoTurk: If two readers are blocked we'll wake only one :) I'll push a patch to fix it.

JoTurk: fixed by https://github.com/pion/dtls/pull/1075/changes/45a3aadc5643a700c6c202305e0e4efac2a0761b

internal/net/buffer.go
```diff
@@ -31,212 +32,190 @@ var ErrTimeout = dtlserrors.ErrNetBufferTimeout
...
-		var n int
-		if b.read < b.write {
-			n = copy(newBuf, b.packets[b.read:b.write])
-		} else {
-			n = copy(newBuf, b.packets[b.read:])
-			n += copy(newBuf[n:], b.packets[:b.write])
-		}
+	if write-read >= uint64(len(b.packets)) {
+		return 0, dtlserrors.ErrBufferTooSmall
+	}
```
JoTurk, line 121: This behavior shouldn't be the default. I'm not against adding a way to bound the memory but we had issues previously when we applied limits like this and it broke some people setups (we had one in SCTP last month).

JoTurk (commented):
LGTM, I think we just need to make the hardcoded limit opt-in.

internal/net/buffer.go
noboruma: Ah good one - thanks!

internal/net/buffer.go
```diff
@@ -184,13 +192,36 @@ func (b *PacketBuffer) ReadFrom(packet []byte) (n int, addr net.Addr, err error)
 }
 
 func (b *PacketBuffer) waitForRead() error {
+	notify := b.notify.Load()
+	notify.waiters.Add(1)
+
+	if b.notify.Load() != notify || b.read.Load() < b.write.Load() || b.closed.Load() {
```
noboruma, line 200: `b.notify.Load() != notify` is guaranteed to be under the condition of `b.notify` being closed - we could skip that check entirely?

JoTurk (approved):
Thank you.

PR #1076, opened 2026-08-27
Optimize PacketConn

noboruma:
This change proposes to simplify the conns bookkeeping inside `packet_conn` by using a `sync.Map`.
It also reduce the number of goroutines per connection.

### DIFF
--- a/internal/net/udp/packet_conn.go
+++ b/internal/net/udp/packet_conn.go
@@ -18,6 +18,7 @@ package udp
 
 import (
 	"context"
+	"io"
 	"net"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


internal/net/udp/packet_conn.go
```diff
@@ -354,11 +363,9 @@ func (c *PacketConn) WriteTo(payload []byte, addr net.Addr) (n int, err error) {
 		// resulting in the remote address entry being dropped prior to the
 		// "real" client transitioning to sending using the alternate
 		// identifier.
-		if id != nil && c.rmraddr.Load() == nil && addr.String() != c.raddr.String() {
-			c.listener.connLock.Lock()
-			delete(c.listener.conns, c.raddr.String())
-			c.rmraddr.Store(true)
-			c.listener.connLock.Unlock()
+		old := c.raddr.Swap(addr)
+		if old != addr {
```
noboruma:
```suggestion
		if old.(*PacketConn).String() != addr.String() { //nolint:forcetypeassert
```

JoTurk (approved):
thank you, sorry that it took a while.

PR #1079, opened 2026-08-27
crypto: refuse non-approved algorithms in FIPS mode

jjinno:
#### Description

PR #817 made the default curve FIPS-safe. This does the same for the symmetric cipher, so a build that must stay inside the Go FIPS module (`crypto/fips140.Enabled()`) doesn't route through a cipher the module doesn't provide.

Under `GODEBUG=fips140=on` the module doesn't reject a non-approved algorithm on its own — it just isn't a validated implementation — so the algorithm has to be kept out of the handshake. This drops ChaCha20-Poly1305 (`golang.org/x/crypto`) and AES-CCM (pure-Go CCM) from negotiation, mirroring the existing `filterFIPSCurves` in `config.go`. AES-GCM and AES-CBC are unaffected.

Kept deliberately narrow — two sites:

- **Negotiation** (`config.go`) — drop the non-approved suites from the negotiated list when FIPS is on, right after the existing version filter and alongside the existing `filterFIPSCurves` curve filter.
- **Resume** (`resume.go`) — `resolveResumeCipherSuite` rebuilds the cipher from the stored `CipherSuiteID` and never re-negotiates, so it bypasses that filter. Refuse a non-approved suite at that resume boundary so a stored suite can't route around it.

Both are no-ops when FIPS mode is off, so existing behavior is unchanged. `GOEXPERIMENT=boringcrypto` builds don't set `fips140.Enabled()` and are unaffected.

#### Changes

1. **`config.go`** — `cipherSuiteFIPSApproved` (predicate, beside `filterFIPSCurves`), applied via `slices.DeleteFunc` during cipher-suite assembly when `fips140.Enabled()`.
2. **`resume.go`** — refuse a non-approved suite at the resume boundary (`resolveResumeCipherSuite`). Cipher implementations themselves are unchanged.
3. **`internal/errors/errors.go`** — adds `ErrCipherSuiteNotFIPSApproved`, returned by the resume guard. Internal-only, matching the existing `ErrNoAvailableCipherSuites` and friends.

No tests, no exported-API changes, and no `go.mod`/`go.sum` changes (`crypto/fips140` is stdlib).

#### Testing

One caveat on testing: this repo's CI runs with FIPS **off**, so `fips140.Enabled()` is false and both guards are inert there — CI confirms the change doesn't regress the non-FIPS path, nothing more. There's no test that asserts a non-approved suite is actually dropped.

### DIFF
--- a/config.go
+++ b/config.go
@@ -154,6 +154,26 @@ func filterFIPSCurves(curves []elliptic.Curve) []elliptic.Curve {
 	return filtered
 }
 
+// cipherSuiteFIPSApproved reports whether a suite's cipher comes from the Go FIPS
+// module. ChaCha20-Poly1305 and AES-CCM don't, so they're not approved in FIPS
+// mode; AES-GCM and AES-CBC are fine.
+func cipherSuiteFIPSApproved(id cryptosuite.ID) bool {
+	switch id {
+	case cryptosuite.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
+		cryptosuite.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+		cryptosuite.TLS_PSK_WITH_CHACHA20_POLY1305_SHA256,
+		cryptosuite.TLS_CHACHA20_POLY1305_SHA256,
+		cryptosuite.TLS_ECDHE_ECDSA_WITH_AES_128_CCM,
+		cryptosuite.TLS_ECDHE_ECDSA_WITH_AES_128_CCM_8,
+		cryptosuite.TLS_PSK_WITH_AES_128_CCM,
+		cryptosuite.TLS_PSK_WITH_AES_128_CCM_8,
+		cryptosuite.TLS_PSK_WITH_AES_256_CCM_8:
+		return false
+	default:
+		return true
+	}
+}
+
 func adaptVerifyConnection(verifyConnection func(*State) error) func(dtlsstate.Active) error {
 	if verifyConnection == nil {
 		return nil
@@ -402,6 +422,14 @@ func selectCipherSuites(selectedIDs []cryptosuite.ID, customCipherSuites func() 
 	cipherSuites = slices.DeleteFunc(cipherSuites, func(suite cryptosuite.Suite) bool {
 		return !slices.ContainsFunc(versions, suite.Capabilities().SupportsVersion)
 	})
+
+	// Drop ciphers the Go FIPS module can't provide when FIPS mode is on,
+	// mirroring effectiveEllipticCurves/filterFIPSCurves above.
+	if fips140.Enabled() {
+		cipherSuites = slices.DeleteFunc(cipherSuites, func(suite cryptosuite.Suite) bool {
+			return !cipherSuiteFIPSApproved(suite.ID())
+		})
+	}
 
 	var foundCertificateSuite, foundPSKSuite, foundTrafficSuite bool
 	i := 0
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -52,6 +52,9 @@ var (
 	ErrNoAvailableCipherSuites = stderrors.New(
 		"connection can not be created, no CipherSuites satisfy this Config",
 	)
+	// ErrCipherSuiteNotFIPSApproved is returned when a cipher the Go FIPS module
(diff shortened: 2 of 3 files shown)


JoTurk:
Hello, is there a reason why we need to gate this at constructors rather than a filter at the handshake boundary, this seems to be really over scopped.
Thank you.

jjinno:
@JoTurk Thanks for looking at this, and fair challenge.

I had looked into only extending what #817 did, but quickly realized that [`ResumeWithOptions`](https://github.com/pion/dtls/blob/dccb7be0a6329aebb65b4d4d7d9d5d719b6698f8/resume.go#L30) rebuilds the suite from the serialized `CipherSuiteID` (via [`ForID`](https://github.com/pion/dtls/blob/dccb7be0a6329aebb65b4d4d7d9d5d719b6698f8/state.go#L235)), which seems to skip the negotiation filter? So my worry was that a resumed session could still end up on one of the ciphers we're filtering out?

Once I made those fail closed, a bunch of existing tests went red because they happened to use those algorithms — so most of the rest of the diff is just pointing those tests at approved ones (or skipping the ones that are specifically about the excluded algorithms) so they still pass in FIPS mode.

JoTurk:
I assume that most ResumeWithOptions calls are calls on the same process or at least same machine or maybe a different node in the cluster with the same setting. So fips is unlucky to flip between export and resume.
Anyways I think a better fix for this is to make a gate there at resume rather than every constructor :)

JoTurk:
Also ResumeWithOptions serializes and restores the DTLS connection state, including the master secret. In a strict FIPS deployment, I would expect exporting such state to be avoided or banned in the first place (I think FIPS prohibit such export/resume because it cant be done in a safe and protected way, please correct me if I'm wrong).

jjinno:
Oh boy, that's my bad. In all honesty I got carried away and missed this ... significantly simpler approach.

ie. There was only ever the one uncovered path (resume) that I built all that extra constructor code for... when it's really just like a 3-line guard I could have added at the resume boundary. 

Let me update the PR to do that and get back to you.

jjinno:
@JoTurk I believe I have updated the PR based on what you had suggested... moving away from the per-cipher guards to instead guarding at the point where the suite is rebuilt from its stored ID — on [resume](https://github.com/jjinno/dtls/blob/db2099f6997d84fe522af51394fda5942f1462bb/state.go#L277-L282) and on [deserialize/import](https://github.com/jjinno/dtls/blob/db2099f6997d84fe522af51394fda5942f1462bb/state.go#L240-L246).

I added a [negative test](https://github.com/jjinno/dtls/blob/db2099f6997d84fe522af51394fda5942f1462bb/state_test.go#L68) covering both. Let me know if this is more aligned with what you were thinking?

JoTurk:
Can this be limited to the resume guard, I'm not sure we can support running and supporting the tests in fips setup, seems out of the scope and hard to maintain with future tests.

jjinno:
@JoTurk Apologies for the churn. I trimmed it down to the minimal this time.

It's now just the handshake-boundary filter (`filterFIPSCipherSuites`), mirroring the existing `filterFIPSCurves`, plus the one resume guard (since `ResumeWithOptions` rebuilds from the stored `CipherSuiteID` and skips that filter, it needed its own check). Both guards are no-ops when FIPS is off.

Dropped everything else: the deserialize guard, the curve/fingerprint/signature-hash guards, and all the FIPS tests.

### Note:
- The suite does fail if you run it with `fips140=on`, because a handful of existing tests pin one of the now-filtered suites.
- Getting it green would mean making those tests FIPS-aware (swapping the incidental ones to an approved suite, skipping the few that specifically test the excluded ones), and the repo has no precedent for FIPS-aware tests today. 

I read your previous message as wanting that kept out of this PR, so let me know if I've misread that.

JoTurk (approved):
Thank you

JoTurk:
Thank you for simplifying the PR, is it ready for merge (I'm asking because it's marked as draft)

jjinno:
@JoTurk Freshened onto latest.

### Clarifying Note:
- The code-coverage gate on this PR will never be green without me adding FIPS-aware tests. (Which I had kept out to meet your reduced scoping request). 

Are you OK merging past the code-coverage gate, or would you rather I add the tests?

JoTurk:
Thank you, I'll merge this once the ci passes.

PR #1081, opened 2026-08-29
Fix packet connection leak

noboruma:
When connection attempt triggers an error we forget to close the packetConn.


JoTurk (approved):
Good catch, I think we should fix this for v3, I'll merge this commit to v3 too and tag it once I'm home.

JoTurk:
Actually i believe this was done at the incorrect layer and there are still cases where we have the leak!

PR #1085, opened 2026-08-29
Close rejected listeners at Accept

JoTurk:
#### Description
cherry-picks https://github.com/pion/dtls/pull/1084 against main.


noboruma (approved):

JoTurk:
Porting this from v3 made me realize that we didn't drop the transient `WithOptions` suffix when we dropped non-functional constructors.

PR #863, opened 2026-06-25
Remove Config

Sean-Der:
API is functional only

PR #1086, opened 2026-08-29
Drop the transient WithOptions suffix

JoTurk:
#### Description
This suffix was added when we added options pattern to v3, we didn't remove it when we removed the config API and went options only https://github.com/pion/dtls/pull/863


JoTurk:
Interop fails because of the API change.

Issue #1007, opened 2026-08-12
DTLS 1.3 - Review the exported pkg APIs and clean up names 


JoTurk:
Before we release v4 we should review all the exported handshake and record-layer code and make sure the API is clean and reusable.
We should maybe break them into 12 and 13 sub packages, and remove the version suffixes like how we did with the internal code.
We should collapse similar functions into one function with options patterns, for example 1.2 UnpackDatagram and ContentAwareUnpackDatagram.

JoTurk:
Since many public APIs use protocol/version now, and we export it in `State`, we should change it from a mutable struct to:

```go
type Version uint16

const (
    Version1_0 Version = 0xfeff
    Version1_2 Version = 0xfefd
    Version1_3 Version = 0xfefc
)
```

major and minor can be exported as methods.

The current API:

https://github.com/pion/dtls/blob/fd27a52a64265097a621432090926a651109b4cb/pkg/protocol/version.go#L8-L20

PR #1088, opened 2026-08-29
Rework client and listener constructors

JoTurk:
#### Description
DTLS@v3 didn't have a direct way to provide a real packconn socket, `Server / ServerOption` Required `address`, and listen required `laddr` both resulted in another weird abstraction with dtls.ListenConfig and eventually with `WithListenConfig` added in https://github.com/pion/dtls/pull/802

This PR removes the public `ListenConfig` struct, and removes the `WithListenConfig` options, and refactors the constructors into:

1. `Client(conn, raddr, opts...)` Start a client-side DTLS session using an existing net.PacketConn.
2. `Server(conn, raddr, opts...)` Start one server-side DTLS session when the peer's packet stream has already been isolated. 
3. `Listen(conn, opts...)` Accept multiple DTLS clients from an already-created / bound net.PacketConn. DTLS handles demultiplexing packets by remote address or `CID`. closes #837 
4. `ListenAddr(network, laddr, opts...)` Convenience API for the common UDP-server case. It creates the packet socket and returns a DTLS listener.
5. `NewListener(inner, opts...)`  DTLS over an existing PacketListener that already accepts isolated packet connections. Useful for custom multiplexers and virtual transports (and advanced users that want to own the demultiplexing in general but don't want to create a new ListenAddr for every connection).  relates to #1075  

I think this covers all of our use cases.

Relates to #1007

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -513,14 +513,14 @@ func clientWithConfig(conn net.PacketConn, rAddr net.Addr, config *dtlsConfig) (
 	return createConn(conn, rAddr, config, true, nil)
 }
 
-// Client establishes a DTLS connection over an existing connection.
-func Client(conn net.PacketConn, rAddr net.Addr, opts ...ClientOption) (*Conn, error) {
+// Client establishes a DTLS connection over an existing packet connection.
+func Client(conn net.PacketConn, raddr net.Addr, opts ...ClientOption) (*Conn, error) {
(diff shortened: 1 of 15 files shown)


Ibn-Butt-Toota (approved):

JoTurk:
(breaks the interop because of the API change, again :)

noboruma (approved):

PR #1093, opened 2026-08-31
Inject OnConnectionAttempt inside AcceptFilter

noboruma:
Add `net.Addr` to `AccepFilter` to allow for IP based rejections.


noboruma:
Dropped in favor of customer PacketListener

PR #1094, opened 2026-09-01
Implement DetachedConn API

JoTurk:
#### Description
Edit: this API initially only implemented the handshake phase, but now it's extended to full conn replacement. 

This API adds a transport-agnostic interface for driving a DTLS connection manually. It avoids exporting the intercept/inject API proposed in [#766](https://github.com/pion/dtls/pull/766), and callers do not need to parse DTLS flights or infer handshake batch boundaries, particularly difficult with DTLS 1.3.
The API is inspired by [crypto/tls.QUICConn](https://cs.opensource.google/go/go/+/go1.27.0:src/crypto/tls/quic.go;l=44).

- `DetachedClient` and `DetachedServer` create and configure a detached DTLS connection.
- `Start` begins the handshake and runs until peer input is required.
- `HandleDatagram` supplies a datagram received from the peer and processes it until DTLS is blocked again.
- `Write` encrypts application data after the handshake.
- `EventReady` signals that output is available.
- `NextEvent` returns complete output datagram batches, received application data, handshake completion, or connection termination.
- `Close` closes the connection.

Unlike the first proposal `DetachedConn` remains in use after the handshake, for pion/webrtc, this will replace packetconn inside the ice write, and removes the need for "mux".

~I didn't add tests for the API in cause we want to modify it through the review. But I'll add tests before it's merged.~

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -224,6 +224,7 @@ type Conn struct {
 
 	reading               chan struct{}
 	handshakeRecv         chan dtlshandshake.RecvHandshakeState
+	detached              *DetachedConn
 	cancelHandshaker      func()
 	cancelHandshakeReader func()
 
@@ -259,9 +260,8 @@ func createConn(nextConn net.PacketConn, rAddr net.Addr, config *dtlsConfig, isC
 }
 
 func newConn(nextConn net.PacketConn, rAddr net.Addr, configValues connConfigValues, handshakeConfig *dtlsconfig.HandshakeConfig, isClient bool) *Conn {
-	return &Conn{
+	conn := &Conn{
 		rAddr:                   rAddr,
-		nextConn:                netctx.NewPacketConn(nextConn),
 		handshakeConfig:         handshakeConfig,
 		fragmentBuffer:          dtlsfragmentbuffer.New(),
 		handshakeCache:          dtlsflight.NewCache(),
@@ -287,6 +287,11 @@ func newConn(nextConn net.PacketConn, rAddr net.Addr, configValues connConfigVal
 
 		state: dtlsstate.NewActive(isClient),
 	}
+	if nextConn != nil {
+		conn.nextConn = netctx.NewPacketConn(nextConn)
+	}
+
+	return conn
 }
 
 // Handshake runs the client or server DTLS handshake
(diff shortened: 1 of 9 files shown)


Sean-Der (approved):

zshang-oai:
@JoTurk, this is the new Codex writing style. I hope you like it. The content has been reviewed by a human to the best of his knowledge.

This should make SPED integration easier by moving packet-processing coordination into Pion and removing the need to switch interfaces when the handshake completes.

## What I would still clarify or improve

### 1. Processing a packet is not the same as producing the next handshake response

Suppose a STUN request contains one fragment of a larger DTLS handshake message.

DTLS can process that fragment without having enough information to advance the handshake. It might produce an acknowledgement—DTLS 1.3 permits this, and Pion has an ACK path—or there might be nothing ready to send yet.

The adapter should therefore:

1. Give the fragment to DTLS.
2. Collect whatever is ready, including possibly only an ACK.
3. Return to ICE so it can finish responding to the STUN request.

It should not wait for the rest of the handshake message while holding up the current STUN response.

**The updated `NextEvent` interface gives us the right mechanism:** drain the available events, then stop when it returns `DetachedNoEvent`.

This corrects my earlier wording. Saying DTLS “cannot produce a response” was too strong. The precise point is that **one received packet does not guarantee that the next handshake response is ready**.

It also narrows the review claim: fragmentation is something the API should handle and tests should cover. **It is not, by itself, evidence that ordinary, default SPED integration will fail.** That claim would need a concrete reproduction.

### 2. DTLS can produce output even when no new packet arrives

Imagine we send a handshake packet and it gets lost.

Nothing comes back, but DTLS still needs to wake up later and retransmit it. An adapter that only checks for output immediately after receiving input would miss that retransmission.

`EventReady` gives us a way to notice this. I would like a small example showing how the caller handles both incoming packets and timer-generated output.

For SPED, the example should also make clear who drains the events. We do not want a background worker taking the response packets that the STUN callback was about to include in its reply.

### 3. An error does not necessarily mean there is nothing left to send

Suppose DTLS rejects a peer’s handshake and creates an alert explaining why.

The operation may return an error, but that alert can still be waiting in the output queue. If our code immediately tears everything down, the peer never receives it and may keep retrying.

**The contract should explicitly say to drain events even when an operation returns an error.** Tests should check that the alert is available before the caller sees the final connection-closed event.

### 4. An output batch does not mean earlier packets were delivered

A batch tells us which outgoing packets are available together. It does not tell us that the peer received them—or that older, unacknowledged SPED packets can be discarded.

SPED still needs to track its own acknowledgements.

I would keep that distinction explicit in the documentation, so an integration does not accidentally treat a batch boundary as permission to clear its delivery cache.

### 5. The queue needs a clear expectation for slow callers

The event queue currently grows as events are added.

If the caller stops draining it, queued data can keep accumulating. The API should either provide a limit or clearly explain how the caller is expected to prevent that.

That is particularly relevant now that this interface carries application data too, rather than only a short handshake.

JoTurk:
@zshang-oai I believe 1, 2 and 3 are already implemented with the new API, I'll add tests. I believe 5 is out of the scope we can add an opt-in queue limit later (Pion APIs are usually unbounded by default). 
I'll look into 4 before this is merged.

Thank you.

JoTurk:
@zshang-oai I also believe that the 4th point doesn't require any code change, a batch in this API means "send these datagrams together." and nothing else, It's a transport-agnostic API (network-side Sans I/O? just not full Sans I/O because the timers are user controlled), SPED or other transports doesn't have or need to clear retained older batches or adopt their retention / retransmit to this API.
I can add a comment.

PR #1096, opened 2026-09-01
Optimize allocations for handshake and compaction

noboruma:
At scale (700k connections) - I detected some heavy allocations, especially at handshake.
This PR aims at addressing those points.


noboruma:
Depends on https://github.com/pion/transport/pull/396

conn.go
```diff
@@ -856,53 +856,108 @@ func (c *Conn) cacheHandshake(outbound *dtlsflight.Outbound, dtlsHandshake *hand
...
 	for _, record := range records {
-		if len(current.raw) > 0 && len(current.raw)+len(record.raw) >= c.maximumTransmissionUnit {
-			datagrams = append(datagrams, current)
-			current = preparedDatagram{}
+		recordSize := len(record.raw)
+
+		flatRaw = append(flatRaw, record.raw...)
+
+		if currentSize > 0 && currentSize+recordSize >= c.maximumTransmissionUnit {
+			datagrams[datagramIndex].raw = flatRaw[offset : offset+currentSize]
```
JoTurk: should bound the cap to prevent writing to the next datagram, this is important for the new DetachedConn API:
```suggestion
			end := offset + currentSize
			datagrams[datagramIndex].raw = flatRaw[offset:end:end]
```

conn.go
```diff
@@ -856,53 +856,108 @@ func (c *Conn) cacheHandshake(outbound *dtlsflight.Outbound, dtlsHandshake *hand
...
+			datagrams[datagramIndex].tracked = append(
+				datagrams[datagramIndex].tracked,
+				*record.tracked,
+			)
 		}
 	}
-	datagrams = append(datagrams, current)
 
-	return datagrams
+	datagrams[datagramIndex].raw = flatRaw[offset : offset+currentSize]
```
JoTurk: same as above:
```suggestion
	end := offset + currentSize
	datagrams[datagramIndex].raw = flatRaw[offset:end:end]
```

conn.go
```diff
@@ -1228,42 +1283,76 @@ func selectHandshakeFragment(offsets map[uint32]uint32, raw []byte) (bool, error
 	return ok && length == header.FragmentLength, nil
 }
 
+var noContentFragments = [][]byte{ //nolint:gochecknoglobals
+	{},
+}
+
 func (c *Conn) fragmentHandshake(dtlsHandshake *handshake.Handshake) ([][]byte, error) {
+	messageSize := dtlsHandshake.Message.MarshalSize()
+	numFragments := (messageSize-1)/c.maximumTransmissionUnit + 1
```
JoTurk: should bound this with max(1) to prevent panics, I don't think this can happen normally unless the user sets the mtu to 1 somehow or something, but you never know :)

```suggestion
	numFragments := max(1, (messageSize-1)/c.maximumTransmissionUnit+1)
```

JoTurk (commented):
Looks good, just 2 comments :)

conn.go
noboruma: Fair enough, we never know indeed :)

JoTurk (approved):
Thank you so much

Issue #10, opened 2018-11-20
Copy CI from `pions/webrtc`


Sean-Der:
Get code coverage and unit testing going

PR #1099, opened 2026-09-10
Reassemble re-fragmented handshake retransmits

tmatth:
#### Description

A retransmitted handshake message may be fragmented differently from the original (RFC 6347 Section 4.2.3). The fragment buffer kept the first fragment seen at each offset and only completed a message when the fragment lengths added up, so after one lost datagram the handshake never finished. OpenSSL servers with a small MTU, such as OvenMediaEngine, trigger this.

Keep the longer fragment at a cached offset, ignore fragments that fall outside the message, and complete a message once every byte is covered.

#### Reference issue
Fixing this would be a client-side mitigation for https://github.com/OvenMediaLabs/OvenMediaEngine/issues/2329 (tl;dr; OME forces DTLS handshakes to be split into 256 byte fragments due it defaulting to a small MTU, upon losing a fragment the handshake will fail for pion/dtls based clients).

### DIFF
--- a/fragment_buffer.go
+++ b/fragment_buffer.go
@@ -4,6 +4,9 @@
 package dtls
 
 import (
+	"slices"
+	"sort"
+
 	"github.com/pion/dtls/v3/pkg/protocol"
 	"github.com/pion/dtls/v3/pkg/protocol/handshake"
 	"github.com/pion/dtls/v3/pkg/protocol/recordlayer"
@@ -15,16 +18,79 @@ const (
 	fragmentBufferMaxCount = 1000
 )
 
-type fragment struct {
-	recordLayerHeader recordlayer.Header
-	handshakeHeader   handshake.Header
-	data              []byte
+// fragmentRange is the payload of one handshake fragment.
+type fragmentRange struct {
+	offset uint32
+	data   []byte
 }
 
+func (r fragmentRange) end() uint32 {
+	return r.offset + uint32(len(r.data)) //nolint:gosec // G115: fragment lengths are 24 bit.
+}
+
+// fragments holds what has been received of one handshake message. A peer may
+// fragment a retransmitted message differently from the original transmission
+// (RFC 6347 Section 4.2.3), so fragments can overlap. ranges is sorted by
+// offset and never holds a fragment that another one fully covers, which
+// makes its ends increase as well; covered is the length of the contiguous
+// prefix of the message received so far and is advanced as fragments arrive,
+// so completeness is known without scanning.
 type fragments struct {
-	fragmentByOffset map[uint32]*fragment
-	fragmentsLength  uint32
-	handshakeLength  uint32
+	ranges          []fragmentRange
+	covered         uint32
+	handshakeType   handshake.Type
+	handshakeLength uint32
+	// From the first fragment received at offset 0.
+	handshakeHeader handshake.Header
+	epoch           uint16
+	hasStart        bool
+}
+
+// add stores a fragment and returns the change in stored fragment count and
+// bytes, which is negative when the fragment replaces smaller ones.
+func (f *fragments) add(offset uint32, data []byte) (fragmentDelta, byteDelta int) {
+	frag := fragmentRange{offset: offset, data: data}
+
+	// First fragment starting at or after offset.
+	i := sort.Search(len(f.ranges), func(k int) bool { return f.ranges[k].offset >= offset })
+	if f.covers(i, frag) {
+		return 0, 0
+	}
+
+	// ranges[i:j] start inside the new fragment and end inside it: covered.
+	j := i
+	for j < len(f.ranges) && f.ranges[j].end() <= frag.end() {
+		fragmentDelta--
+		byteDelta -= len(f.ranges[j].data)
+		j++
+	}
+	f.ranges = slices.Replace(f.ranges, i, j, frag)
+	f.advanceCovered(i)
+
+	return fragmentDelta + 1, byteDelta + len(data)
+}
+
+// covers reports whether a stored fragment already carries every byte of
+// frag, whose insertion index is i. Since ends increase with offsets, only
+// the fragment before i and one starting at the same offset can.
+func (f *fragments) covers(i int, frag fragmentRange) bool {
+	if i > 0 && f.ranges[i-1].end() >= frag.end() {
+		return true
+	}
+
+	return i < len(f.ranges) && f.ranges[i].offset == frag.offset && f.ranges[i].end() >= frag.end()
+}
+
+// advanceCovered extends the covered prefix after inserting ranges[i]. Only
+// a fragment touching the prefix can extend it, possibly through fragments
+// received earlier that start within its reach.
+func (f *fragments) advanceCovered(i int) {
+	if f.ranges[i].offset > f.covered {
+		return
+	}
+	for k := i; k < len(f.ranges) && f.ranges[k].offset <= f.covered; k++ {
+		f.covered = max(f.covered, f.ranges[k].end())
+	}
 }
 
 type fragmentBuffer struct {
@@ -49,6 +115,7 @@ func (f *fragmentBuffer) size() int {
 // Attempts to push a DTLS packet to the fragmentBuffer
 // when it returns true it means the fragmentBuffer has inserted and the buffer shouldn't be handled
 // when an error returns it is fatal, and the DTLS connection should be stopped.
+// The fragments keep referencing buf, so it must not be reused by the caller.
 func (f *fragmentBuffer) push(buf []byte) (isHandshake, isRetransmit bool, err error) { //nolint:cyclop
 	if f.size()+len(buf) >= fragmentBufferMaxSize || f.totalFragmentCount >= fragmentBufferMaxCount {
 		return false, false, errFragmentBufferOverflow
@@ -64,45 +131,50 @@ func (f *fragmentBuffer) push(buf []byte) (isHandshake, isRetransmit bool, err e
 		return false, false, nil
 	}
 
-	frag := new(fragment)
-	for buf = buf[recordlayer.FixedHeaderSize:]; len(buf) != 0; frag = new(fragment) { //nolint:gosec // G602
-		if err := frag.handshakeHeader.Unmarshal(buf); err != nil {
+	for buf = buf[recordlayer.FixedHeaderSize:]; len(buf) != 0; { //nolint:gosec // G602
+		var header handshake.Header
+		if err := header.Unmarshal(buf); err != nil {
 			return false, false, err
 		}
 
 		// Fragment is a retransmission. We have already assembled it before successfully
-		isRetransmit = frag.handshakeHeader.FragmentOffset == 0 &&
-			frag.handshakeHeader.MessageSequence < f.currentMessageSequenceNumber
+		isRetransmit = header.FragmentOffset == 0 && header.MessageSequence < f.currentMessageSequenceNumber
 
-		end := int(handshake.HeaderLength + frag.handshakeHeader.FragmentLength)
+		end := int(handshake.HeaderLength + header.FragmentLength)
 		if end > len(buf) {
 			return false, false, errBufferTooSmall
 		}
-		if frag.handshakeHeader.MessageSequence < f.currentMessageSequenceNumber {
-			buf = buf[end:]
+		data := buf[handshake.HeaderLength:end]
+		buf = buf[end:]
 
+		// Already assembled, or claims bytes outside the message.
+		if header.MessageSequence < f.currentMessageSequenceNumber ||
+			header.FragmentOffset > header.Length ||
+			header.FragmentLength > header.Length-header.FragmentOffset {
 			continue
 		}
 
-		messageFragments, ok := f.cache[frag.handshakeHeader.MessageSequence]
+		messageFragments, ok := f.cache[header.MessageSequence]
 		if !ok {
-			messageFragments = &fragments{
-				fragmentByOffset: map[uint32]*fragment{}, handshakeLength: frag.handshakeHeader.Length,
-			}
-			f.cache[frag.handshakeHeader.MessageSequence] = messageFragments
+			messageFragments = &fragments{handshakeType: header.Type, handshakeLength: header.Length}
+			f.cache[header.MessageSequence] = messageFragments
+		} else if header.Type != messageFragments.handshakeType || header.Length != messageFragments.handshakeLength {
+			// Disagrees with the message's earlier fragments; admitting it
+			// would break the bounds pop relies on.
+			continue
+		}
+		if header.FragmentOffset == 0 && !messageFragments.hasStart {
+			messageFragments.handshakeHeader = header
+			messageFragments.epoch = recordLayerHeader.Epoch
+			messageFragments.hasStart = true
+		}
+		if len(data) == 0 {
+			continue
 		}
 
-		// Discard all headers, when rebuilding the packet we will re-build
-		frag.data = append([]byte{}, buf[handshake.HeaderLength:end]...)
-		frag.recordLayerHeader = recordLayerHeader
-
-		if _, ok = messageFragments.fragmentByOffset[frag.handshakeHeader.FragmentOffset]; !ok {
-			messageFragments.fragmentByOffset[frag.handshakeHeader.FragmentOffset] = frag
-			messageFragments.fragmentsLength += frag.handshakeHeader.FragmentLength
-			f.totalBufferSize += int(frag.handshakeHeader.FragmentLength)
-			f.totalFragmentCount++
-		}
-		buf = buf[end:]
+		fragmentDelta, byteDelta := messageFragments.add(header.FragmentOffset, data)
+		f.totalFragmentCount += fragmentDelta
+		f.totalBufferSize += byteDelta
 	}
 
 	return true, isRetransmit, nil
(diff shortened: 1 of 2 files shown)


JoTurk:
Hello @tmatth we don't tag from main until we finish `v4` which is a bit delayed because of all the remaining work, do you want to target `v3` and i can port it to `v4` later?

JoTurk:
v3 goes in https://github.com/pion/dtls/tree/v3 we tagged it recently.

fragment_buffer.go
```diff
@@ -21,12 +21,68 @@ type fragment struct {
...
+
+// assemble concatenates the cached fragments into the message body. A peer
+// may fragment a retransmitted message differently from the original
+// transmission (RFC 6347 Section 4.2.3), so the cached fragments can overlap
+// and the message is complete once every byte is covered rather than when the
+// fragment lengths add up to the message length. A complete message always
+// has a fragment at offset 0: the one covering byte 0, or for an empty
+// message the only fragment its bounds allow.
+func (f *fragments) assemble() ([]byte, bool) {
+	body := make([]byte, 0, f.handshakeLength)
```
JoTurk: Do we need to introduce this copy? we have been trying to reduce allocation through network path to record layers.

tmatth:
The close/reopen was just to unwedge CI :)

fragment_buffer.go
```diff
@@ -21,12 +21,68 @@ type fragment struct {
...
+	for offset := uint32(0); offset < f.handshakeLength; {
+		frag := f.covering(offset)
+		if frag == nil {
```
JoTurk: and `covering(o)` scans the entire fragment map, this seems rather expensive? and not constant time anymore. can we make push increment coverage, instead of computing it everytime?

JoTurk (commented):
Thank you, just two comments

JoTurk:
> The close/reopen was just to unwedge CI :)

this github bug is really annoying :)

fragment_buffer.go
tmatth: The old `pop` already built the message through an append-grown `rawMessage` and then copied it again into `append(rawHeader, rawMessage...)`, so this wasn't a new copy, but it was one more than needed. Reworked so `pop` allocates the output once (header length + message length) and appends the fragments straight into it, and `push` no longer copies each fragment's payload either.

fragment_buffer.go
tmatth: Yeah agreed, addressing this one expanded the scope of the PR but is worth it.

tmatth:
Here's a reproducer to show the behavior I observed IRL (source files at the bottom), also it is quite a few steps (server, client, certs, etc.):

## Reproducer

Both clients below wrap their UDP socket so every DTLS handshake datagram from the server is logged, and the Nth one can be dropped to make the server retransmit its flight. The server side re-fragments on retransmit, and stock pion/dtls (v3.1.8) never completes the handshake.

### A. Real-world: OvenMediaEngine over WHIP (docker + pion/webrtc)

OvenMediaEngine's DTLS BIO never answers `BIO_CTRL_DGRAM_QUERY_MTU`, so OpenSSL uses its 256-byte minimum and every server flight is fragmented (https://github.com/OvenMediaLabs/OvenMediaEngine/issues/2329).

```sh
# 1. OME with its stock config
docker run --rm -p 3333:3333 -p 10000:10000/udp \
  -e OME_HOST_IP=127.0.0.1 -e OME_WEBRTC_CANDIDATE_PORT=10000/udp \
  airensoft/ovenmediaengine:v0.21.0

# 2. in another shell: WHIP publisher (whip-repro/main.go below), dropping the 2nd server datagram
mkdir whip-repro && cd whip-repro      # save main.go
go mod init whip-repro && go get github.com/pion/webrtc/v4@v4.2.19 && go mod tidy
go run . -drop 2
```

Result with stock pion/dtls: OME retransmits the Certificate as `[0,231)` + `[231,282)` after originally sending `[0,108)` + `[108,282)`; the client sits in `connecting` with ICE connected while OME retransmits at 1s, 2s, 4s, 10s:

```
DTLS <- 127.0.0.1:10000   256B  #1  hs type=2 seq=0 len=98 frag=[0,98) | hs type=11 seq=1 len=282 frag=[0,108)
DTLS <- 127.0.0.1:10000   256B  #2  DROPPED: hs type=11 seq=1 len=282 frag=[108,282) | hs type=12 seq=2 len=112 frag=[0,32)
DTLS <- 127.0.0.1:10000   203B  #3  hs type=12 seq=2 len=112 frag=[32,112) | hs type=13 seq=3 len=48 frag=[0,48) | hs type=14 seq=4 len=0 frag=[0,0)
DTLS <- 127.0.0.1:10000   123B  #4  hs type=2 seq=0 len=98 frag=[0,98)
DTLS <- 127.0.0.1:10000   256B  #5  hs type=11 seq=1 len=282 frag=[0,231)
DTLS <- 127.0.0.1:10000    76B  #6  hs type=11 seq=1 len=282 frag=[231,282)
...
RESULT: NOT CONNECTED after 25s, pc=connecting ice=connected (server handshake datagrams: 27 received, 1 dropped)
```

With the fix (`go mod edit -replace github.com/pion/dtls/v3=/path/to/fixed/branch`): `RESULT: connected (server handshake datagrams: 10 received, 1 dropped)`.

### B. Standalone: pion/dtls client against `openssl s_server` (no WebRTC, no docker)

```sh
# 1. throwaway certificate
openssl req -x509 -newkey ec -pkeyopt ec_paramgen_curve:prime256v1 -nodes \
  -keyout key.pem -out cert.pem -subj /CN=repro -days 1

# 2. OpenSSL DTLS server with a 256-byte MTU (restart it between runs, -naccept 1)
openssl s_server -dtls1_2 -accept 4444 -mtu 256 -cert cert.pem -key key.pem -naccept 1 -quiet

# 3. in another shell: the client (dtls-repro/main.go below), dropping the 2nd server datagram
mkdir dtls-repro && cd dtls-repro      # save main.go
go mod init dtls-repro && go get github.com/pion/dtls/v3@v3.1.8
go run . -drop 2
```

Stock v3.1.8: the retransmitted Certificate arrives as `[0,203)` + `[203,382)`. `[0,203)` is dropped because offset 0 is already cached, `[203,382)` is stored at a new offset, and the summed fragment lengths (561) can never equal the message length (382):

```
DTLS <-   48B  #1  hs type=3 seq=0 len=23 frag=[0,23)
DTLS <-  228B  #2  DROPPED: hs type=2 seq=1 len=89 frag=[0,89) | hs type=11 seq=2 len=382 frag=[0,89)
DTLS <-  228B  #3  hs type=11 seq=2 len=382 frag=[89,292)
DTLS <-  228B  #4  hs type=11 seq=2 len=382 frag=[292,382) | hs type=12 seq=3 len=111 frag=[0,88)
DTLS <-   73B  #5  hs type=12 seq=3 len=111 frag=[88,111) | hs type=14 seq=4 len=0 frag=[0,0)
DTLS <-  114B  #6  hs type=2 seq=1 len=89 frag=[0,89)
DTLS <-  228B  #7  hs type=11 seq=2 len=382 frag=[0,203)
DTLS <-  204B  #8  hs type=11 seq=2 len=382 frag=[203,382)
...
RESULT: handshake failed after 20s: handshake error: context deadline exceeded
```

(type 3 = HelloVerifyRequest, 2 = ServerHello, 11 = Certificate, 12 = ServerKeyExchange, 13 = CertificateRequest, 14 = ServerHelloDone)

With the fix, the same run ends `RESULT: handshake completed` right after datagram #10.

<details>
<summary>whip-repro/main.go (pion/webrtc v4.2.19)</summary>

```go
// WHIP publish against an OvenMediaEngine endpoint that logs every DTLS
// handshake datagram received from the server and can drop the Nth one to
// force OpenSSL to retransmit its flight.
package main

import (
	"bytes"
	"context"
	"encoding/binary"
	"flag"
	"fmt"
	"io"
	"log"
	"net"
	"net/http"
	"os"
	"strings"
	"sync/atomic"
	"time"

	"github.com/pion/webrtc/v4"
)

type dtlsLoggingConn struct {
	net.PacketConn
	dropNth  int
	received atomic.Int32
	dropped  atomic.Int32
}

func describeRecords(datagram []byte) (string, bool) {
	var parts []string
	handshake := false
	for rest := datagram; len(rest) >= 13; {
		contentType := rest[0]
		if rest[1] != 0xfe || (rest[2] != 0xfd && rest[2] != 0xff) {
			return "", false
		}
		epoch := binary.BigEndian.Uint16(rest[3:5])
		length := int(binary.BigEndian.Uint16(rest[11:13]))
		if 13+length > len(rest) {
			return "", false
		}
		body := rest[13 : 13+length]
		switch {
		case contentType == 22 && epoch == 0 && len(body) >= 12:
			handshake = true
			msgType := body[0]
			msgLen := int(body[1])<<16 | int(body[2])<<8 | int(body[3])
			seq := binary.BigEndian.Uint16(body[4:6])
			off := int(body[6])<<16 | int(body[7])<<8 | int(body[8])
			fragLen := int(body[9])<<16 | int(body[10])<<8 | int(body[11])
			parts = append(parts, fmt.Sprintf("hs type=%d seq=%d len=%d frag=[%d,%d)", msgType, seq, msgLen, off, off+fragLen))
		case contentType == 22:
			handshake = true
			parts = append(parts, fmt.Sprintf("hs epoch=%d (encrypted, %dB)", epoch, length))
		case contentType == 20:
			parts = append(parts, "change_cipher_spec")
		case contentType == 21:
			parts = append(parts, fmt.Sprintf("alert (%dB)", length))
		default:
			parts = append(parts, fmt.Sprintf("type=%d (%dB)", contentType, length))
		}
		rest = rest[13+length:]
	}
	return strings.Join(parts, " | "), handshake
}

func (c *dtlsLoggingConn) ReadFrom(b []byte) (int, net.Addr, error) {
	for {
		n, addr, err := c.PacketConn.ReadFrom(b)
		if err != nil {
			return n, addr, err
		}
		desc, isHandshake := describeRecords(b[:n])
		if !isHandshake {
			return n, addr, err
		}
		k := int(c.received.Add(1))
		if k == c.dropNth {
			c.dropped.Add(1)
			log.Printf("DTLS <- %s  %4dB  #%d  DROPPED: %s", addr, n, k, desc)
			continue
		}
		log.Printf("DTLS <- %s  %4dB  #%d  %s", addr, n, k, desc)
		return n, addr, err
	}
}

func main() {
	whip := flag.String("whip", "http://127.0.0.1:3333/app/{stream}?direction=whip", "WHIP endpoint ({stream} becomes a fresh name per run)")
	drop := flag.Int("drop", 0, "drop the Nth DTLS handshake datagram received from the server (0 = none)")
	timeout := flag.Duration("timeout", 30*time.Second, "give up if not connected after this long")
	flag.Parse()
	*whip = strings.ReplaceAll(*whip, "{stream}", fmt.Sprintf("repro%d", time.Now().UnixNano()%1000000))
	log.SetFlags(log.Ltime | log.Lmicroseconds)

	udp, err := net.ListenUDP("udp4", &net.UDPAddr{IP: net.IPv4zero})
	if err != nil {
		log.Fatal(err)
	}
	conn := &dtlsLoggingConn{PacketConn: udp, dropNth: *drop}

	se := webrtc.SettingEngine{}
	se.SetICEUDPMux(webrtc.NewICEUDPMux(nil, conn))
	se.SetNetworkTypes([]webrtc.NetworkType{webrtc.NetworkTypeUDP4})
	se.SetInterfaceFilter(func(name string) bool {
		for _, p := range []string{"virbr", "docker", "veth", "br-", "cni", "flannel", "lxc"} {
			if strings.HasPrefix(name, p) {
				return false
			}
		}
		return true
	})
	api := webrtc.NewAPI(webrtc.WithSettingEngine(se))
	pc, err := api.NewPeerConnection(webrtc.Configuration{})
	if err != nil {
		log.Fatal(err)
	}
	defer pc.Close()

	video, err := webrtc.NewTrackLocalStaticSample(webrtc.RTPCodecCapability{MimeType: webrtc.MimeTypeH264}, "video", "repro")
	if err != nil {
		log.Fatal(err)
	}
	audio, err := webrtc.NewTrackLocalStaticSample(webrtc.RTPCodecCapability{MimeType: webrtc.MimeTypeOpus}, "audio", "repro")
	if err != nil {
		log.Fatal(err)
	}
	if _, err = pc.AddTrack(video); err != nil {
		log.Fatal(err)
	}
	if _, err = pc.AddTrack(audio); err != nil {
		log.Fatal(err)
	}

	connected := make(chan struct{}, 1)
	pc.OnConnectionStateChange(func(s webrtc.PeerConnectionState) {
		log.Printf("PeerConnection state: %s", s)
		if s == webrtc.PeerConnectionStateConnected {
			connected <- struct{}{}
		}
	})
	pc.OnICEConnectionStateChange(func(s webrtc.ICEConnectionState) { log.Printf("ICE state: %s", s) })

	offer, err := pc.CreateOffer(nil)
	if err != nil {
		log.Fatal(err)
	}
	gathered := webrtc.GatheringCompletePromise(pc)
	if err = pc.SetLocalDescription(offer); err != nil {
		log.Fatal(err)
	}
	<-gathered

	ctx, cancel := context.WithTimeout(context.Background(), *timeout)
	defer cancel()
	req, _ := http.NewRequestWithContext(ctx, http.MethodPost, *whip, bytes.NewBufferString(pc.LocalDescription().SDP))
	req.Header.Set("Content-Type", "application/sdp")
	resp, err := http.DefaultClient.Do(req)
	if err != nil {
		log.Fatal(err)
	}
	body, _ := io.ReadAll(resp.Body)
	resp.Body.Close()
	if resp.StatusCode != http.StatusCreated {
		log.Fatalf("WHIP POST: %s: %s", resp.Status, body)
	}
	log.Printf("WHIP POST %s, resource %s", resp.Status, resp.Header.Get("Location"))
	if loc := resp.Header.Get("Location"); loc != "" {
		defer func() {
			u := loc
			if strings.HasPrefix(u, "/") {
				base := *whip
				if i := strings.Index(base, "/app"); i > 0 {
					u = base[:i] + loc
				}
			}
			r, _ := http.NewRequest(http.MethodDelete, u, nil)
			if d, err := http.DefaultClient.Do(r); err == nil {
				d.Body.Close()
			}
		}()
	}
	if err = pc.SetRemoteDescription(webrtc.SessionDescription{Type: webrtc.SDPTypeAnswer, SDP: string(body)}); err != nil {
		log.Fatal(err)
	}

	select {
	case <-connected:
		log.Printf("RESULT: connected (server handshake datagrams: %d received, %d dropped)", conn.received.Load(), conn.dropped.Load())
	case <-ctx.Done():
		log.Printf("RESULT: NOT CONNECTED after %s, pc=%s ice=%s (server handshake datagrams: %d received, %d dropped)",
			*timeout, pc.ConnectionState(), pc.ICEConnectionState(), conn.received.Load(), conn.dropped.Load())
		os.Exit(1)
	}
}
```

</details>

<details>
<summary>dtls-repro/main.go (pion/dtls v3)</summary>

```go
// DTLS client that logs every handshake datagram received from the server and
// can drop the Nth one, to force the server to retransmit its flight.
package main

import (
	"context"
	"encoding/binary"
	"flag"
	"fmt"
	"log"
	"net"
	"os"
	"strings"
	"sync/atomic"
	"time"

	"github.com/pion/dtls/v3"
)

type droppingConn struct {
	net.PacketConn
	dropNth  int
	received atomic.Int32
}

func describe(datagram []byte) (string, bool) {
	var parts []string
	handshake := false
	for rest := datagram; len(rest) >= 13; {
		if rest[1] != 0xfe || (rest[2] != 0xfd && rest[2] != 0xff) {
			return "", false
		}
		contentType, epoch := rest[0], binary.BigEndian.Uint16(rest[3:5])
		length := int(binary.BigEndian.Uint16(rest[11:13]))
		if 13+length > len(rest) {
			return "", false
		}
		body := rest[13 : 13+length]
		switch {
		case contentType == 22 && epoch == 0 && len(body) >= 12:
			handshake = true
			msgLen := int(body[1])<<16 | int(body[2])<<8 | int(body[3])
			off := int(body[6])<<16 | int(body[7])<<8 | int(body[8])
			fragLen := int(body[9])<<16 | int(body[10])<<8 | int(body[11])
			parts = append(parts, fmt.Sprintf("hs type=%d seq=%d len=%d frag=[%d,%d)",
				body[0], binary.BigEndian.Uint16(body[4:6]), msgLen, off, off+fragLen))
		case contentType == 22:
			handshake = true
			parts = append(parts, fmt.Sprintf("hs epoch=%d (encrypted, %dB)", epoch, length))
		case contentType == 20:
			parts = append(parts, "change_cipher_spec")
		default:
			parts = append(parts, fmt.Sprintf("type=%d (%dB)", contentType, length))
		}
		rest = rest[13+length:]
	}
	return strings.Join(parts, " | "), handshake
}

func (c *droppingConn) ReadFrom(b []byte) (int, net.Addr, error) {
	for {
		n, addr, err := c.PacketConn.ReadFrom(b)
		if err != nil {
			return n, addr, err
		}
		desc, isHandshake := describe(b[:n])
		if !isHandshake {
			return n, addr, err
		}
		k := int(c.received.Add(1))
		if k == c.dropNth {
			log.Printf("DTLS <- %4dB  #%d  DROPPED: %s", n, k, desc)
			continue
		}
		log.Printf("DTLS <- %4dB  #%d  %s", n, k, desc)
		return n, addr, err
	}
}

func main() {
	server := flag.String("server", "127.0.0.1:4444", "DTLS server address")
	drop := flag.Int("drop", 0, "drop the Nth handshake datagram received from the server (0 = none)")
	timeout := flag.Duration("timeout", 20*time.Second, "handshake timeout")
	flag.Parse()
	log.SetFlags(log.Ltime | log.Lmicroseconds)

	raddr, err := net.ResolveUDPAddr("udp", *server)
	if err != nil {
		log.Fatal(err)
	}
	udp, err := net.ListenUDP("udp", &net.UDPAddr{IP: net.IPv4zero})
	if err != nil {
		log.Fatal(err)
	}
	conn, err := dtls.Client(&droppingConn{PacketConn: udp, dropNth: *drop}, raddr, &dtls.Config{InsecureSkipVerify: true})
	if err != nil {
		log.Fatal(err)
	}
	defer conn.Close()

	ctx, cancel := context.WithTimeout(context.Background(), *timeout)
	defer cancel()
	if err := conn.HandshakeContext(ctx); err != nil {
		log.Printf("RESULT: handshake failed after %s: %v", *timeout, err)
		os.Exit(1)
	}
	log.Printf("RESULT: handshake completed")
}
```

</details>

fragment_buffer.go
```diff
@@ -64,88 +153,77 @@ func (f *fragmentBuffer) push(buf []byte) (isHandshake, isRetransmit bool, err e
...
-
-	f.totalBufferSize -= int(frags.fragmentsLength)
-	f.totalFragmentCount -= len(frags.fragmentByOffset)
+	// The message is contiguous, so each fragment continues at or before
+	// where the previous one ended.
+	content = make([]byte, 0, len(rawHeader)+int(frags.handshakeLength))
+	content = append(content, rawHeader...)
+	cursor := uint32(0)
+	for _, frag := range frags.ranges {
+		content = append(content, frag.data[cursor-frag.offset:]...)
```
JoTurk, line 201: Since we removed the return on gap https://github.com/pion/dtls/blob/474c3e2e7a4e97ee83ff57ae550cf6e4f717dcfa/internal/fragmentbuffer/fragment_buffer.go#L299-L301

this can cause a panic if the length in header doesn't match the fragment length, we can return the gap check again or better if we add a `header.Length != messageFragments.handshakeLength` or similar check in `push` before we admit cache

JoTurk (commented):
thank you for all the fixes, just one comment before we can merge i think

fragment_buffer.go
tmatth: Thanks, I've pushed a fix for this and a test to cover it.

tmatth:
I'm going to do one more pass to clean-up some of the more handrolled looping etc. (to use stdlib idioms instead).

**UPDATE:** DONE, now it's a bit shorter/cleaner.

tmatth:
`fix / Fix go.sumExpected — Waiting for status to be reported` seems stuck, going to close/open the PR again.

JoTurk:
This check is just broken since we don't target main don't worry about it, I'll re review this pr in like an hour thank you.

JoTurk (approved):
Thank you, I see two out of scope nits, I'll open PRs and port it to v4

Issue #1053, opened 2026-08-22
Record-layer API rework, ownership refactor


JoTurk:
This is the review summary for the problems in the current DTLS public record-layer API in `pkg/protocol/recordlayer` and all the related record layer code in conn and handshake, this is a part of #1007  

1. Multiple unpacking APIs `UnpackDatagram` and `ContentAwareUnpackDatagram` for 1.2, and `UnpackDatagram13` for 13. These APIs have duplicated logic and behave differently. And can be refactored into a single API with options.
2. incorrect datagram format selection, the first record can determine the parse for the whole datagram even tho DTLS 1.3 datagrams may mix plaintext and ciphertext records. This issue becomes visible with interop.
3. framing is mixed with no direct input version policy, raw record parsing decides whether 1.2 or 1.3 records are accepted.
4. CID length is inferred from the input preallocated ConnectionID, which is confusing for a public API.
5. Content type whitelists are all over the place, some even exist outside of the API in conn.
6. Records are parsed many times, datagram splitting, cid, connection processing etc parse the same header again.
7. no common record format, each format has a different path.
8. public types are mutable and allows invalid combinations.
9. marshal mutates the input objects!
10. ^ DTLS 1.3 wire format isn't preserved after marshal.
11. 8-bit sequence numbers can be truncated (maybe we should only support 16-bit SNs like boringssl).
12. invalid epoch bits can be silently masked.
13. there is no bound validation for payload sizes.
14. ^ zero-length records are rejected at the framer.
15. ^ content decoding isn't always restricted to the exact declared record length and consume trailing data (extensions had this issue).
16. returned records and CIDs sometimes alias the receive datagram without documentation for this lifetime behavior.
18. ^ some fields are copied and some are borrowed.
19. CID encoding logic is duplicated many times.
20. conn owns many record-layer behavior, framing, version validation, cid routing decryption, reply handling, content dispatch etc
21. ^ protocol-rule ownership is unclear, records can be rejected at different layers and it's not clear which layer owns the rules.
22. ^ every new record content-type requires widespread edits and it's hard to review and test. When I was adding RRC I had to touch many several layers from conn, to record layer, to handshake, to flights.

PR #1103, opened 2026-09-13
Continue receiving after truncated datagrams

JoTurk:
#### Description
Treat short-buffer errors as recoverable in reads, https://github.com/pion/dtls/issues/642 reported that Pion behavior is undocumented. And https://github.com/pion/dtls/pull/1016 made the buffer configurable, but we currently can terminate the connection over over-sized datagrams. this discards the remaining read and consumes the prefix datagrams if they fit the buffer.

part of #1053


JoTurk:
https://github.com/pion/dtls/actions/runs/34749645049/job/103703680631 windows-related bug :)

edit: turned out Go's udp.read can propagate `WSAEMSGSIZE` to the user code on Windows :) fixed with https://github.com/pion/dtls/pull/1103/commits/f998c2caed589f58722f0a457fbb786414e9d4ec luckily I was able to test it with Wine :)

treejadey (approved):

PR #1122, opened 2026-09-17
Replace mutable record-layer with parse/marshal

### LINKED ISSUE #1053: Record-layer API rework, ownership refactor

JoTurk:
#### Description
I have been working on this past a few weeks this is the final work to finish the record-layer API for v4, and fixes DTLS 1.3 handling.
The DTLS 1.3 integration with original record-layer API was "hacky" at best rebuilding a parsed header could change sequence width or length presence, and separate header and record wrappers caused more code and duplicated handling.
Remove Header, UnifiedHeader, RecordLayer, CiphertextRecord, and InnerPlaintext in favor of borrowed parsed records and pure encoders that are DTLS 1.3 friendly. With clear ownership

This is a logic migration work, it migrates connection handling, CID routing, protection metadata APIs etc and their tests while preservers the actual record-layer code.

#### Reference issue

closes #1053


Sean-Der (approved):
🔥 here comes v4!

PR #1133, opened 2026-09-21
Add DTLS 1.3 keying-material exporter

### LINKED ISSUE #1098: DTLS 1.3: implement `State.ExportKeyingMaterial`
DTLS 1.3 is still in progress, and the public key-export API appears to be missing its 1.3 implementation.

[#914](https://github.com/pion/dtls/pull/914) added internal exporter-master-secret derivation, and [#926](https://github.com/pion/dtls/pull/926) added coverage for that derivation. However, this isn’t connected to `State.ExportKeyingMaterial` yet.

### Current behavior

On `main` at `2a3670e7`:

- [`generateState13`](https://github.com/pion/dtls/blob/2a3670e7defc11bb332035eb8ca1d08136894b0e/state.go#L114-L149) constructs the public state without preserving the DTLS 1.3 exporter master secret. It also leaves the legacy `masterSecret` field empty.
- [`ExportKeyingMaterial`](https://github.com/pion/dtls/blob/2a3670e7defc11bb332035eb8ca1d08136894b0e/state.go#L352-L380) does not branch by protocol version. It uses the DTLS 1.2 `PHash` calculation with that empty `masterSecret`.

Consequently, after a successful DTLS 1.3 handshake, a call such as:

```go
state.ExportKeyingMaterial("EXTRACTOR-dtls_srtp", nil, 60)
```

returns bytes without an error, but those bytes are not a valid DTLS 1.3 export. They are derived from an empty secret, the label, and the public ClientHello/ServerHello random values.

This matters for applications using exported material as encryption keys, including DTLS-SRTP. It is separate from the keys used to encrypt DTLS records.

### Expected behavior

Use the DTLS 1.3 exporter construction from [RFC 8446 §7.5](https://www.rfc-editor.org/rfc/rfc8446.html#section-7.5), with the DTLS-specific label prefix from [RFC 9147 §5.9](https://www.rfc-editor.org/rfc/rfc9147.html#section-5.9).

Until implemented, explicitly rejecting key export for DTLS 1.3 would be safer than returning invalid material successfully.

### Suggested coverage

- Test the public `State.ExportKeyingMaterial` API after an actual DTLS 1.3 handshake.
- Compare exported bytes against an independent implementation such as BoringSSL, including the SRTP label.
- Preserve DTLS 1.2 behavior.

Related: #736, #1000.

enobufs:
## What problem does this PR solve?
ExportKeyingMaterial previously always used the DTLS 1.2 PRF over the master secret. On a DTLS 1.3 connection that secret is not populated, so exported keying material (e.g. EXTRACTOR-dtls_srtp for DTLS-SRTP) could not be derived.

## Why is this change needed?
State.ExportKeyingMaterial (RFC 5705) is what WebRTC uses to key DTLS-SRTP. It was DTLS 1.2-only, so on a DTLS 1.3 connection SRTP could not be keyed. This blocks DTLS 1.3 for any SRTP/WebRTC use.

## Any related specifications or other implementations, if applicable.
* Demo: Pion (w/ the fix) <-> Browser using DTLS v1.3 - demonstrates the problem this fixes: https://github.com/enobufs/dtls13-webrtc-demo

Closes #1098 part of #1000


JoTurk:
This is part of the todo in #1000 and "reported" by #1098, I was going to implement this today, Thank you.

state.go
```diff
@@ -451,6 +477,37 @@ func (s *State) ExportKeyingMaterial(label string, context []byte, length int) (
...
+	}
+
+	h := hashFunc()
+	if _, err := h.Write(context); err != nil {
+		return nil, err
+	}
+
+	return keyschedule.HkdfExpandLabel(hashFunc, exporterSecret, "exporter", h.Sum(nil), length)
+}
+
```
JoTurk: Shouldn't use `cipherSuiteDescriptor` directly, should normlize the code between 1.2 and 1.3 and call .cipherSuite() => .HashFunc()

https://github.com/pion/dtls/blob/c2c2da0c6525077b7715077727f06ff6db9c4b6f/internal/ciphersuite/ciphersuite.go#L35

state.go
```diff
@@ -451,6 +477,37 @@ func (s *State) ExportKeyingMaterial(label string, context []byte, length int) (
...
+// exportKeyingMaterialHKDF implements the HKDF-based TLS 1.3 exporter (RFC 8446
+// section 7.5), as used by DTLS 1.3 (RFC 9147). For a given label and context it
+// computes:
+//
+//	Derive-Secret(exporter_master_secret, label, "")             -> secret
+//	HKDF-Expand-Label(secret, "exporter", Hash(context), length) -> keying material
+//
+// HKDF-Expand-Label uses RFC 9147 section 5.9's "dtls13" label prefix (applied
+// by the keyschedule package). Unlike the DTLS 1.2 PRF path, the exporter output
+// does not depend on the endpoint role or the handshake randoms.
```
JoTurk: > // does not depend on the endpoint role or the handshake randoms.

I don't think this comment is correct, exporter_master_secret is still derived using cienthello...server Finished, i would try to avoid adding too many comments not directly referenced from the spec :)

state.go
```diff
@@ -41,6 +42,12 @@ type State struct {
 	NegotiatedProtocol string
 	// KeyUsage is the current record keys' usage.
 	KeyUsage *KeyUsageStats
+	// exporterMasterSecret is the DTLS 1.3 exporter_master_secret
+	// (RFC 8446 section 7.5, inherited by RFC 9147). It is only populated for a
+	// DTLS 1.3 connection and is consumed by ExportKeyingMaterial. It is not
+	// serialized (DTLS 1.3 state serialization is unsupported). The hash used to
+	// derive from it comes from cipherSuiteDescriptor.
```
JoTurk: nit: shouldn't describe implementation details.

JoTurk (changes requested):
Thank you, It's great to see you back :)

state.go
```diff
@@ -433,6 +441,24 @@ func (s *State) ExportKeyingMaterial(label string, context []byte, length int) (
...
 	}
+
+	// DTLS 1.3 derives exported keying material with the HKDF-based TLS 1.3
+	// exporter (RFC 8446 section 7.5, inherited by RFC 9147); DTLS 1.2 uses the
+	// legacy TLS 1.2 PRF (RFC 5705). Note DTLS version numbers decrease as the
+	// version increases (1.2 = 0xfefd, 1.3 = 0xfefc), so this is an exact-version
+	// dispatch, not an ordered comparison.
+	if s.version == protocol.Version1_3 {
+		return s.exportKeyingMaterialHKDF(label, context, length)
+	}
```
JoTurk, line 442: Should move this check before
https://github.com/pion/dtls/blob/76aad1b826e5f2f38a943dafb56ea53af472326f/state.go#L439-L441

Since the 1.3 implementation doesn't have the same limitation as 1.2 (with optional context).

Sean-Der:
Welcome back @enobufs !

JoTurk:
I'm going to address the nits and merge this because I planned to add webrtc interop today.

conn_test.go
```diff
@@ -5804,3 +5804,176 @@ func TestConnectionKeyUsage(t *testing.T) {
...
+	assert.Equal(t, clientKM, serverKM)
+
+	// A different label must produce different material.
+	otherKM, err := clientState.ExportKeyingMaterial("EXPORTER-other", nil, exportLen)
+	assert.NoError(t, err)
+	assert.NotEqual(t, clientKM, otherKM)
+
+	// A non-empty context is not currently supported and must be rejected.
+	_, err = clientState.ExportKeyingMaterial(exportLabel, []byte{0x00}, exportLen)
+	assert.ErrorIs(t, err, dtlserrors.ErrContextUnsupported)
```
JoTurk: This isn't correct for DTLS 1.3

conn_test.go
```diff
@@ -5804,3 +5804,176 @@ func TestConnectionKeyUsage(t *testing.T) {
...
+			serverState, ok := server.ConnectionState()
+			assert.True(t, ok)
+
+			assert.Equal(t, tc.expectNegotiated, clientState.NegotiatedVersion(),
+				"client negotiated version")
+			assert.Equal(t, tc.expectNegotiated, serverState.NegotiatedVersion(),
+				"server negotiated version")
+		})
+	}
+}
```
JoTurk: This test isn't related to this PR, we have similar version tests

JoTurk:
I removed the tests, and will rework the 1.2 ones to also test 1.3

JoTurk:
reworked the existing TestExportKeyingMaterial to test both 1.2 and 1.3 the PR is very minimal now.

JoTurk (approved):

enobufs:
Thanks @JoTurk - !
Looking forward to using DTLS v1.3 with Pion!

PR #1134, opened 2026-09-21
Fix WithMaxVersion's default version comment

NotAFlightRisk:
#### Description

The godoc on `WithMaxVersion` says DTLS 1.2 is the default minimum and the only version supported, so anyone readng pkg.go.dev for the default upper bound gets told the wrong one. An unset `MaxVersion` normalises to 1.2, and that's the maximum. 1.3 goes through the option fine.

Its `WithMinVersion`'s line copied down, where "minimum" is right and the "only supported version" half is equally out of date.

#### Reference issue

None.


JoTurk (approved):
1.3 isn't released yet in any tagged version, we were going to change this comment before we change default version, thank you

PR #1136, opened 2026-09-21
Allow intercepting and injecting DTLS handshake packets

fippo:
which allows embedding them into STUN.

(squash and rebase of #766)


JoTurk:
I still don't understand why this can't be done with detachedConn? :) sounds like a really bad idea to export this level of abstraction from a generic DTLS library, maybe if it was internal to pion/webrtc

JoTurk:
Maybe try it on top of https://github.com/pion/webrtc/pull/3555 (this is what DTLS will look like in pion/webrtc after v4)

fippo:
well, detachedConn basically made it impossible to tell whether a packet was part of the handshake, even internally.  Which is why the mechanical rebase restores this.

Food for thought: that we are talking about api/rebase churn on a thing for which OpenSSL has offered behavior for a decade or so.

Ignore me for a while... where is the "draft" button hiding?

JoTurk:
> well, detachedConn basically made it impossible to tell whether a packet was part of the handshake, even internally. Which is why the mechanical rebase restores this.

This isn't true, Internally pion knows, and we export the event DetachedHandshakeDone, we can make the reader export the datagram state / stage. since it's not tagged yet.

> Ignore me for a while... where is the "draft" button hiding?

I'm just trying to avoid wasting your time, we have a really small window before we release v4 and we'll not be able to change the existing APIs.

fippo:
pion/dtls internally knows but it does not expose that which makes the internal APIs meh. The previous design was better *internally* without leaking.

No worries wrt wasting time, all good.

JoTurk:
Sounds good, feel free to wire this metadata from writePacketsWithResultLocked to the event struct https://github.com/pion/dtls/blob/main/detached_conn.go#L41-L48 we'll move to DetachedConn once v4 is released, webrtc will not use conn anymore, sounds like a simple change.

Edit: I understand what you mean now, DetachedConn return flight batches not datatagram, i thought you need to know the handshake / flight state, if you just want to know if it's a flight then it just does that :)

JoTurk:
Also @fippo detachedConn guarantees that the batch is a full a flight (at the DTLS level like a full "ClientHello flight"), it returns byte[][], and when the handshake is done the event is DetachedHandshakeDone, so you can know related flight / handshake datagrams because they are just returned in the same batch. and you know when the handshake is done.

So "DetachedConn basically made it impossible to tell whether a packet was part of the handshake" isn't true but if you want to add an explicit meta data for the flight stage / name that's also fine.

